PR #7, opened 2016-01-11
Update the URL links for the kits to be correct (not relative)

jeffaco:


jeffaco:
Approved by Anurag at my desk.

PR #15, opened 2016-01-22
Hotfix for Nagios/Zabbix Alert

sw47:
Make sure output plugin does not group empty or null records

### DIFF
--- a/source/code/plugins/filter_nagios_log.rb
+++ b/source/code/plugins/filter_nagios_log.rb
@@ -12,7 +12,11 @@ module Fluent
 			
 		# each record represents one line from the nagios log
 		def filter(tag, time, record)
-			return @nagios_lib.parse_and_wrap(record["message"])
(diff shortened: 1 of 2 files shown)


sw47:
@Microsoft/omsagent-devs Please Review

sw47:
@Microsoft/omsagent-devs  Please Review.

Only changing alert input / filter plugins for Nagios / Zabbix. Validated on Nagios and Zabbix server. No other changes.

jeffaco:
Nicolas has comments, so don't commit until he chimes in.

Please squash your commits down to a single commit, something like:

Bug fix: Make sure Nagios/Zabbix doesn't group up empty or null records

Do not commit the fact that you reverted a change.

If you need help with squashing, let me or Nicolas know.

Otherwise, I sign off.

niroyb:
Also add defensive programming in the output plugin to prevent this from happening. (Ignore null records, Maybe warn once)
Otherwise signed off.

PR #18, opened 2016-01-29
Changing method name to get rid of log error message

agup006:
This is referenced by in line 40 of filter_syslog.rb as warn_once, changing the method name to be the same

```ruby
host_ip = @ip_cache.get_ip(hostname)
      if host_ip.nil?
          OMS::Log.warn_once("Failed to get the IP for #{hostname}.")
      else
```


agup006:
@Microsoft/omsagent-devs

robbiezhang:
signoff

jeffaco:
This has been merged, closing

PR #19, opened 2016-01-29
fix the wrong warning method name in oms_configuration.rb

robbiezhang:
It uses the wrong method to log the warning. Same issue as in the filter_syslog.rb. 

@Microsoft/omsagent-devs


niroyb:
Sign off

PR #24, opened 2016-02-09
Change onwership to user after Ruby build is moved to intermediate dir

jeffaco:
This allows 'git clean' to work properly without sudo elevation


niroyb:
Signed off

PR #29, opened 2016-02-24
Adding documentation for high volume syslog collection, and Azure Ext…

agup006:
..ension documentation

@Microsoft/omsagent-devs

### DIFF
--- a/docs/OMS-Agent-for-Linux.md
+++ b/docs/OMS-Agent-for-Linux.md
@@ -361,6 +361,7 @@ If you are using selinux, the semanage tool can be used to allow TCP traffic for
 	  tag oms.syslog
 </source>
 ```
+**Note:** ports 0-1024 are privileged, and require special permissions. As the OMS Agent for Linux runs as the `omsagent` user trying to allocate a source port of 0 to 1024 results in agent startup failure
 
 ### Performance metrics
 Performance metrics to collect are controlled by the configuration in `/etc/opt/microsoft/omsagent/conf/omsagent.conf`. The appendix to this document details available classes and metrics in this release of the agent.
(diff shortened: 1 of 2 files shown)


agup006:
I also ran this file under dos2unix which is why the line endings changed

jeffaco:
Use [this](https://github.com/Microsoft/OMS-Agent-for-Linux/pull/29/files?w=1) to ignore the line ending changes.

niroyb:
Signed off, nice video tutorial :+1:

docs/OMS-Agent-for-Linux.md
```diff
@@ -361,6 +361,7 @@ If you are using selinux, the semanage tool can be used to allow TCP traffic for
 	  tag oms.syslog
 </source>
 ```
+**Note:** ports 0-1024 are privileged, and require special permissions. As the OMS Agent for Linux runs as the `omsagent` user trying to allocate a source port of 0 to 1024 results in agent startup failure
```
jeffaco: I'm not sure this is the best of English, but if it makes sense to you, I guess it's fine. I certainly understand what you mean, but I'm unsure if an unexperienced user would.

Perhaps reword to something like:

**Note:** Ports 1-1024 are privileged, and require root permissions. Since the OMS Agent runs as the `omsagent` user (not `root`), using a port between 1 thru 1024 will result in an agent startup failure.

jeffaco:
Signed off. Note my suggestion, though.

agup006:
Checked in commit https://github.com/Microsoft/OMS-Agent-for-Linux/commit/f5d1c65536cf1eb732384988d851b18c7711baa8

PR #30, opened 2016-03-01
Compress the http request body by default

niroyb:
@Microsoft/omsagent-devs

### DIFF
--- a/source/code/plugins/oms_common.rb
+++ b/source/code/plugins/oms_common.rb
@@ -5,6 +5,7 @@ module OMS
     require 'net/http'
     require 'net/https'
     require 'time'
+    require 'zlib'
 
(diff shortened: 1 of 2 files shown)


source/code/plugins/oms_common.rb
```diff
@@ -76,16 +77,23 @@ def create_ods_http(ods_uri, proxy={})
       #   record: Hash. body of the request
       # returns:
       #   HTTPRequest. request to ODS
-      def create_ods_request(path, record)
+      def create_ods_request(path, record, compress=true)
```
robbiezhang: Please update the out_oms_blob.rb to set the compress to false.

robbiezhang:
Signoff

PR #31, opened 2016-03-02
Update the suffix of the blob to use a guid

robbiezhang:
The suffix does not requires to contain the file path. Use the time (year/month/day/hour) + uuid to make it unique should be fine.
@Microsoft/omsagent-devs


niroyb:
Signed off

PR #33, opened 2016-03-02
Make output plugin compression configurable

niroyb:
This allows to turn off the compression in the omsagent.conf file.
Compression is enabled by default.
@Microsoft/omsagent-devs


sw47:
signed off

robbiezhang:
Signoff

PR #34, opened 2016-03-03
Fix the bug in the encode before sending the request

robbiezhang:
we should expect record to be: {"DataItems" => [ {"Message" => 'message1'}, {"Message" => 'message2'}] }
@Microsoft/omsagent-devs @sw47


niroyb:
Signed off

PR #35, opened 2016-03-03
Changing SUDOERs file to be safer

agup006:
@Microsoft/omsagent-devs  @KrisBash


KrisBash:
We've found some cases where sudoers doesn't like the alias specification. It's not needed, so safer to use username directly.

niroyb:
Signed off

PR #36, opened 2016-03-03
Agent reliability / Malformed data

niroyb:
Multiple customers have reported reliability issues when forwarding data (i.e. syslog) to a Linux agent. We have identified (and fixed) some instances where encoding of these forwarded items fail to upload. However, the output plugin will buffer forever when a malformed data item cannot be processed.
=> extend output plugin to drop invalid data and continue processing.
=> Do not retry if uncertain to fail again
@Microsoft/omsagent-devs

### DIFF
--- a/source/code/plugins/oms_common.rb
+++ b/source/code/plugins/oms_common.rb
@@ -1,5 +1,10 @@
 module OMS
 
+  class RetryRequestException < Exception
+    # Throw this exception to tell the fluentd engine to retry and
+    # inform the output plugin that it is indeed retryable
+  end
+
   class Common
     require 'json'
     require 'net/http'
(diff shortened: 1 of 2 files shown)


source/code/plugins/oms_common.rb
```diff
@@ -144,23 +149,32 @@ def start_request(req, secure_http, ignore404 = false)
...
-          if res
-            if ignore404 and res.code == "404"
-              return ''
-            end
+          if ignore404 and res.code == "404"
+            return ''
+          end
 
-            res_summary = "(#{res.code} #{res.message} #{res.body})"
+          res_summary = "(#{res.class.name} #{res.code} #{res.message} #{res.body})"
```
robbiezhang: can you add separator between the 4 parts?

source/code/plugins/out_oms.rb
```diff
@@ -95,6 +95,15 @@ def write(chunk)
       unmergable_records.each { |tag, record|
         handle_record(tag, record)
       }
+    rescue OMS::RetryRequestException => e
+      @log.debug "Encountered retryable exception. Will retry sending data later. #{e}"
+      # Re-raise the exception to inform the fluentd engine we want to retry sending this chunk of data later.
```
robbiezhang: is it possible to set an exponential retry interval?

robbiezhang:
Signoff

source/code/plugins/out_oms.rb
niroyb: It's already handled by the engine and is configurable. http://docs.fluentd.org/articles/output-plugin-overview#retrywait-maxretrywait

PR #39, opened 2016-03-08
Change sudoers logic to be sensitive to #includedir directive

jeffaco:
@Microsoft/ostc-devs 

Doing a pbuild now to test resulting kit


niroyb:
Signed off

PR #41, opened 2016-03-08
Log into debug when OMS Blob output plugin successfully upload the data

robbiezhang:
Log into debug when OMS Blob output plugin successfully upload the data
@Microsoft/omsagent-devs


niroyb:
Signed off

jeffaco:
Your commit message was a little unclear (although the code was very clear).  Perhaps do something like:

```
commit -m "Log using debug severity when OMS Blob output plugin successfully uploads the data" --amend
```

Otherwise, signed off.

PR #42, opened 2016-03-10
Don't deregsiter the OMS service when upgrading via shell bundle

jeffaco:
This is problematic when installing, then upgrading in place (from
same version to same version). RPM won't perform the upgrade since
it doesn't need to, but since shell bundle deregistered the service,
you're left in a broken state.

In retrospect, this wasn't needed. We deregister during pre-install,
and that's good enough.


niroyb:
Signed off

PR #43, opened 2016-03-10
Create and deploy new uninstall script for omsagent

jeffaco:

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -398,61 +398,71 @@ cd $EXTRACT_DIR
 # Do we need to remove the package?
 set +e
 if [ "$installMode" = "R" -o "$installMode" = "P" ]; then
-    pkg_rm omsconfig
-    pkg_rm omsagent
-
-    # If MDSD is installed and we're just removing (not purging), leave SCX
-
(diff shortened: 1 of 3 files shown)


installer/bundle/bundle_skel.sh
```diff
@@ -398,61 +398,71 @@ cd $EXTRACT_DIR
...
-            # Now just simply pkg_rm scx and omi
-	    pkg_rm scx
-	    pkg_rm omi
+	if [ ! -d /var/lib/waagent/Microsoft.OSTCExtensions.LinuxDiagnostic-*/mdsd -o "$installMode" = "P" ]; then
+	    if [ -f /opt/microsoft/scx/bin/uninstall ]; then
+		/opt/microsoft/scx/bin/uninstall $installMode
+	    else
+		for i in /opt/microsoft/*-cimprov; do
+		    PKG_NAME=`basename $i`
+		    if [ "$PKG_NAME" != "*-cimprov" ]; then
```
niroyb: How can `"$PKG_NAME" != "*-cimprov" ` if the loop is on files ending in `*-cimprov`?

niroyb:
Signed off

installer/bundle/bundle_skel.sh
jeffaco: If there are no matching files, then bash will iterate once over `/opt/miccrosoft/*-cimprov`, and I didn't want to pass that to pkg_rm.

PR #46, opened 2016-03-15
Modify to not upgrade to same version of kit already installed

jeffaco:
@Microsoft/ostc-devs @Microsoft/omsagent-devs 

This will be going into a lot of different repos (here, SCX, apache, mysql, docker). A close look would be appreciated.

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -6,7 +6,6 @@
 
 # This script is a skeleton bundle file for ULINUX only for project OMS.
 
-set -e
 PATH=/usr/bin:/usr/sbin:/bin:/sbin
 umask 022
 
@@ -53,6 +52,8 @@ usage()
     echo "  --restart-deps         Reconfigure and restart dependent service(s)."
     echo "  --source-references    Show source code reference hashes."
     echo "  --upgrade              Upgrade the package in the system."
+    echo "  --version              Version of this shell bundle."
+    echo "  --version-check        Check versions already installed to see if upgradable."
     echo "  --debug                use shell debug mode."
     echo
     echo "  -w id, --id id         Use workspace ID <id> for automatic onboarding."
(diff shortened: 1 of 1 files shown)


jeffaco:
Note that I untabified for consistency. I tried tabifying, but that resulted in more changes. You may want to review ignoring white space changes ...

installer/bundle/bundle_skel.sh
```diff
@@ -87,6 +88,80 @@ cleanup_and_exit()
...
+    return 1
+}
+
+getVersionNumber()
+{
+    # Parse a version number from a string.
+    #
+    # Parameter 1: string to parse version number string from
+    #     (should contain something like mumble-4.2.2.135.universal.x86.tar)
+    # Parameter 2: prefix to remove ("mumble" in above example)
```
niroyb: The prefix to remove is `mumble-`

installer/bundle/bundle_skel.sh
```diff
@@ -87,6 +88,80 @@ cleanup_and_exit()
...
+
+    if [ $# -ne 2 ]; then
+        echo "INTERNAL ERROR: Incorrect number of parameters passed to getVersionNumber" >&2
+        cleanup_and_exit 1
+    fi
+
+    echo $1 | sed -e "s/$2//" -e 's/\.universal\..*//' -e 's/\.x64.*//' -e 's/\.x86.*//' -e 's/-/./'
```
niroyb: `' -e 's/\.x64.*//' -e 's/\.x86.*//'` is redundant if we remove `.universal.` and everything after.

installer/bundle/bundle_skel.sh
jeffaco: Fixed.

installer/bundle/bundle_skel.sh
jeffaco: Unfortunately not. DSC kits look like this:

```
DSC_PKG=omsconfig-1.1.1-121.x64
```

No `.universal.` prefix, so I needed a special case to handle that.

installer/bundle/bundle_skel.sh
```diff
@@ -307,6 +448,42 @@ do
...
+            # OMS agent itself
+            versionInstalled=`getInstalledVersion omsagent`
+            versionAvailable=`getVersionNumber $OMS_PKG omsagent-`
+            if shouldInstall_omsagent; then shouldInstall="Yes"; else shouldInstall="No"; fi
+            printf '%-15s%-15s%-15s%-15s\n' omsagent $versionInstalled $versionAvailable $shouldInstall
+
+            # omsconfig
+            versionInstalled=`getInstalledVersion omsconfig`
+            versionAvailable=`getVersionNumber $DSC_PKG omsconfig-`
+            if ! ptyon_ctypes_installed 1> /dev/null 2> /dev/null; then ctypes_text=" (No ctypes)"; fi
```
niroyb: python typo?

niroyb:
Signed off

installer/bundle/bundle_skel.sh
jeffaco: I only tested this in the failed case and, since stderr/stdout was redirected, I missed this in my testing. Thanks for the good catch!

PR #48, opened 2016-03-18
Fix the bundle upgrade issue in DPKG

robbiezhang:
In DPKG, the $version string contains whitespace, which will be regarded as 2 parameters for the getVersionNumber.
@Microsoft/omsagent-devs

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -315,7 +315,7 @@ getInstalledVersion()
     if check_if_pkg_is_installed $1; then
         if [ "$INSTALLER" = "DPKG" ]; then
             local version=`dpkg -s $1 2> /dev/null | grep "Version: "`
-            getVersionNumber $version "Version: "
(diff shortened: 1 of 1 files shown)


installer/bundle/bundle_skel.sh
```diff
@@ -315,7 +315,7 @@ getInstalledVersion()
     if check_if_pkg_is_installed $1; then
         if [ "$INSTALLER" = "DPKG" ]; then
             local version=`dpkg -s $1 2> /dev/null | grep "Version: "`
-            getVersionNumber $version "Version: "
+            getVersionNumber "$version" "Version: "
```
jeffaco: I just observed this in a different repo. This alone isn't enough, as the line above will fail as well due to a space in the assignment.

jeffaco:
This isn't enough, as mentioned above. I'll push a fix to mainline now, please revert this.

robbiezhang:
Canceled since Jeff has a fix

PR #49, opened 2016-03-18
Made onboarding curl verbose by default

sugr4:
'-v' can't be turned on when onboarding using file. So made this curl verbose to troubleshoot error during onboarding request.


robbiezhang:
@Microsoft/omsagent-devs 

Suju, please @ the both group for review.

niroyb:
This is not the correct solution to the problem.
Verbose is optional. I'll send a fix shortly.

niroyb:
Look at https://github.com/Microsoft/OMS-Agent-for-Linux/pull/50

PR #50, opened 2016-03-18
Fix file onboard to work with verbose flag

niroyb:
@Microsoft/omsagent-devs @sugr4

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -215,10 +215,7 @@ parse_args()
         fi
     fi
 
-    if [ "$VERBOSE" = "1" ]; then
-        echo "Workspace ID:  $WORKSPACE_ID"
-        echo "Shared key:    $SHARED_KEY"
-    else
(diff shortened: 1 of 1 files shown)


robbiezhang:
Looks good. Signoff.

installer/scripts/omsadmin.sh
```diff
@@ -215,7 +215,7 @@ parse_args()
         fi
     fi
 
-    if [ "$VERBOSE" = "1" ]; then
+    if [ "$VERBOSE" = "1" -a -n "$WORKSPACE_ID" ]; then
```
sugr4: Looks good, but what about when $VERBOSE=1 -a -z "WORKPSACE_ID"?
Maybe just edit
    if [ $VERBOSE -ne 1 ]; then
        CURL_VERBOSE=-s
    fi

niroyb: Made it cleaner. Thanks

PR #51, opened 2016-03-22
Don't force shell bundle install unless --force specifed on command line

jeffaco:
@Microsoft/ostc-devs @Microsoft/omsagent-devs


niroyb:
Signed off

PR #55, opened 2016-03-30
Upload the file changes to the same blob for Custom Log

robbiezhang:
Upload the file changes to the same blob by calculating the sha256 hash of the local file path. Consist to the Windows CustomLog Behavior
@Microsoft/omsagent-devs

### DIFF
--- a/source/code/plugins/out_oms_blob.rb
+++ b/source/code/plugins/out_oms_blob.rb
@@ -10,6 +10,7 @@ module Fluent
       super
 	  
       require 'base64'
+      require 'digest'
       require 'json'
       require 'net/http'
       require 'net/https'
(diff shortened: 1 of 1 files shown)


source/code/plugins/out_oms_blob.rb
```diff
@@ -218,10 +220,17 @@ def handle_records(tag, records)
           # extra tags for CustomLog:
           # tags[4]: custom data type
           custom_data_type = tags[4]
-          suffix = Time.now.utc.strftime("d=%Y%m%d/h=%H/#{SecureRandom.uuid}")
+
+          # tags[5..-1]: monitoring file name
+          # concat all the rest parts with /
+          filePath = ''
```
niroyb: Use the [array join method.](http://ruby-doc.org/core-2.2.0/Array.html#method-i-join)
`filepath = tags[5..-1].join('/')`
Then on line 230, there is no need to remove the trailing slash.

niroyb:
Signed off

PR #56, opened 2016-03-31
Change log level of retryable exception message

niroyb:
In reply to [this comment ](https://github.com/Microsoft/OMS-Agent-for-Linux/commit/0cb36e71dd6d5c7da0104220c26b86cba660aca8#commitcomment-16920041) by @agup006 
Moved the actual printing of the error to a separate debug message because the error detail is not interesting at the info level.
@Microsoft/omsagent-devs


robbiezhang:
Signoff

PR #57, opened 2016-03-31
Add a proxy test

niroyb:
This change makes it easier to understand why test_onboard_proxy_failure fails.
@Microsoft/omsagent-devs


robbiezhang:
Signoff

PR #60, opened 2016-04-02
Do not copy the change tracking config into the omsagent.d folder

robbiezhang:
@Microsoft/omsagent-devs @niroyb 
We should not copy the change tracking config into the omsagent.d in the installation. It should be added by the nxOMSPlugin module based on whether the customer enables the change tracking solution.
I still binplace the file into the /etc/opt/microsoft/omsagent/sysconf/omsagent.d folder, so that customer can copy them manually.


agup006:
Sign off

PR #61, opened 2016-04-06
Made the default num of threads as 5

sugr4:
@Microsoft/omsagent-devs  @agup006


niroyb:
Why 5? Is it arbitrary?
I'm guessing it's for performance. It would be nice if the reason of the change was part of the commit message.
Otherwise signed off

sugr4:
Yes it's arbitrary, to manage excessive buffering.

PR #63, opened 2016-04-07
Set timezone info in the CustomLog blob header

robbiezhang:
Get the current timezone from the agent, map it to the Windows Time Zone, and set it into the blob header
@Microsoft/omsagent-devs


source/code/plugins/oms_common.rb
```diff
@@ -11,17 +11,494 @@ class Common
...
+
+        tzID = 'Unknown'
+        
+        begin
+          md5sum = Digest::MD5.file('/etc/localtime').hexdigest
+          baseFolder = '/usr/share/zoneinfo/'
+          rightFolder = 'right/'
+          
+          Dir.glob("#{baseFolder}**/*") { |x|
+            if !File.directory? x and Digest::MD5.file(x).hexdigest == md5sum
```
niroyb: Use `File.file?` instead of `!File.directory?`</br>
Rename `x` to something meaningful.
This is calculating the hash of every file globbed but only enters the if clause when the md5 matches. Why not execute the if clause directly on /etc/localtime?

source/code/plugins/oms_common.rb
```diff
@@ -11,17 +11,494 @@ class Common
...
+              tzID = "#{x[baseFolder.length..-1]}"
```
niroyb: No need to convert a string to string. Use `x[baseFolder.length..-1]`

source/code/plugins/oms_common.rb
```diff
@@ -11,17 +11,494 @@ class Common
...
+        return @@CurrentTimeZone if !@@CurrentTimeZone.nil?
+
+        tzID = 'Unknown'
+        
+        begin
+          md5sum = Digest::MD5.file('/etc/localtime').hexdigest
+          baseFolder = '/usr/share/zoneinfo/'
+          rightFolder = 'right/'
+          
+          Dir.glob("#{baseFolder}**/*") { |x|
```
niroyb: Why not restrict globbing to `/usr/share/zoneinfo/right`?
It seems to be the intent?

source/code/plugins/oms_common.rb
```diff
@@ -11,17 +11,494 @@ class Common
...
+            if !File.directory? x and Digest::MD5.file(x).hexdigest == md5sum
+              tzID = "#{x[baseFolder.length..-1]}"
+              tzID = tzID[rightFolder.length..-1] if tzID.start_with?(rightFolder)
```
niroyb: This looks redundant with the previous line of code. Are there situations where the path has `/right/right/`?

source/code/plugins/oms_common.rb
```diff
@@ -11,17 +11,494 @@ class Common
...
+             
+              if !@@tzMapping[tzID].nil?
```
niroyb: Use `if @@tzMapping.has_key?(tzID)` for clarity.
The nil value of an non-existing key is more of a side effect.

source/code/plugins/oms_common.rb
robbiezhang: It looks for a match with /etc/localtime in the /usr/share/zoneinfo/. I cannot skip the right/ folder, because the file is different from the one not in right.

source/code/plugins/oms_common.rb
robbiezhang: updated

source/code/plugins/oms_common.rb
robbiezhang: Update

source/code/plugins/oms_common.rb
robbiezhang: /usr/share/zoneinfo is the base folder. there won't be right/right/.

niroyb:
Signed off

robbiezhang:
update the PR per feedbacks
@Microsoft/omsagent-devs

source/code/plugins/oms_common.rb
```diff
@@ -11,17 +11,520 @@ class Common
...
+        return tzID
+      end # end get_unified_timezoneid
+
+      def get_current_timezone
+        return @@CurrentTimeZone if !@@CurrentTimeZone.nil?
+
+        tzID = 'Unknown'
+        
+        begin
+          # looks for a file in the /usr/share/zoneinfo/, which is identical to /etc/localtime. use the file name as the timezone
```
niroyb: Move this comment lower to where it is relevant.
There could be a comment here like "if /etc/localtime is a valid symlink use the target as the current timezone."

PR #65, opened 2016-04-08
Change the header name of the TimeZoneId to all lowercase

robbiezhang:
Align with the Windows one
@Microsoft/omsagent-devs


agup006:
signoff

PR #66, opened 2016-04-11
Preserve Case in Http Request

robbiezhang:
Create CaseSensitiveString < String class (override the downcase and capitalize method to return self) and use it in the Http Request header to preserve the case in header name. This is a workaround that Ruby library converts the header name (http://stackoverflow.com/questions/10258893/stop-ruby-http-request-modifying-header-name) while the OMS workflow is case sensitive.
@Microsoft/omsagent-devs

### DIFF
--- a/source/code/plugins/oms_common.rb
+++ b/source/code/plugins/oms_common.rb
@@ -729,4 +729,13 @@ module OMS
 
   end
 
+  class CaseSensitiveString < String
+    def downcase
+        self
+    end
+    def capitalize
+        self
+    end
+  end
+
 end # module OMS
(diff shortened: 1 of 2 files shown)


jeffaco:
Signed off.

Note that git standards call for a short summary for the commit string, and that commit strings are <= 72 bytes per line. Given that, I'd suggest redoing your commit message to something like:


Create CaseSensitiveString to preserve case in header name

Create CaseSensitiveString < String class (override the downcase and capitalize method to return self) and use it in the Http Request header to preserve the case in header name. This is a workaround that Ruby library converts the header name (http://stackoverflow.com/questions/10258893/stop-ruby-http-request-modifying-header-name) while the OMS workflow is case sensitive.


That way, the primary message (for git lol, etc) is short and understandable.

PR #67, opened 2016-04-14
Show timeskew error when onboarding if 403 is caused by timeskew

sugr4:


jeffaco:
Assuming that you actually tested this on a machine with this change: Signed off.

If you did not test on a machine, please set the time to something wrong and try with this change just to be sure!

sugr4:
Yes I did test this :)

PR #68, opened 2016-04-14
Copy dsc zip files to omsagent/target

sugr4:
@Microsoft/omsagent-devs

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -254,11 +254,15 @@ $(IN_PLUGINS_LIB) : $(STATIC_PLUGINS_OBJFILES)
 
 ifeq ($(ULINUX),1)
 
-kit : $(TARGET_DIR)/$(OUTPUT_PACKAGE_PREFIX).sh
+kit : $(TARGET_DIR)/$(OUTPUT_PACKAGE_PREFIX).sh $(TARGET_DIR)/dsc
 
 $(TARGET_DIR)/$(OUTPUT_PACKAGE_PREFIX).sh : $(INTERMEDIATE_DIR)/$(OUTPUT_PACKAGE_PREFIX).tar
 	@echo "========================= Performing Building shell bundle"
 	../installer/bundle/create_bundle.sh $(TARGET_DIR) $(INTERMEDIATE_DIR) $(OUTPUT_PACKAGE_PREFIX).tar
+
+$(TARGET_DIR)/dsc : $(DSC_TARGET_DIR)
+        @echo "========================= Performing Copying DSC resources"
+        $(MKPATH) $(TARGET_DIR)/dsc; $(COPY) $(DSC_TARGET_DIR)/*.zip $(TARGET_DIR)/dsc/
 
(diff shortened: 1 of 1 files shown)


robbiezhang:
Signoff

build/Makefile
```diff
@@ -296,6 +296,8 @@ $(INTERMEDIATE_DIR)/$(OUTPUT_PACKAGE_PREFIX).tar : $(RUBY_DEST_DIR) $(IN_PLUGINS
 	cd $(DSC_DIR)/release; $(COPY) `ls omsconfig-*.ssl_100.*.deb | sort | tail -1` $(INTERMEDIATE_DIR)/100
 	cd $(INTERMEDIATE_DIR)/100; for f in omsconfig-*.{rpm,deb}; do SOURCE=$$f; DEST=`echo $$SOURCE | sed 's/.ssl_100././'` ; $(MV) $$SOURCE $$DEST; done
 
+	cd $(DSC_DIR)/release; $(COPY) `ls *.zip` $(TARGET_DIR)/
```
robbiezhang: Can you copy it into $(TARGET_DIR)/$(ARCH)/? It will help us to find the packages and merge them in build runner.

jeffaco: No, I think this is wrong. $(TARGET_DIR) already contains an architecture designation, so you don't need it. I think it's redundant, on the build disk, to have something like omsagent/version/target_dir_with_arch/arch/<stuff>.

Please don't commit this yet, I'd like to take a closer look at this tomorrow morning. But this change is not right, not unless we decide to change the build structure in general.

jeffaco: Sorry for the delay in the detailed review.

While, technically, your change would work, I wouldn't suggest this. The area you're putting the change into is copying bits for the embedded tar file in the shell bundle, and this has nothing to do with that.

At line 257, there is a dependency for building a kit. I'd make this part of that. The current line at 257 reads:

```
kit : $(TARGET_DIR)/$(OUTPUT_PACKAGE_PREFIX).sh
```

I would replace this with something like:

```
kit : $(TARGET_DIR)/$(OUTPUT_PACKAGE_PREFIX).sh $(TARGET_DIR)/dsc

$(TARGET_DIR)/dsc : $(DSC_TARGET_DIR)
	@echo "========================= Performing Copying DSC resources"
	cd $(DSC_TARGET_DIR); $(COPY) *.zip $(TARGET_DIR)/ 
```

This puts the change in it's own section, where it belongs since it has nothing to do with building a .tar file. It's also dependency based, insuring that DSC is built first due to prior rules.

Finally, there was no need to do an `ls *.tar` in your copy line. You can just copy directly without doing the 'ls' operation.

Run this through pbuild to insure that it does what we expect. But this looks right.

sugr4:
I made the changes and it works correctly in pbuild.

jeffaco:
Signed off.

Please revert the Jenkins changes after this is committed, thanks (if that wasn't already done).

PR #69, opened 2016-04-18
Increased num of threads for output plugin

sugr4:
@agup006


jeffaco:
Signed off.

Suggested amending commit message to not needlessly abbreviate.  Something like:

`Increase number of threads for output plugin`

This can be done with `commit -m "Increase number of threads for output plugin" --amend`. Then do `git log` to make sure you only have one commit message, and that it's correct. Then merge normally to the master branch.

niroyb:
Signed off

PR #70, opened 2016-04-19
Script to merge DSC modules.

robbiezhang:
Move the merge DSC module script into the agent repository.
The gpg key and passphrase are supposed to be in test/config/testsigning folder. Or, you can specify the folder as the parameter.
@jeffaco @Microsoft/omsagent-devs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,9 @@
 # git ignore file for omsagent project
 
-intermediate
-target
-build/config.mak
-build/Makefile.version
-test/config
+/intermediate
+/target
+/build/config.mak
+/build/Makefile.version
+/test/config
+/build/x64
+/build/x86
--- a/build/mergeallmodules.sh
+++ b/build/mergeallmodules.sh
@@ -1,0 +1,157 @@
+#!/bin/bash
+
+#Assumes that ./x64 and ./x86 exist and contain the arch-specific copies of a DSC resource module
+
+zip -L >/dev/null 2>&1
+if [ $? != 0 ]; then
+    echo "zip must be installed"
+    exit 1
+fi
+
+unzip -Z >/dev/null 2>&1
+if [ $? != 0 ]; then
+    echo "unzip must be installed"
+    exit 1
+fi
+
+IntermediateDir="../intermediate"
+TargetDir="../target"
+DscModuleIntermediateDir="$IntermediateDir/merging_dsc_modules"
+DscModuleTargetDir="$TargetDir/dsc_modules"
+X64Dir="./x64"
+X86Dir="./x86"
+TestSigningDir=$1
+
+if [ "$TestSigningDir" = "" ]
+then
+    TestSigningDir="../test/config/testsigning"
+fi
+
+if [ -d "$TestSigningDir" ]
+then
+    gpg --list-secret-keys
+    if [ $? != 0 ]
+    then
+        echo "Failed to run the gpg command"
+        exit 1
+    fi
+
+    gpg --list-secret-keys | grep -q '2048R/8C3B51C6'
+    if [ $? != 0 ]
+    then
+        TestGPGKey="$TestSigningDir/gpg.asc"
+        if [ -f "$TestGPGKey" ]
+        then
+            echo "Import the test GPG key"
+            gpg --import $TestGPGKey
+            if [ $? != 0 ]
+            then
+                echo "Failed to import the test GPG key"
+                exit 1
+            fi
+        else
+            echo "Did not find the test GPG key at $TestGPGKey"
+            exit 1
+        fi
+    fi  
+
+    SigningKeyFilePath="$TestSigningDir/signingkeys.gpg"
+    SigningKeyPassphrase="$TestSigningDir/passphrase"
+
+    if [ ! -f "$SigningKeyFilePath" ]
+    then
+        echo "GPG key file is not found"
+        exit 1
+    fi
+
+    if [ ! -f "$SigningKeyPassphrase" ]
+    then
+        echo "GPG key passphrase is not found"
+        exit 1
+    fi
+else
+    echo "Signing key directory doesn't exist at location: $TestSigningDir"
+    exit 1
+fi
+
+set -e
+
+if [ -d "$DscModuleIntermediateDir" ]
+then
+    rm -rf "$DscModuleIntermediateDir"
+fi
+
+mkdir -p "$DscModuleIntermediateDir"
+
+for ModuleFilePath in $X64Dir/*.zip;
+do
+    # expected file name: ./x64/nxOMSPlugin_1.0.zip
+    ModuleFileName=`basename $ModuleFilePath`
+    ModuleName=`echo $ModuleFileName | cut -d _ -f 1`
+    
+    if [ -f "$X64Dir/$ModuleFileName" ]
+    then
+        echo "Found x64 module .zip: "$X64Dir/$ModuleFileName""
+    else
+        echo "Did not find x64 module .zip: "$X64Dir/$ModuleFileName""
+        exit 1
+    fi
+    
+    if [ -f "$X86Dir/$ModuleFileName" ]
+    then
+        echo "Found x86 module .zip: "$X86Dir/$ModuleFileName""
+    else
+        echo "Did not find x86 module .zip: "$X86Dir/$ModuleFileName""
+        exit 1
+    fi
+    
+    #Merge modules to ./intermediate/ModuleFileName
+    unzip -o $X86Dir/$ModuleFileName -d ./$DscModuleIntermediateDir
+    unzip -o $X64Dir/$ModuleFileName -d ./$DscModuleIntermediateDir
+    
+    WorkingDir="./$DscModuleIntermediateDir/$ModuleName/"
+    
+    (
+        #Generate Sha256sums
+        echo "Generating sha256sums..."
+        cd $WorkingDir
+        find . -type f -print0 | xargs -0 sha256sum |grep -v sha256sums > ./$ModuleName.sha256sums
+    
+        #Test Sha256sums
+        echo "Test sha256sums..."
+        sha256sum -c $ModuleName.sha256sums
+    )
+
+    #Test sign module
+    echo "Signing .sha256sums file..."
+    if ! cat $SigningKeyPassphrase | gpg --no-default-keyring --keyring $SigningKeyFilePath --yes --passphrase-fd 0 --output ./$WorkingDir/$ModuleName.asc --armor --detach-sign ./$WorkingDir/$ModuleName.sha256sums; then
+        echo "Failed to sign the .sha256sums file"
+        exit 1
+    fi
+
+    echo "Verifying signature..."
+    if ! gpg --no-default-keyring --keyring $SigningKeyFilePath --verify  ./$WorkingDir/$ModuleName.asc ./$WorkingDir/$ModuleName.sha256sums; then
+        echo "Signature is invalid"
+        exit 1
+    fi
+
+    (
+        echo "Producing .zip file"
+        cd $DscModuleIntermediateDir
+        zip -r $ModuleName.zip $ModuleName/*
+    )
+done
+
+if [ -d "$DscModuleTargetDir" ]
+then
+    rm -rf "$DscModuleTargetDir"
+fi
+
+mkdir -p $DscModuleTargetDir
+
+if ! cp $DscModuleIntermediateDir/*.zip $DscModuleTargetDir/; then
+    echo "Failed to copy the merged packages into the target folder"
+    exit 1
+fi
+
+exit 0


jeffaco:
In your PR message you said:

```
NOTE: the machine which runs the script needs to import gpg.asc in the private repo.
```

I need clarification on that. If I need to run something special to do this, I need to know exactly how to run the command (specific example please, preferably in the top of the script in comments).

Can the script itself test if the gpg.asc has been imported into the private repo?

build/mergeallmodules.sh
```diff
@@ -0,0 +1,123 @@
...
+
+unzip -Z >/dev/null 2>&1
+if [ $? != 0 ]; then
+        echo "unzip must be installed"
+        exit 1
+fi
+
+TargetDir="./target"
+x64dir="./x64"
+x86dir="./x86"
```
jeffaco: If this is intended to run from the build directory, then you should probably add:

```
/build/target
/build/x64
/build/x86
```

To the .gitignore file. Or better yet: Change x86dir and x64dir to be in the intermediate directory, and TargetDir to be in the target directory, and make sure the script works with that (i.e. do 'mkdir -p' to insure the target directory exists). If you do this, then no .gitignore changes are needed, as those directories are already ignored.

build/mergeallmodules.sh
```diff
@@ -0,0 +1,123 @@
...
+		exit 1
+	fi
+
+	if [ ! -f "$SigningKeyPassphrase" ]
+	then
+		echo "GPG key passphrase is not found"
+		exit 1
+	fi
+else
+	echo "Signing key directory doesn't exit. $TestSigningDir"
```
jeffaco: Suggest changing to something like:

`echo "Signing key directory doesn't exist at location: $TestSingingDir"`

This is clearer.

build/mergeallmodules.sh
```diff
@@ -0,0 +1,123 @@
...
+if [ -d "$TargetDir" ]
+then
+	rm -rf "$TargetDir"
+fi
+
+mkdir "$TargetDir"
+
+for ModuleFilePath in x64/*.zip;
+do
+        ModuleFileName=`echo $ModuleFilePath | xargs -n 1 basename`
```
jeffaco: Indentation

build/mergeallmodules.sh
```diff
@@ -0,0 +1,123 @@
...
+	ModuleName=`echo $ModuleFileName | cut -d _ -f 1`
+	Version=`echo $ModuleFileName | cut -d _ -f 2 |cut -d . -f 1,2`
```
jeffaco: Gosh, this seems fragile, but if you need the ModuleName, you need it. I wonder, though, if you could get away without cutting the version information? If you need this, then please add a comment as to the exact format of the filename that you expect, thanks.

Note that `Version` doesn't appear to be used anywhere, so you can likely delete it.

build/mergeallmodules.sh
```diff
@@ -0,0 +1,123 @@
...
+	cat $SigningKeyPassphrase | gpg --no-default-keyring --keyring $SigningKeyFilePath --yes --passphrase-fd 0 --output ./$WorkingDir/$ModuleName.asc --armor --detach-sign ./$WorkingDir/$ModuleName.sha256sums
+	if [ $? != 0 ]; then
+		echo "Failed to sign the .sha256sums file"
+		exit 1
+	fi
+
+	echo "verifying signature..."
+	gpg --no-default-keyring --keyring $SigningKeyFilePath --verify  ./$WorkingDir/$ModuleName.asc ./$WorkingDir/$ModuleName.sha256sums
+	if [ $? != 0 ]; then
+		echo "signture is invalid"
```
jeffaco: Capitalize 's' of 'signature'.

build/mergeallmodules.sh
```diff
@@ -0,0 +1,123 @@
...
+	fi
+
+	echo "Producing .zip file"
+	(
+		cd ./target
+		zip -r $ModuleName.zip $ModuleName/*
+	)
+done
+
+cp ./target/*.zip ../target/
```
jeffaco: So what it is, EXACTLY, that's getting published. What are you after here, the "real" target directory, at one level up from the architecture specific directories. It sounds like "./target" isn't really a target directory, it's a working directory. I strongly recommend that, rather than creating new directories with confusion of what's what, you modify this to meet our existing standards: Temporary directories go into "intermediate" and files to stage go into "target" (based from omsagent directory).

Otherwise, this is just needlessly confusing. I'm not even certain what to stage to the build directories at this point.

jeffaco:
I would like to see one more iteration (minimally) clarifying what is intermediate and what is to be staged, and where, on the build disk. Thanks.

build/mergeallmodules.sh
```diff
@@ -0,0 +1,123 @@
...
+do
+        ModuleFileName=`echo $ModuleFilePath | xargs -n 1 basename`
+	ModuleName=`echo $ModuleFileName | cut -d _ -f 1`
+	Version=`echo $ModuleFileName | cut -d _ -f 2 |cut -d . -f 1,2`
+	
+	if [ -f "$x64dir/$ModuleFileName" ]
+	then
+		echo "Found x64 module .zip: "$x64dir/$ModuleFileName""
+	else
+		echo "Did not file x64 module .zip: "$x64dir/$ModuleFileName""
```
niroyb: Typo here and line 75. *find* instead of *file*

build/mergeallmodules.sh
```diff
@@ -0,0 +1,123 @@
...
+	#Merge modules to ./target/ModuleFileName
+	unzip -o $x86dir/$ModuleFileName -d ./target
+	unzip -o $x64dir/$ModuleFileName -d ./target
+	
+	WorkingDir="./target/$ModuleName/"
+	
+	#Generate Sha256sums
+	echo "Generating sha256sums..."
+	cd $WorkingDir
+	find . -type f -print0 | xargs -0 sha256sum |grep -v sha256sums > ./$ModuleName.sha256sums
```
niroyb: I don't think `grep -v sha256sums` is needed. It's part of the error message trying to calculate the hash of a folder. Here, find filters out folders and the error is printed on standard error, not standard out.

niroyb:
Signed off

jeffaco:
Just for clarification, I did not sign off. I would like to minimally see one more iteration clarifying what is intermediate and what is to be staged, and where, on the build disk. Thanks.

robbiezhang:
Use ../intermediate/merging_dsc_modules as the working folder
Use ../target/dsc_modules as the target folder
Ignore the x64 and x86 folder in the .gitignore
Check whether the test GPG key is imported in the machine. If not, import gpg.asc
Fix some wording and messaging issues
@jeffaco

jeffaco:
Please make this a single commit, we shouldn't have a second commit that says "Incoporate the feedbacks:", most Pull Requests have feedback.

The first commit comment is no longer appropriate ("Script to merge DSC modules."), since the script is elsewhere. I'd save the changes but recommit this with a better message, something like "Add components to sign DSC modules".

If you're unsure how to do this, swing by or ask me and I'll swing by your office. It's super easy to do.

Otherwise, signed off. I'll look at the other PR now in OMS-Agent-For-Linux.

.gitignore
```diff
@@ -5,3 +5,5 @@ target
 build/config.mak
 build/Makefile.version
 test/config
+build/x64
+build/x86
```
jeffaco: This is actually wrong. Please add a "/" before every line in this file (except for the comment and blank line).

If you have "build/x86", then that ignores "build/x86" in any directory path, and that's not what you want.

If you have "/build/x86", that that ignores build/x86 from the root of the .gitignore file, which is what you want.  The prior lines were in error too, we learned this after setting up OMS-Agent-For-Linux.

Thanks.

build/mergeallmodules.sh
```diff
@@ -0,0 +1,170 @@
...
+	fi
+
+	if [ ! -f "$SigningKeyPassphrase" ]
+	then
+		echo "GPG key passphrase is not found"
+		exit 1
+	fi
+else
+	echo "Signing key directory doesn't exist at location: $TestSigningDir"
+        exit 1
```
jeffaco: Indentation.

build/mergeallmodules.sh
```diff
@@ -0,0 +1,170 @@
+#!/bin/bash
+
+#Assumes that ./x64 and ./x86 exist and contain the arch-specific copies of a DSC resource module
+
+zip -L >/dev/null 2>&1
+if [ $? != 0 ]; then
+        echo "zip must be installed"
+        exit 1
```
jeffaco: Inconsistent indentation. What are you using to edit?

Space indent for clarity, with stops every 4 bytes. This should only be indented 4 bytes.

build/mergeallmodules.sh
```diff
@@ -0,0 +1,170 @@
...
+fi
+
+unzip -Z >/dev/null 2>&1
+if [ $? != 0 ]; then
+        echo "unzip must be installed"
+        exit 1
```
jeffaco: Indentation

build/mergeallmodules.sh
```diff
@@ -0,0 +1,170 @@
...
+TargetDir="../target"
+DscModuleIntermediateDir="$IntermediateDir/merging_dsc_modules"
+DscModuleTargetDir="$TargetDir/dsc_modules"
+X64Dir="./x64"
+X86Dir="./x86"
+TestSigningDir=$1
+
+if [ "$TestSigningDir" = "" ]
+then
+	TestSigningDir="../test/config/testsigning"
```
jeffaco: This indentation is okay! I won't mention indentation further, but please make sure it's correct. FWIW, emacs makes this trivially easy.

build/mergeallmodules.sh
```diff
@@ -0,0 +1,170 @@
...
+	fi
+
+	gpg --list-secret-keys | grep -q '2048R/8C3B51C6'
+	if [ $? != 0 ]
+	then
+		TestGPGKey="$TestSigningDir/gpg.asc"
+		if [ -f "$TestGPGKey" ]
+		then
+			echo "Import the test GPG key"
+			gpg --import $TestGPGKey
```
jeffaco: Do you need sudo access for this, or is it imported into an account-specific keyring?

build/mergeallmodules.sh
```diff
@@ -0,0 +1,170 @@
...
+	then
+		echo "GPG key passphrase is not found"
+		exit 1
+	fi
+else
+	echo "Signing key directory doesn't exist at location: $TestSigningDir"
+        exit 1
+fi
+
+set -e
```
jeffaco: Why not do `set -e` much earlier, and then test properly if the command succeeds in cases where you care? Most of the time, I do `set -e` at the top of the script.

build/mergeallmodules.sh
```diff
@@ -0,0 +1,170 @@
...
+
+if [ ! -d "$IntermediateDir" ]
+then
+	mkdir $IntermediateDir
+fi
+
+if [ ! -d "$TargetDir" ]
+then
+	mkdir $TargetDir
+fi
```
jeffaco: Suggest you remove lines 79-87, you don't really need them (`mkdir -p` is good for this).

build/mergeallmodules.sh
```diff
@@ -0,0 +1,170 @@
...
+
+if [ -d "$DscModuleIntermediateDir" ]
+then
+	rm -rf "$DscModuleIntermediateDir"
+fi
+
+mkdir "$DscModuleIntermediateDir"
```
jeffaco: mkdir -p

build/mergeallmodules.sh
```diff
@@ -0,0 +1,170 @@
...
+		zip -r $ModuleName.zip $ModuleName/*
+	)
+done
+
+if [ -d "$DscModuleTargetDir" ]
+then
+	rm -rf "$DscModuleTargetDir"
+fi
+
+mkdir $DscModuleTargetDir
```
jeffaco: mkdir -p

build/mergeallmodules.sh
```diff
@@ -0,0 +1,170 @@
...
+	
+		#Test Sha256sums
+		echo "Test sha256sums..."
+		sha256sum -c $ModuleName.sha256sums
+	)
+
+	#Test sign module
+	echo "Signing .sha256sums file..."
+	cat $SigningKeyPassphrase | gpg --no-default-keyring --keyring $SigningKeyFilePath --yes --passphrase-fd 0 --output ./$WorkingDir/$ModuleName.asc --armor --detach-sign ./$WorkingDir/$ModuleName.sha256sums
+	if [ $? != 0 ]; then
```
jeffaco: I'm not 100% sure that form of test will work with `set -e`. Please test this with a bad passphrase or something.

When using `set -e`, we generally use the `if <stmt>` type test, which definitely works. in this case, it would be something like:

```
if ! cat $SigningKeyPassphrase | gpg --no-default-keyring --keyring $SigningKeyFilePath --yes --passphrase-fd 0 --output ./$WorkingDir/$ModuleName.asc --armor --detach-sign ./$WorkingDir/$ModuleName.sha256sums; then
```

build/mergeallmodules.sh
```diff
@@ -0,0 +1,170 @@
...
+		echo "Failed to sign the .sha256sums file"
+		exit 1
+	fi
+
+	echo "Verifying signature..."
+	gpg --no-default-keyring --keyring $SigningKeyFilePath --verify  ./$WorkingDir/$ModuleName.asc ./$WorkingDir/$ModuleName.sha256sums
+	if [ $? != 0 ]; then
```
jeffaco: Same as above, please test to be sure this does what you want, or change to if; then form discussed above. Check all your `if [ $? ` with set -e, or change.

jeffaco:
Fix commits to be a single meaningful commit message, incorporate my other comments, test to be sure everything works (particularly check a few failure cases to make sure your 'if' testing is correct with `set -e`.

Do all that, and you're signed off.

When done, please e-mail me a log of EXACTLY how to run this puppy, and I'll look at incorporating it into Jenkins for nightly builds (assuming the build was successful, I won't do it if the build fails).

I presume you want this run under Jenkins for labeled builds as well?

build/mergeallmodules.sh
robbiezhang: no, account-specific will be fine.

build/mergeallmodules.sh
robbiezhang: the "gpg --list-secret-keys" will exit with non-zero if there is no key. I have to put it after it to avoid the command exit unexpected.

build/mergeallmodules.sh
robbiezhang: I tested with a bad passphrase. It will stop immediately. Here is the error:

You need a passphrase to unlock the secret key for
user: "Kristopher Bash (TestSignKey) <krbash@Microsoft.com>"
2048-bit RSA key, ID 8C3B51C6, created 2016-04-04

gpg: no default secret key: bad passphrase
gpg: signing failed: bad passphrase
Failed to sign the .sha256sums file

PR #72, opened 2016-04-21
Switch to single change tracking filter

niroyb:
This prevents a race condition calling DSC for services and package listings.
@Microsoft/omsagent-devs

### DIFF
--- a/source/code/plugins/changetracking_lib.rb
+++ b/source/code/plugins/changetracking_lib.rb
@@ -12,6 +12,8 @@ class ChangeTracking
         @@log = value
     end
 
+    @@force_send_last_upload = Time.now
+
     @@prev_hash = ""
     def self.prev_hash= (value)
         @@prev_hash = value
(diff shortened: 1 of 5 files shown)


source/code/plugins/changetracking_lib.rb
```diff
@@ -76,11 +78,15 @@ def self.isServiceInstanceXML(instanceXML)
...
     end
 
-    def self.transform_and_wrap(inventoryXMLstr, host, time, force_send = false)
+    def self.transform_and_wrap(inventoryXMLstr, host, time, force_send_run_interval = 0)
 
         # Do not send duplicate data if we are not forced to
         hash = Digest::SHA256.hexdigest(inventoryXMLstr)
-        if hash == @@prev_hash and force_send == false
+
+        if force_send_run_interval > 0 and Time.now - @@force_send_last_upload > force_send_run_interval
```
robbiezhang: same, use utc time

source/code/plugins/changetracking_lib.rb
```diff
@@ -12,6 +12,8 @@ def self.log= (value)
         @@log = value
     end
 
+    @@force_send_last_upload = Time.now
```
robbiezhang: suggest to use utc time
shall we persist the last upload time? otherwise, every restart will reset the time. The DSC configuration triggers a restart, whenever there is a change.

niroyb: This is just an internal variable. The time zone has no effect.
I'd rather have no state files but I see your concern. It is addressed since we send the data on each restart.

robbiezhang:
Signoff

PR #74, opened 2016-04-26
Add omsagent process stats to telemetry

niroyb:

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -45,6 +45,8 @@ RESP_HEARTBEAT=$TMP_DIR/resp_heartbeat.xml
 BODY_RENEW_CERT=$TMP_DIR/body_renew_cert.xml
 RESP_RENEW_CERT=$TMP_DIR/resp_renew_cert.xml
 
+PROCESS_STATS=$TMP_DIR/process_stats_$$
+
 AGENT_USER=omsagent
 AGENT_GROUP=omsagent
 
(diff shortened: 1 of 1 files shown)


installer/scripts/omsadmin.sh
```diff
@@ -275,6 +285,8 @@ append_telemetry()
     echo "      <ProcessorArchitecture>x64</ProcessorArchitecture>" >> $1
     echo "      <Version>$OSVersion</Version>" >> $1
     echo "      <InContainer>$InContainer</InContainer>" >> $1
+    echo "      <ProcessStats PercentUserTime=\"$PercentUserTime\" PercentPrivilegedTime=\"$PercentPrivilegedTime\" \
```
agup006: Can we change this to be a generic `<Telemetry>` tag?

installer/scripts/omsadmin.sh
```diff
@@ -45,6 +45,8 @@ RESP_HEARTBEAT=$TMP_DIR/resp_heartbeat.xml
 BODY_RENEW_CERT=$TMP_DIR/body_renew_cert.xml
 RESP_RENEW_CERT=$TMP_DIR/resp_renew_cert.xml
 
+PROCESS_STATS=$TMP_DIR/process_stats
```
jeffaco: I guess there's no way two instances can run at once? I'm wondering if you should make the filename $TMP_DIR/process_stats_$$ for uniqueness?

jeffaco:
Assuming the server accepts this data (obvioiusly!), signed off.

installer/scripts/omsadmin.sh
niroyb: Will do

installer/scripts/omsadmin.sh
niroyb: This will be run at 20 minute intervals. I still added the $$ for uniqueness because it can be run manually as well.

PR #78, opened 2016-05-02
Fixing missing files after upgrade issues

johnkord:
If some files are missing after upgrading successfully, then those files were removed as part of the postuninstall of the previous kit.  This change will make sure those files get recreated after upgrade.

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -799,6 +799,16 @@ case "$installMode" in
             shouldInstall_omsconfig
             pkg_upd $DSC_PKG omsconfig $?
             DSC_EXIT_STATUS=$?
+            
+            if [ $DSC_EXIT_STATUS -eq 0 ]; then
+                if [ ! -f /opt/microsoft/omsconfig/Scripts/2.6x-2.7x/Scripts/nxOMSAgent.py ]; then
+                    su - omsagent -c "/opt/microsoft/omsconfig/Scripts/InstallModule.py /opt/microsoft/omsconfig/module_packages/nxOMSAgent_1.0.zip 0"
+                fi
+                if [ ! -f /etc/opt/microsoft/omsagent/conf/omsagent.d/omsconfig.consistencyinvoker.conf ]; then
+                    cp -f /opt/microsoft/omsconfig/etc/omsconfig.consistencyinvoker.conf /etc/opt/microsoft/omsagent/conf/omsagent.d/omsconfig.consistencyinvoker.conf
+                    chown omsagent /etc/opt/microsoft/omsagent/conf/omsagent.d/omsconfig.consistencyinvoker.conf
+                fi
+            fi
         else
             DSC_EXIT_STATUS=0
         fi


installer/bundle/bundle_skel.sh
```diff
@@ -799,6 +799,16 @@ case "$installMode" in
             shouldInstall_omsconfig
             pkg_upd $DSC_PKG omsconfig $?
             DSC_EXIT_STATUS=$?
+            
+            if [ $DSC_EXIT_STATUS -eq 0 ]; then
+                if [ ! -f /opt/microsoft/omsconfig/Scripts/2.6x-2.7x/Scripts/nxOMSAgent.py ]; then
```
niroyb: You could define the hardcoded paths at the top of this file with the others.

niroyb:
I'm not sure this addresses the root cause.
Should upgrade succeed after an uninstall?
Should we install instead of upgrade if it was previously uninstalled?
Why is it just those two files that are missing?

jeffaco:
@niroyb 

I think this does address the root cause. You weren't privy to some hallway conversations about this.

The intent here is that this is run on an initial install (or upgrade if the files don't already exist). Once the files are there, they should stay there until removed.

On uninstall, these files are already removed as part of the omsconfig removal.

There are more than two files "missing", but we don't check all of them, we just check some "marker" files. If those are missing, then a bunch will be created.

jeffaco:
@johnkord 

I think you're missing some stuff here. Did you test this?

You need:

* Modify installbuilder to no longer stage these files at all,
* Modify the makefile to make sure the .zip file(s) are part of the tarball

Or did you pick a different approach?

johnkord:
@jeffaco 
Yes, I have tested it.

For point 1:
The py files are no longer being maintained by the package managers and has been this way for awhile.
The .conf file continues to be maintained by the package manager, which is copied into place during PostInstall. However there was a line in the previous package PostUninstall step that removed this copied file, which is called during upgrade.  This file should only be removed on a removal, not upgrade, and I submitted a pull request in omsconfig to change that yesterday.

For point 2:
The zip files are already included with the omsconfig package.

jeffaco:
@johnkord 

Okay, I obviously don't understand something. I think I must have misunderstood the whiteboard conversation. When I look at the shell bundle, I see:

```
-rw-r--r--. 1 jeffcof jeffcof  5175782 May  2 17:34 omi-1.0.8-5.universal.x64.deb
-rw-r--r--. 1 jeffcof jeffcof  5232892 May  2 17:34 omi-1.0.8-5.universal.x64.rpm
-rw-r--r--. 1 jeffcof jeffcof 12755242 May  2 17:34 omsagent-1.1.0-153.universal.x64.deb
-rw-r--r--. 1 jeffcof jeffcof 13507443 May  2 17:32 omsagent-1.1.0-153.universal.x64.rpm
-rw-r--r--. 1 jeffcof jeffcof  5017524 May  2 17:34 omsconfig-1.1.1-171.x64.deb
-rw-r--r--. 1 jeffcof jeffcof  5050386 May  2 17:34 omsconfig-1.1.1-171.x64.rpm
-rw-r--r--. 1 jeffcof jeffcof  1974478 May  2 17:34 scx-cimprov-1.6.2-281.universal.x64.deb
-rw-r--r--. 1 jeffcof jeffcof  2000939 May  2 17:34 scx-cimprov-1.6.2-281.universal.x64.rpm
```

So the .zip files may be part of the omsconfig package, but I thought that was the problem, and you didn't want them to be part of the package. So my understanding (obviously wrong?) was that they were going to be part of the tarball, and you'd just pick them up from there in the shell bundle, after the package was installed.

Looks like your OOF. Can you link up with me so I can just understand the design here? Thanks!

johnkord:
I've synced with Jeff on his issues with this, and I'm still awaiting signoff

jeffaco:
Um. Taking another look, this won't merge well with my changes in flight. Let me take a closer look, then stop by.

jeffaco:
Discussed F2F. This will have a problem with changes in flight, but John knows how to handle that. I'll let John know when my changes go in so he can handle that.

Signed off.

jeffaco:
I ended up merging this myself with John's help to reduce testing time by OMS team. This PR should be closed without committing.

PR #79, opened 2016-05-04
Fix extraneous message when makefile runs on a 32-bit platform

jeffaco:
Currently, the OMS OSS kits only contains 64-bit shell bundles.
Thus, no longer look for installer/oss-kits/*.i686.sh files since
they won't ever exist.

@Microsoft/ostc-devs


ericgable:
Looks good!  Sign-off.

PR #80, opened 2016-05-04
Change bundle to install/upgrade all packages as a set

jeffaco:
Also fix missing files for omsconfig after upgrade (this was an issue
where postuninstall was running after the current postinstall).

@Microsoft/ostc-devs


ericgable:
Looks good!  Sign-off.

sugr4:
Sign off

PR #81, opened 2016-05-05
Fix omsadmin system test to not require having omi installed

niroyb:
Telemetry uses omi to obtain teh info on the omsagent process. This change mocks the omi output so that it is not required to be installed for testing.

Before:
```
========================= Performing System Test
[...]
Started
./tmp/test_omsadmin_30225/omsadmin.sh: line 270: /opt/omi/bin/omicli: No such file or directory
./tmp/test_omsadmin_30305/omsadmin.sh: line 270: /opt/omi/bin/omicli: No such file or directory
./tmp/test_omsadmin_30387/omsadmin.sh: line 270: /opt/omi/bin/omicli: No such file or directory
./tmp/test_omsadmin_30474/omsadmin.sh: line 270: /opt/omi/bin/omicli: No such file or directory
../tmp/test_omsadmin_30578/omsadmin.sh: line 270: /opt/omi/bin/omicli: No such file or directory
/tmp/test_omsadmin_30578/omsadmin.sh: line 270: /opt/omi/bin/omicli: No such file or directory
./tmp/test_omsadmin_30727/omsadmin.sh: line 270: /opt/omi/bin/omicli: No such file or directory
/tmp/test_omsadmin_30727/omsadmin.sh: line 270: /opt/omi/bin/omicli: No such file or directory
./tmp/test_omsadmin_30880/omsadmin.sh: line 270: /opt/omi/bin/omicli: No such file or directory
..
Finished in 14.135203 seconds.
```
@Microsoft/omsagent-devs


test/installer/scripts/prep_omsadmin.sh
```diff
@@ -20,6 +20,19 @@ sed -i s,INSTALL_INFO=.*,INSTALL_INFO=$BASE_DIR/installer/conf/installinfo.txt,1
...
+# The output generated by omi is mocked in process_stats.
+sed -i s,/opt/omi/bin/omicli.*,,1 $OMSADMIN
+
+cat <<EOF > $TESTDIR/process_stats
+    PercentUserTime=2
+    PercentPrivilegedTime=1
+    UsedMemory=27712
+    PercentUsedMemory=3
+EOF
+
```
jeffaco: I'd be happier if you just included the complete output from an OMI enemeration. It's a better unit test than bastardized output just to make the test pass ...

jeffaco:
Signed off with the above in mind.

test/installer/scripts/prep_omsadmin.sh
niroyb: This is the exact output of the command mocked. Here it is to try it out:
```
/opt/omi/bin/omicli wql root/scx "SELECT PercentUserTime, PercentPrivilegedTime, UsedMemory, PercentUsedMemory FROM SCX_UnixProcessStatisticalInformation where Name like 'omsagent'" | grep =
```

jeffaco: I didn't realize that a SELECT was being done, I thought it was just an enumeration (which would result in very different output, but output that the grep should still work with).

Sorry for the confusion!

PR #85, opened 2016-05-14
Running omsagent as group omiusers

johnkord:
Setting the omsagent user to have the omiusers group, and silencing stderr messages from userdel that would sometimes warn that the group that was created to go with this user was not being deleted.

### DIFF
--- a/installer/scripts/omsagent.systemd
+++ b/installer/scripts/omsagent.systemd
@@ -6,7 +6,7 @@ Wants=omid.service
 [Service]
 Type=simple
 User=omsagent
-Group=omsagent
+Group=omiusers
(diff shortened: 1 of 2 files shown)


installer/scripts/omsagent.systemd
```diff
@@ -6,7 +6,7 @@ Wants=omid.service
 [Service]
 Type=simple
 User=omsagent
-Group=omsagent
+Group=omiusers
```
jeffaco: What sort of effect would that have if you're starting omsagent as an init.d service? I'd think you'd need to account for this change there as well?

Suggest you test on RH6 or SLES11 or something older than system to verify (if you didn't already).

jeffaco:
Verify behavior on init.d style systems. Otherwise, signed off.

PR #86, opened 2016-05-17
Restart omsagent service after upgrade

johnkord:
omsagent is not restarted after a successful upgrade, but it likely should be for future releases of the omsagent bundle (and definitely in this release of the bundle, as not restarting omsagent caused a bug when the omsconfig.consistencyinvoker.conf file was deleted and then immediately replaced after an upgrade).

However, I don't know the reasoning as to why omsagent wasn't previously restarted after an upgrade, perhaps it was intentional, so I'd like to get @jeffaco and @niroyb to review this!

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -801,6 +801,7 @@ case "$installMode" in
                 fi
             fi
             /opt/omi/bin/service_control restart
+            /opt/microsoft/omsagent/bin/service_control restart
         fi
 
         # Upgrade bundled providers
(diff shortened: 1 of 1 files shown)


jeffaco:
I do NOT sign off on this.

In general, we shouldn't do this sort of thing in the shell bundle, we should do this sort of thing in the installation scripts. We only resort to shell bundle behavior if a prior bug requires this for some reason. And indeed, omsagent itself is doing this during installation. It calls service_control with "enable", which will both enable and start/restart the omsagent server.

Given this, if the problem is that omsconfig is making configuration changes and should restart the omsagent server, that's why we have service_control for OMS: so it's very easy for omsconfig to do that.

johnkord:
Ping after commit to @jeffaco @niroyb

jeffaco:
Signed off

PR #87, opened 2016-05-18
Monitor agent health

varunkumta:
- Add standard monitor_agent input plugin (emit_config is commented out which is supported as part of https://github.com/fluent/fluentd/pull/963)
- Add filter plugin which will return a record when output plugin's buffer queue is more than 90% full
- Add new instance of out_oms plugin for oms.operation tag only
- Add unit test

Waiting on Anurag to provide field details (operationstatus, detail, category. solution, helplink)

@Microsoft/omsagent-devs

### DIFF
--- a/installer/conf/omsagent.d/monitor.conf
+++ b/installer/conf/omsagent.d/monitor.conf
@@ -1,0 +1,24 @@
+<source>
+  type monitor_agent
+  tag oms.operation
+  emit_interval 5m
+# emit_config true
+</source>
+
+<filter oms.operation.**>
+  type filter_operation
+</filter>
+
+<match oms.operation.**>
+  type out_oms
+  log_level info
+  buffer_chunk_limit 1m
+  buffer_type file
+  buffer_path /var/opt/microsoft/omsagent/state/out_oms_operation*.buffer
+  buffer_queue_limit 5
+  flush_interval 20s
+  retry_limit 10
+  retry_wait 30s
+  max_retry_wait 5m
+</match>
+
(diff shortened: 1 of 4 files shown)


installer/conf/omsagent.conf
```diff
@@ -89,6 +89,19 @@
   retry_wait 30s
 </match>
 
+<match oms.operation.**>
```
niroyb: This can go in monitor.conf as well.
Since the omsagent.d conf files are included at the top of omsagent.conf, `match oms.operation.**` will still come before the more generic `match oms.**`

installer/conf/omsagent.d/monitor.conf
```diff
@@ -0,0 +1,10 @@
+<source>
+  type monitor_agent
```
niroyb: Where is this input plugin?

installer/conf/omsagent.conf
```diff
@@ -89,6 +89,19 @@
...
+  type out_oms
+  log_level info
```
niroyb: Default log level is usually warning although there is nothing logged in `operation_lib.rb`

niroyb:
Signed off

installer/conf/omsagent.conf
varunkumta: updated

installer/conf/omsagent.conf
varunkumta: keeping it in line with the other out_oms* plugins

installer/conf/omsagent.d/monitor.conf
varunkumta: it's fluentd's standard input plugin .. fluentd-0.12.14/lib/fluent/plugin/in_monitor_agent.rb

PR #88, opened 2016-05-19
Fix exception type for buffer retry logic

niroyb:
It turns out that the fluentd engine does not handle custom exceptions properly.
By rethrowing the message, we are creating a Runtime exception instead of a RetryRequest Exception.
This is what fluend expects.
@Microsoft/omsagent-devs


robbiezhang:
Signoff

jeffaco:
Signed off

PR #89, opened 2016-05-23
Add --collectd option to omsadmin.sh

sugr4:
This allows us to set up collectd with OMS if the omsagent is already installed.
@Microsoft/omsagent-devs

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -5,6 +5,8 @@ set -e
 TMP_DIR=/var/opt/microsoft/omsagent/tmp
 CERT_DIR=/etc/opt/microsoft/omsagent/certs
 CONF_DIR=/etc/opt/microsoft/omsagent/conf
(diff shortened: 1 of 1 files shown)


installer/scripts/omsadmin.sh
```diff
@@ -582,6 +598,17 @@ renew_cert()
     fi
 }
 
+collectd()
+{
+    if [ -d "$COLLECTD_DIR" ]; then
+        cp $SYSCONF_DIR/omsagent.d/oms.conf $COLLECTD_DIR/oms.conf
+        cp $SYSCONF_DIR/omsagent.d/collectd.conf $CONF_DIR/omsagent.d/collectd.conf
```
niroyb: Make sure the permissions of the copied files are correct. Files in omsagent.d are owned by omsagent but files in sysconf dir are owned by root.
Note that if the omsagent agent is running when this code is executed then the plugin will only become active after a restart.

niroyb:
Signed off

PR #90, opened 2016-05-24
New CustomLog Metadata and PostBlobUploadNotification API

robbiezhang:
1. Add 2 new metadata to the CustomLog blob:
    ComputerName: the current computer name
    FilePath: the target file path API
2. Call the PostBlobUploadNotification API to notify
the that the blob is uploaded.
3. Add system test for out_oms_blob.rb plugin

@Microsoft/omsagent-devs


niroyb:
Signed off.

PR #92, opened 2016-05-25
Fix --collectd for omsdmin.sh for all other options to work correctly

sugr4:
Tested with 'make systemtest' and pbuild.
@Microsoft/omsagent-devs

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -80,7 +80,7 @@ usage()
     echo "$basename [-u user] -p host[:port]"
     echo
     echo "Enable collectd:"
-    echo "$basename --collectd"
+    echo "$basename -c"
 
     clean_exit 1
 }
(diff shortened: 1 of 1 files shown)


installer/scripts/omsadmin.sh
```diff
@@ -170,10 +170,11 @@ parse_args()
...
     for arg in "$@"; do
-        if [ "$arg" = "--collectd" ]; then
-          set -- "$@" "-c"
-        fi
-        shift;
+        shift
+        case "$arg" in
+          "--collectd") set -- "$@" "-c" ;;
+          *)            set -- "$@" "$arg" ;;
+        esac
```
jeffaco: I'm not going to give signoff on this, sorry.

First, I'm not 100% sure what `set --` does, nor is Nick. But that's not that important.

What is important is that this script takes a bunch of different parameters, and NONE need to be special cased this way.

I'd like to see one of two things done:

1. Eliminate this code fragment and just accept `-c` for the collectd function. This allows consistency with all of the other parameters, or
2. Refactor the parameter parsing (like the OMS build/configure script) to allow both shorter and longer parameters. In this way, if you REALLY want to support `--collectd`, you can.

Personally, I'd just do the first - I think that's fine. But I'll leave it up to you.

jeffaco:
NOT signed off, sorry. See above.

jeffaco:
Somehow I missed that there was a new iteration of this. Sorry about that, but GitHub didn't send me E-Mail.

This looks fine, signed off.

PR #94, opened 2016-05-27
Changes for building local and ULINUX on Debian/Ubuntu:

ericgable:
configure - Improve non-ULINUX error message that occurs when the ostc-open-ssl are installed
due to previous ULINUX build.  Explain that removing the ostc-open-ssl will allow a local non-ULINUX build.

Makefile - Add SHELL=/bin/bash.  Debian variants default to /bin/sh, which breaks brace expansion.

@jeffaco


build/configure
```diff
@@ -233,6 +233,8 @@ if [ $ULINUX -eq 0 -a -d /usr/local_ssl_0.9.8 -a -d /usr/local_ssl_1.0.0 ]; then
     echo "Given that OMI makes assumptions based on these directories, there is an" >& 2
     echo "extraordinarily good chance that your build will fail. You should do a" >& 2
     echo "ULINUX build (--enable-ulinux) on this system." >& 2
+    echo "If you wish to continue with a non ULINUX build remove " >& 2
+    echo "/usr/local_ssl_0.9.8 and /usr/local_ssl_1.0.0, then re-run configure." >& 2
```
jeffaco: Suggest you replace these two lines with something like:

```
echo "" >& 2
echo "If you really do not want a universal (ULINUX) build, then run:" >& 2
echo "    sudo rm -rf /usr/local_ssl_0.9.8 /usr/local_ssl_1.0.0" >& 2
echo "and then re-run configure. IMPORTANT: Only try this on a private system," >& 2
echo "Do NOT do this on a shared system without consulting your team!" >& 2
```

jeffaco:
Signed off with above change.

Note: You have a bunch of other changes in various repositories to address Ubuntu builds, some of which I rejected. If this fixes your problem, could you delete the old branches from other repositories (i.e. PAL, etc)?

ericgable:
Agree - will fix in next commit.

PR #95, opened 2016-05-28
Update to fluentd v0.12.24

varunkumta:
- update to fluentd's commit hash of 12.24
- update the fluentd gems in source/ext/patches/fluentd
- update ruby.data installer file for new fluentd gems
- update buildRuby.sh to install 12.24.gem

- update conf to use new features in 12.24 (drop_oldest_chunk and emit_config)
- update base_omsagent.data to drop conf and plugins when bundle is installed

Verified build on CentOS 5.0 
Verified by installing and onboarding the agent on Ubuntu and RedHat .. verified that syslog, perf data is sent up, syslog and perf config is pulled down on both distros

@jeffaco According to fluentd GitHub page sub-second precision is not part of v0.12. Also, kill -sigterm does not work on the RedHat system - it fails to shutdown the last i.e. 4th oms_omi input plugin - and stops it only after 5 minutes - probably a timeout fluentd enforces
@niroyb 
@Microsoft/omsagent-devs


jeffaco:
I really thought that the next version of FluentD fixed the timeout problems. That's what the PM from Treasure Data had to say. We might want to follow up with them to see what happened.

These changes look fine, but some of the above comments are useful enough to consider doing to `git commit --amend' and adding to the actual commit message (i.e. multi-line commit), like:

```
Update to fluentd v0.12.24 

- update to fluentd's commit hash of 12.24
- update the fluentd gems in source/ext/patches/fluentd
- update ruby.data installer file for new fluentd gems
- update buildRuby.sh to install 12.24.gem

- update conf to use new features in 12.24 (drop_oldest_chunk and emit_config)
- update base_omsagent.data to drop conf and plugins when bundle is installed

Verified build on CentOS 5.0 
```

Multi-line comments are good if they add usefulness, and that one in particular is a good footprint for how to do this again down the road! :smile:

Signed off.

niroyb:
Signed off

PR #96, opened 2016-06-04
ChangeTracking config call PerformInventory.py with params

robbiezhang:
Update the ChangeTracking configuration to call the
PerformInventory.py with --InMOF and --OutXML parameters.
Add the ChangeTracking inventory mof into the installer.
@Microsoft/omsagent-devs @johnkord @KrisBash

### DIFF
--- a/installer/conf/omsagent.d/change_tracking.conf
+++ b/installer/conf/omsagent.d/change_tracking.conf
@@ -1,7 +1,7 @@
 <source>
   type exec
   tag oms.changetracking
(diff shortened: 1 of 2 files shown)


niroyb:
Signed off. Is the mof file supposed to be part of the package or is it just there for reference?

robbiezhang:
Same strategy as the conf file. I'll include the MOF in the nxOMSPlugin module, so that it will be copied into the omsagent.d folder when the customer enables the ChangeTracking solution

PR #108, opened 2016-06-16
Update the default set of Performance Counters

robbiezhang:
@Microsoft/omsagent-devs @KrisBash 
To be consistent with Windows load,
only collect the following counters by default.

Logical Disk
  % Used Inodes
  Free Megabytes
  % Used Space
  Disk Transfers/sec
  Disk Reads/sec
  Disk Writes/sec

Processor
  % Processor Time
  % Privileged Time

Memory
  Available MBytes Memory
  % Used Memory
  % Used Swap Space

### DIFF
--- a/installer/conf/omsagent.conf
+++ b/installer/conf/omsagent.conf
@@ -14,17 +14,9 @@
 
 <source>
   type oms_omi
-  object_name "Physical Disk"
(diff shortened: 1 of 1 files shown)


niroyb:
Signed off

PR #113, opened 2016-06-20
Fixed the patch-scx unittest dependency

vrdmr:
Added new parameters in all the methods
to make this more test friendly, by accepting,
mocks.

The test was done with removing the 
scx-release file as well as by porting the 
plugin to a live system and making sure it 
works in the normal cases, when the test
parameters are all nil.

### DIFF
--- a/source/code/plugins/patch_management_lib.rb
+++ b/source/code/plugins/patch_management_lib.rb
@@ -87,6 +87,14 @@ class LinuxUpdates
         @@prev_hash = value
     end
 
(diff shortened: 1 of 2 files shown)


vrdmr:
Please review
@Microsoft/omsagent-devs @robbiezhang 
@niroyb @jeffaco @KrisBash

robbiezhang:
Signoff

vrdmr:
Thanks @robbiezhang.

PR #114, opened 2016-06-21
Nit change: Missed a optional parameter

vrdmr:
I missed parameter 4 in the earlier checking
which caused the build to fail. Corrected now.


robbiezhang:
Signoff

PR #116, opened 2016-06-23
Update the ruby rev to 2.2.0b

robbiezhang:
@Microsoft/omsagent-devs @varunkumta @jeffaco


jeffaco:
That's not going to be particularly useful, except to break the build. This is good (to force a new build of Ruby), but it needs to be paired with appropriate changes to buildRuby.sh so the unit tests run properly.

My suggestion is to camp out on one of the DEV systems and get this working properly. When on the dev system, you can just do a:

```
sudo rm -rf /usr/local/ruby-2.2.0a
```

Then, if you build, you can verify the unit tests run properly. For purposes of the test, I would hack Makefile to not build the RPM and DPKG kits since (1) that takes a long time, and (2) you don't need that to run the unit tests.

After the tests are running properly, then verify that the kit we build will still install. If not, then you'll need to rebuild ruby.data (using the shell script I mentioned in E-Mail) to capture the changes in files.

Note: If you go through this process, you probably want to consider building with shared libraries, as that will allow other uses that are coming up that I didn't originally anticipate. That will result in a slightly larger footprint (and more memory usage), but not if other consumers start using our instance of Ruby.

So, as is, I do not sign off - this will only serve to break the build. You need to couple this change with the other changes outlined here.

jeffaco:
Note: I just pushed branch jeff-omsbuild to project pbuild. If you switch to this branch and run pbuild, then pbuild will run OMS unit tests at completion of the build. This, coupled with a change to the directory name for ruby in /usr/local, will allow you to verify that the build will succeed.

Then verify that the generated kit installs and runs properly. THEN you're good for checkin.

Once this is fixed up, I'll commit the change to the pbuild project.

varunkumta:
you will also need to re-generate the ruby.data file using genRubyInstaller.sh

jeffaco:
This looks fine to me. Please do some basic install verification (try on CentOS 7 and Ubuntu 14 or 16 minimally). If it installs and starts properly, I sign off.

robbiezhang:
verified the bundle installation and basic functionality in CentOS 7 x64 & Ubuntu 14 x64.

PR #117, opened 2016-06-23
Add new scripts OMS(Yum/Zypper)Updates.sh to sudoers.

ericgable:
The companion commit for this in DSC has been made.
@KrisBash 
@johnkord


jeffaco:
This looks reasonable to me. But I'd definitely like @KrisBash to look at this and make sure it jives with your DSC changes.

KrisBash:
Looks good to me

PR #118, opened 2016-06-24
Bug fixes in the oms_common.rb

robbiezhang:
1. Fix the nil class issue in parsing the timezone for symlink using relative file path
2. Fix the warning in compiling that the error is rescued, but not used.

@Microsoft/omsagent-devs

### DIFF
--- a/test/code/plugins/oms_common_test.rb
+++ b/test/code/plugins/oms_common_test.rb
@@ -345,5 +345,16 @@ module OMS
       # sanity check
       assert_equal('Pacific Standard Time', timezone, "timezone is not expected")
     end
+    
+    def test_get_current_timezone_relative_symlink
+      $log = MockLog.new
+      File.delete(@tmp_localtime_file.path)
+      File.symlink('/etc/../usr/share/zoneinfo/right/America/Los_Angeles', @tmp_localtime_file.path)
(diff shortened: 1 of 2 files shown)


robbiezhang:
@Microsoft/omsagent-devs

test/code/plugins/oms_common_test.rb
```diff
@@ -345,5 +345,16 @@ def test_get_current_timezone_symlink
       # sanity check
       assert_equal('Pacific Standard Time', timezone, "timezone is not expected")
     end
+    
+    def test_get_current_timezone_relative_symlink
+      $log = MockLog.new
+      File.delete(@tmp_localtime_file.path)
+      File.symlink('/usr/../usr/share/zoneinfo/right/America/Los_Angeles', @tmp_localtime_file.path)
```
jeffaco: You don't need /usr/../usr/share/zoneinfo/right/America/Los_Angelese

Take off the leading `/usr/..`, as it serves no (obvious) purpose

jeffaco: I guess that seems to be a valid path on CentOS, SuSE, and Ubuntu ...

jeffaco:
Signed off

PR #121, opened 2016-06-30
Add security filter files to base_omsagent.data

mgladi:


varunkumta:
How are this and the existing syslog filter plugin related? Are they mutually exclusive i.e. only one of them will be active at a time? You will have to add an entry to omsagent.conf so that this plugin is executed in the fluentd pipeline.

Otherwise, Signed off.

mgladi:
This and the syslog filter plugin are not mutually exclusive. Although they are similar they have different purposes. This plugin is meant for collecting security events from 3rd party tools (i.e. firewalls, vulnerability scanner, etc.).
I'm aware that configuration is required to enable the plugin. It is coming soon.

PR #125, opened 2016-07-05
Removing installed packages info and unittest fixes

vrdmr:
Changed the way to use ruby rather than using `cat` to read the 
release file and host details. Also, cleaned the way shortname is created
to help the unit-test now pass the os name and version for testing.

### DIFF
--- a/source/code/plugins/patch_management_lib.rb
+++ b/source/code/plugins/patch_management_lib.rb
@@ -12,28 +12,65 @@ class LinuxUpdates
     @@delimiter = "_"
     @@force_send_last_upload = Time.now
     MAJOR_MINOR_VERSION_REGEX = /([^\.]+)\.([^\.]+).*/
+    OMS_ADMIN_FILE = "/etc/opt/microsoft/omsagent/conf/omsadmin.conf"
+    SCX_RELEASE_FILE = "/etc/opt/microsoft/scx/conf/scx-release"
 
-    @@agentDetailsMap = Hash[(`cat /etc/opt/microsoft/omsagent/conf/omsadmin.conf`.split(/\r?\n/).map {|s| 
-                                s.split("=")}).map {|key, value| 
-                                    [key, value]
-                                }
-                            ]
+    def self.getAgentDetails()
+        ret = {}
 
-    @@hostOSDetailsMap = Hash[(`cat /etc/opt/microsoft/scx/conf/scx-release`.split(/\r?\n/).map {|s|
-                               s.split("=")}).map {|key, value|
-                                    [key, value]
-                                }
-                            ]
-    
-    @os_major_version = @@hostOSDetailsMap["OSVersion"][MAJOR_MINOR_VERSION_REGEX, 1] unless @@hostOSDetailsMap["OSVersion"].nil?
-    @os_minor_version = @@hostOSDetailsMap["OSVersion"][MAJOR_MINOR_VERSION_REGEX, 2] unless @@hostOSDetailsMap["OSVersion"].nil?
+        if File.exist?(OMS_ADMIN_FILE) # If file exists
+            File.open(OMS_ADMIN_FILE, "r") do |f| # Open file
+                f.each_line do |line|       # Split each line and
+                    line.split(/\r?\n/).reject{ |l|  # reject all those line
+                        !l.include? "=" }.map {|s|    # which do not
+                            s.split("=")}.map {|key, value| # have "="; split
+                                ret[key] = value     # by "=" and add to ret.
+                        }
+                end
+            end
+        else
+            @@log.debug "Could not find the file #{OMS_ADMIN_FILE}"
+        end
+        return ret
+    end
+
+    def self.getHostOSDetails()
+        ret = {}
+
+        if File.exist?(SCX_RELEASE_FILE) # If file exists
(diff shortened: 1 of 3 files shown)


vrdmr:
@Microsoft/omsagent-devs @robbiezhang @niroyb @ericgable @KrisBash Please review.

source/code/plugins/patch_management_lib.rb
```diff
@@ -12,28 +12,48 @@ class LinuxUpdates
...
-    def self.getOSShortName()
-        # match string of the form (1 or more non . chars)- followed by a . - (1 or more non . chars) - followed by anything
+    def self.getOSShortName(os_short_name = nil, os_version=nil)
         version = ""
 
-        case @@hostOSDetailsMap["OSShortName"].split("_")[0]
+        # match string of the form (1 or more non . chars)- followed by a . - (1 or more non . chars) - followed by anything
+        osName = (os_short_name == nil) ? @@hostOSDetailsMap["OSShortName"].split("_")[0] : os_short_name.split("_")[0]
+        if (os_version == nil)
+            @os_major_version = @@hostOSDetailsMap["OSVersion"][MAJOR_MINOR_VERSION_REGEX, 1] unless @@hostOSDetailsMap["OSVersion"].nil?
```
niroyb: Should os_version be set in this block?

source/code/plugins/patch_management_lib.rb
```diff
@@ -12,28 +12,48 @@ class LinuxUpdates
...
+
+    @@hostOSDetailsMap = if File.exist?(SCX_RELEASE_FILE)
+                            Hash[ File.open(SCX_RELEASE_FILE, "r") do |f|
+                                    f.each_line do |line|
+                                        line.split(/\r?\n/).map {|s| s.split("=")}.map {|key, value| [key, value]}
+                                    end
+                                end ]
+                        end
+
+    # @@hostOSDetailsMap = Hash[(`cat /etc/opt/microsoft/scx/conf/scx-release`.split(/\r?\n/).map {|s|
```
niroyb: Remove dead code.

source/code/plugins/patch_management_lib.rb
```diff
@@ -12,28 +12,48 @@ class LinuxUpdates
...
+                                    f.each_line do |line|
+                                        line.split(/\r?\n/).map {|s| s.split("=")}.map {|key, value| [key, value]}
+                                    end
+                                end ]
+                        end
+
+    @@hostOSDetailsMap = if File.exist?(SCX_RELEASE_FILE)
+                            Hash[ File.open(SCX_RELEASE_FILE, "r") do |f|
+                                    f.each_line do |line|
+                                        line.split(/\r?\n/).map {|s| s.split("=")}.map {|key, value| [key, value]}
```
niroyb: You may want to do error handling : if the file is not readable, if the line does not contain "="

test/code/plugins/patch_management_lib_test.rb
```diff
@@ -204,25 +204,25 @@ def test_available_updates_transform_and_wrap
...
+                    "Repository" => "Ubuntu:15.04/xenial-updates",
                     "Timestamp" => "1970-01-01T00:00:00.000Z"
                 }]
             }]
         }
     expectedTime = Time.utc(2016,3,15,19,2,38.5776)
     wrappedHash = LinuxUpdates::transform_and_wrap(@available_updates_xml_str, "HostName", expectedTime,
-                                                   86400, "AgentId-123", "Ubuntu", "Ubuntu 16.04",
-                                                   "16.04", "Ubuntu_16.04")
+                                                   86400, "AgentId-123", "Ubuntu", "Ubuntu 15.04",
```
niroyb: The dependency injection pattern would be worth to take a look at for future testing needs.

niroyb:
Signed off

source/code/plugins/patch_management_lib.rb
vrdmr: Done.

source/code/plugins/patch_management_lib.rb
vrdmr: os_version is being passed to the function. If it does not get anything, then it'll try to find the file (go and see if the file exists.) The version setting happens in the later parts where some flooring is happening.

test/code/plugins/patch_management_lib_test.rb
vrdmr: I'll investigate the ruby DI frameworks and see how to integrate with our case.

source/code/plugins/patch_management_lib.rb
vrdmr: Yes. Adding the checks now. Also, adding the filter.

vrdmr: Done. Added two separate functions rather than this complecated code to create a dictionary.

source/code/plugins/patch_management_lib.rb
vrdmr: Got the individual functions - `availableUpdatesXMLtoHash` and `installedPackageXMLtoHash` - to now compusarily take the os_short_name rather than rely on the function `getOSShortName` themselves. This would reduce the unnecesary calls to find OS shortnames.

vrdmr:
Fixed as per the comments. Tested and verified.

Merging.

PR #127, opened 2016-07-06
Set syslog event time

varunkumta:
Today we populate the syslog event time (in the workflow) from the TimeGenerated. Instead, add a new field with the actual event time (rounded to second - best that fluent v0.12 can do) and populate it from that.

@Microsoft/omsagent-devs 
@mgladi

### DIFF
--- a/source/code/plugins/filter_syslog.rb
+++ b/source/code/plugins/filter_syslog.rb
@@ -30,6 +30,7 @@ module Fluent
       # Use Time.now, because it is the only way to get subsecond precision in version 0.12.
       # The time may be slightly in the future from the ingestion time.
       record["Timestamp"] = OMS::Common.format_time(Time.now.to_f)
+      record["EventTime"] = OMS::Common.format_time(time)
       hostname = record["host"]
       record["Host"] = hostname
       record.delete "host"
--- a/source/code/plugins/filter_syslog_security.rb
+++ b/source/code/plugins/filter_syslog_security.rb
@@ -35,6 +35,7 @@ module Fluent
       # Use Time.now, because it is the only way to get subsecond precision in version 0.12.
       # The time may be slightly in the future from the ingestion time.
       record['Timestamp'] = OMS::Common.format_time(Time.now.to_f)
+      record['EventTime'] = OMS::Common.format_time(time)
 
       record['Message'] = record['message']
       record.delete 'message'


source/code/plugins/filter_syslog.rb
```diff
@@ -30,6 +30,7 @@ def filter(tag, time, record)
       # Use Time.now, because it is the only way to get subsecond precision in version 0.12.
       # The time may be slightly in the future from the ingestion time.
       record["Timestamp"] = OMS::Common.format_time(Time.now.to_f)
+      record["EventTime"] = time
```
mgladi: Why not format the time similar to the line above?

mgladi: What's the timezone of the value that's sent to the filter? Is local time?

varunkumta: formatting time using Common.format_time now .. the time that comes to the filter is the unix time (i.e. seconds since epoch) - so it is UTC time

source/code/plugins/filter_syslog.rb
```diff
@@ -30,6 +30,7 @@ def filter(tag, time, record)
       # Use Time.now, because it is the only way to get subsecond precision in version 0.12.
       # The time may be slightly in the future from the ingestion time.
       record["Timestamp"] = OMS::Common.format_time(Time.now.to_f)
+      record["EventTime"] = OMS::Common.format_time(time)
```
mgladi: I did some tests on my machine. It seems that FluentD parses the event time from syslog as being in UTC instead of in local time. The result is that, after formatting, 'EventTime' and 'TimeStamp' are a few hours apart.

For example:
"Timestamp":"2016-07-08T08:47:54.263Z",
"EventTime":"2016-07-08T15:32:45.000Z"

niroyb: The reason we use Time.now.to_f is that if two identical syslog events happen in the same second, ODS will only show one. Also, different events within the same second would appear out of order in the OMS portal...  @robbiezhang could give more details but sub second precision is very important. I'm not signing off

niroyb:
Signed off

source/code/plugins/filter_syslog.rb
varunkumta: Can you provide me access to the machines? I don't see how this could happen. The 'time' parameter comes from the raw syslog message (rounded to second) and is the unix/epoch time integer - so inherently UTC. So both, Time.now and 'time' are in UTC - I don't see how 'time' could be later/after Time.now.

varunkumta: Correction, Time.now.to_f and 'time' are both unix/epoch times - so in UTC

mgladi: After a short investigation it seems that the issue is caused by how Fluentd formats the timestamp from syslog. The Fluentd syslog parser assumes the timestamp is in UTC whereas in reality it might be in a different time zone. 
Time.now.to_f and 'time' are indeed both epoch times but one was taken from the actual local time (and time zone) and the other was parsed (possibly incorrectly) from a string.

I sent you the details of my machine.

varunkumta: I checked on the machine you shared with me. The /var/log/syslog file was actually logging UTC timestamps, whereas Time.strptime in ruby - which is what fluentd uses to parse the string assumes local time. So, for example, the syslog file says 19:05 - which is the actual UTC time but strptime assumes it is Asia/Jerusalem time (/etc/timezone) and makes the UTC time as 16:05. I restarted the syslog daemon (sudo service rsyslogd restart) and now it is logging local time and fluentd's timestamp is also fixed. 

So, I guess if you change the timezone on the machine you need to restart the syslog daemon

varunkumta: Typo: sudo service rsyslog restart

mgladi: That makes it clearer. Thanks for the investigation.

There's still a potential issue with syslog messages that are forwarded from another machine with a different time zone. It seems that the syslog timestamp is not adjusted to the receiving machine time zone in this case.
I sent you the details of the machine if you'd like to take a look.

mgladi:
Signed off

source/code/plugins/filter_syslog.rb
varunkumta: Fluentd supports syslog forwarding, so I would think that the fluentd config or syslog config would have to handle that

PR #130, opened 2016-07-12
Syslog security filter: add 'ident' value to logged message

mgladi:


source/code/plugins/filter_syslog_security.rb
```diff
@@ -36,7 +36,7 @@ def filter(tag, time, record)
       # The time may be slightly in the future from the ingestion time.
       record['Timestamp'] = OMS::Common.format_time(Time.now.to_f)
 
-      record['Message'] = record['message']
+      record['Message'] = record['ident'] + ': ' + record['message']
       record.delete 'message'
       record.delete 'time'
```
robbiezhang: delete 'ident' too.

robbiezhang:
Signoff

source/code/plugins/filter_syslog_security.rb
varunkumta: Can you explain why you need to do this? The 'ident' field is ingested into search as the 'ProcessName' field

mgladi: We are parsing a subset of the messages sent by the agent in the workflow. The parsing extracts relevant fields from the message (source/dest ip, protocol, action etc.). For messages that are not parsed we'd like to provide the full message to the users. 
There is no ProcessName field in the type to which the messages are mapped (CommonSecurityLog)

The 'ident' field is also required

PR #133, opened 2016-07-18
Change field name from Device to Computer in heartbeat data

varunkumta:
The heartbeat workflow now expects the machine name in the 'Computer' field, not 'Device' field


robbiezhang:
@Microsoft/omsagent-devs

robbiezhang:
Signoff.

PR #135, opened 2016-07-22
Changing the data frequency for Updates and other changes.

vrdmr:
Changed the following:
1. Changing the run_interval to 3h, and removed the forcefully sending.
2. Adding patch_management_immediate_run.conf file which is a copy
of the patch_management.conf without run_interval to start as soon as
agent starts. Different tag used for the same.
3. If _BuildDate_ and _InstalledOn_ is either nil or **BuildDate**, **InstalelledOn**, 
then send nil.
4. Also made changes in the tests.

@Microsoft/omsagent-devs @stasku Could you please review the changes?

### DIFF
--- a/source/code/plugins/patch_management_lib.rb
+++ b/source/code/plugins/patch_management_lib.rb
@@ -150,9 +150,13 @@ class LinuxUpdates
         ret["CollectionName"] = availableUpdatesHash["Name"] + @@delimiter + 
                                 availableUpdatesHash["Version"] + @@delimiter + os_short_name
         ret["PackageName"] = availableUpdatesHash["Name"]
+        ret["Architecture"] = availableUpdatesHash.key?("Architecture") ? availableUpdatesHash["Architecture"] : nil
         ret["PackageVersion"] = availableUpdatesHash["Version"]
-        ret["Timestamp"] = OMS::Common.format_time(availableUpdatesHash["BuildDate"].to_i)
-        ret["Repository"] = availableUpdatesHash.key?("Repository") ? availableUpdatesHash["Repository"] : ""
+        ret["Timestamp"] = (availableUpdatesHash["BuildDate"].nil? or 
+                            availableUpdatesHash["BuildDate"].strip == "" or
+                            availableUpdatesHash["BuildDate"].strip == "BuildDate") ? nil :
+                                OMS::Common.format_time(availableUpdatesHash["BuildDate"].to_i)
+        ret["Repository"] = availableUpdatesHash.key?("Repository") ? availableUpdatesHash["Repository"] : nil
         ret["Installed"] = false
         ret
     end
@@ -164,10 +168,14 @@ class LinuxUpdates
         ret["CollectionName"] = packageHash["Name"] + @@delimiter + 
                                 packageHash["Version"] + @@delimiter + os_short_name
         ret["PackageName"] = packageHash["Name"]
-        ret["PackageVersion"] = packageHash["Version"]
-        ret["Timestamp"] = OMS::Common.format_time(packageHash["InstalledOn"].to_i)
+        ret["Architecture"] = packageHash.key?("Architecture") ? packageHash["Architecture"] : nil
+        ret["PackageVersion"] = packageHash["Version"]  
+        ret["Timestamp"] = (packageHash["InstalledOn"].nil? or 
+                            packageHash["InstalledOn"].strip == "" or
+                            packageHash["InstalledOn"].strip == "Unknown") ? nil : 
+                                OMS::Common.format_time(packageHash["InstalledOn"].to_i)
         ret["Size"] = packageHash["Size"]
-        ret["Repository"] = packageHash.key?("Repository") ? packageHash["Repository"] : "" 
+        ret["Repository"] = packageHash.key?("Repository") ? packageHash["Repository"] : nil
         ret["Installed"] = true
         ret
     end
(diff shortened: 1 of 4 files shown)


installer/conf/omsagent.d/patch_management_immediate_run.conf
```diff
@@ -0,0 +1,12 @@
+<source>
+  type exec
+  tag oms.patch_management_immediate_run
+  command /opt/microsoft/omsconfig/Scripts/PerformInventory.py --InMOF /etc/opt/microsoft/omsagent/conf/omsagent.d/patch_management_inventory.mof --OutXML /etc/opt/omi/conf/omsconfig/configuration/CompletePackageInventory.xml > /dev/null && cat /etc/opt/omi/conf/omsconfig/configuration/CompletePackageInventory.xml
```
robbiezhang: is it by purpose to let it run when agent restarts? do you need to distinguish the 2 types of data in the service side?

will it cause race condition with the scheduled one, as they use the same output file?

source/code/plugins/patch_management_lib.rb
```diff
@@ -150,8 +150,11 @@ def self.availableUpdatesXMLtoHash(availableUpdatesXML, os_short_name)
         ret["CollectionName"] = availableUpdatesHash["Name"] + @@delimiter + 
                                 availableUpdatesHash["Version"] + @@delimiter + os_short_name
         ret["PackageName"] = availableUpdatesHash["Name"]
+        ret["Architecture"] = availableUpdatesHash.key?("Architecture") ? availableUpdatesHash["Architecture"] : nil
         ret["PackageVersion"] = availableUpdatesHash["Version"]
-        ret["Timestamp"] = OMS::Common.format_time(availableUpdatesHash["BuildDate"].to_i)
+        ret["Timestamp"] = (availableUpdatesHash["BuildDate"] == nil or
```
robbiezhang: use .nil? to check nil

source/code/plugins/patch_management_lib.rb
```diff
@@ -164,8 +167,11 @@ def self.installedPackageXMLtoHash(packageXML, os_short_name)
         ret["CollectionName"] = packageHash["Name"] + @@delimiter + 
                                 packageHash["Version"] + @@delimiter + os_short_name
         ret["PackageName"] = packageHash["Name"]
+        ret["Architecture"] = packageHash.key?("Architecture") ? packageHash["Architecture"] : nil
         ret["PackageVersion"] = packageHash["Version"]
-        ret["Timestamp"] = OMS::Common.format_time(packageHash["InstalledOn"].to_i)
+        ret["Timestamp"] = (packageHash["InstalledOn"] == nil or
```
robbiezhang: use .nil?

robbiezhang:
Add test for BuildDate and InstalledOn

installer/conf/omsagent.d/patch_management_immediate_run.conf
vrdmr: Yes, as out run_interval is now set to 3 hours (earlier, it was 12 hours), the users do not see any data as soon as they onboard. From fluentd forums, the preferred way was to crontab the python separately and use in_tail (https://groups.google.com/forum/#!topic/fluentd/Yq3BRYxYQuI), but then we lose the flexibility to restart if anything happens to the python script as well as the installers then woudl be changed, which was not feasible.

The data is the same from oms.patch_management_immediate_run and oms.patch_management. Also, the oms.patch_management_immediate_run should run only once based on the fluentd code.

source/code/plugins/patch_management_lib.rb
vrdmr: Yes. I'll do that.

source/code/plugins/patch_management_lib.rb
vrdmr: Yes.

vrdmr:
Added extra test-cases for nil and correct BuildDate/InstalledOn. The unittests are now added and works 100% successful.

source/code/plugins/patch_management_lib.rb
```diff
@@ -164,10 +168,13 @@ def self.installedPackageXMLtoHash(packageXML, os_short_name)
         ret["CollectionName"] = packageHash["Name"] + @@delimiter + 
                                 packageHash["Version"] + @@delimiter + os_short_name
         ret["PackageName"] = packageHash["Name"]
-        ret["PackageVersion"] = packageHash["Version"]
-        ret["Timestamp"] = OMS::Common.format_time(packageHash["InstalledOn"].to_i)
+        ret["Architecture"] = packageHash.key?("Architecture") ? packageHash["Architecture"] : nil
+        ret["PackageVersion"] = packageHash["Version"]  
+        ret["Timestamp"] = (packageHash["InstalledOn"].strip.nil? or
```
robbiezhang: you should not call .strip here, since it's a nil check

source/code/plugins/patch_management_lib.rb
```diff
@@ -150,9 +150,13 @@ def self.availableUpdatesXMLtoHash(availableUpdatesXML, os_short_name)
         ret["CollectionName"] = availableUpdatesHash["Name"] + @@delimiter + 
                                 availableUpdatesHash["Version"] + @@delimiter + os_short_name
         ret["PackageName"] = availableUpdatesHash["Name"]
+        ret["Architecture"] = availableUpdatesHash.key?("Architecture") ? availableUpdatesHash["Architecture"] : nil
         ret["PackageVersion"] = availableUpdatesHash["Version"]
-        ret["Timestamp"] = OMS::Common.format_time(availableUpdatesHash["BuildDate"].to_i)
-        ret["Repository"] = availableUpdatesHash.key?("Repository") ? availableUpdatesHash["Repository"] : ""
+        ret["Timestamp"] = (availableUpdatesHash["BuildDate"].strip == nil or
```
robbiezhang: availableUpdatesHash["BuildDate"].nil?

vrdmr:
Changes done.

robbiezhang:
Signoff

vrdmr:
Merged the two patch management related confs into one. Tested. Now merging the changes.

PR #137, opened 2016-07-25
Use computer name instead of fqdn for heartbeat

varunkumta:
Using fqdn for heartbeat but computer name for other types causes the same computer to appear twice when grouping by computer name. Switch heartbeat to use computer name.


varunkumta:
@Microsoft/omsagent-devs

robbiezhang:
Signoff

PR #139, opened 2016-07-26
AuditLog parser

robbiezhang:
Create auditlog_lib.rb as a utility class for audit log related operations.
Create parser_auditlog.rb as FluentD parser plugin to convert the audit log into json object
Add unit tests for the utility class
@Microsoft/omsagent-devs

### DIFF
--- a/source/code/plugins/auditlog_lib.rb
+++ b/source/code/plugins/auditlog_lib.rb
@@ -1,0 +1,161 @@
+module AuditLogModule
+  class LoggingBase
+    def log_error(text)
+    end
+  end
+
+  class RuntimeError < LoggingBase
+    def log_error(text)
+      $log.error "RuntimeError: #{text}"
+    end
+  end
+
+  class AuditLogParser
+    require 'date'
+    require 'etc'
+    require 'digest'
+    require_relative 'oms_common'
+
+    def initialize(error_handler)
+      @error_handler = error_handler
+      @sha256 = Digest::SHA256.new
+    end
+
+    SKETCH_REGEX = /type=(?<record>[A-Z_]+) msg=audit\((?<audit_id>(?<time_epoch>[0-9\.]+):(?<serial_number>[0-9]+))\): (?<additional_fields>.+)/
+    ADDITIONAL_MSG_REGEX = /(?<additional_msg>^[^=]+) [a-z0-1_\-]+=/
+    FIELDS_REGEX = /(msg=')?([a-z0-9_\-]+)=("[^"]*"|'[^']*'|[^ ]*)?/
+
+    def parse(line)
+      data = {}
+      time = Time.now.to_f
+
+      begin
+        SKETCH_REGEX.match(line) { |match|
+          data['Computer'] = OMS::Common.get_hostname
+          data['RecordType'] = match['record']
+          data['AuditID'] = match['audit_id']
+          time = match['time_epoch'].to_f
+          data['Timestamp'] = OMS::Common.format_time(time)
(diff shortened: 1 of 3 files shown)


robbiezhang:
Ping @Microsoft/omsagent-devs

source/code/plugins/auditlog_lib.rb
```diff
@@ -0,0 +1,161 @@
...
+          return USER_ID_FIELDS_MAPPING[name], get_user_name_by_id(value.to_i)
+        end
+      
+        if GROUP_ID_FIELDS_MAPPING.has_key? name
+          return GROUP_ID_FIELDS_MAPPING[name], get_group_name_by_id(value.to_i)
+        end
+      rescue => e
+        # type doesn't match
+        @error_handler.log_error("Unable to translate the field: '#{name}': '#{value}'. Message: #{e}")
+      end
```
niroyb: What happens if none of the mappings have the "name" key?

source/code/plugins/auditlog_lib.rb
```diff
@@ -0,0 +1,161 @@
...
+    end
+    
+    def get_typed_value(name, value)
+      # determine the type of the value by name and convert
+      return value
```
niroyb: This doesn't do anything

source/code/plugins/auditlog_lib.rb
```diff
@@ -0,0 +1,161 @@
...
+    end
+    
+    def get_process_name_by_id(pid)
+      begin
+        return IO.read("/proc/#{pid}/cmdline").tr("\000", ' ').strip
```
niroyb: `\0` and `\000` are equivalent. The second form is the octal notation. I think most people would prefer the first.

source/code/plugins/auditlog_lib.rb
```diff
@@ -0,0 +1,161 @@
...
+      rescue
+        # Process terminated
+      end
+    end
+    
+    def get_user_name_by_id(uid)
+      return 'unset' if uid == 4294967295  # stand for 0xffffffff
```
niroyb: You can use `0xff` in ruby.

source/code/plugins/auditlog_lib.rb
```diff
@@ -0,0 +1,161 @@
...
+        @error_handler.log_error("Unable to parse the line #{e}")
+      end
+
+      return time, data
+    end
+    
+
+    private
+
+    def remove_quotations(value)
```
niroyb: If you want to remove all quotations consider using `String.tr` otherwise rename to strip_quotations.

niroyb:
Signed off

source/code/plugins/auditlog_lib.rb
robbiezhang: The column won't be translated if there is no mapping, and no extra property will be added.

source/code/plugins/auditlog_lib.rb
robbiezhang: We don't convert the type for now.

source/code/plugins/auditlog_lib.rb
robbiezhang: Good point. I just want to strip the leading and ending quotes

source/code/plugins/auditlog_lib.rb
robbiezhang: Good point

source/code/plugins/auditlog_lib.rb
robbiezhang: Updated

PR #142, opened 2016-07-27
Adding a new field - UpdateState to be sent for reporting

vrdmr:
Adding a new field "UpdateState". This is required for using CxP
processing of the the update data. Updated the unittests as well.

"UpdateState"=>"Needed" when reporting available updates.
"UpdateState"=>"NotNeeded" when reporting installed packages.

@Microsoft/omsagent-devs @stasku Could you please review the changes.

Thanks

### DIFF
--- a/source/code/plugins/patch_management_lib.rb
+++ b/source/code/plugins/patch_management_lib.rb
@@ -158,6 +158,7 @@ class LinuxUpdates
                                 OMS::Common.format_time(availableUpdatesHash["BuildDate"].to_i)
         ret["Repository"] = availableUpdatesHash.key?("Repository") ? availableUpdatesHash["Repository"] : nil
         ret["Installed"] = false
+        ret["UpdateState"] = "Needed"
         ret
     end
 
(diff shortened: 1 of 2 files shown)


source/code/plugins/patch_management_lib.rb
```diff
@@ -158,6 +158,7 @@ def self.availableUpdatesXMLtoHash(availableUpdatesXML, os_short_name)
                                 OMS::Common.format_time(availableUpdatesHash["BuildDate"].to_i)
         ret["Repository"] = availableUpdatesHash.key?("Repository") ? availableUpdatesHash["Repository"] : nil
         ret["Installed"] = false
+        ret["UpdateState"] = "Needed"
```
robbiezhang: looks like it's equivalent to the "Installed" field. why do you need both field?

stasku: Since we only send Available updates, we might retire "Installed" field in future. Right now ingestion is referring to "Installed", I am switching it to use "UpdateState" instead.

robbiezhang:
Signoff

PR #143, opened 2016-07-29
Stop x86 bundle from being installed into x64

sugr4:
machine and vice versa

@Microsoft/omsagent-devs

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -287,6 +287,31 @@ pkg_upd_list() {
     fi
 }
 
+get_arch()
+{
+    if [ $(uname -m) = 'x86_64' ]; then
+        echo "x64"
+    else
+        echo "x86"
+    fi
+}
+
+compare_arch()
+{
+    #check if the user is trying to install the correct bundle (x64 vs. x86) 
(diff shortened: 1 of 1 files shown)


installer/bundle/bundle_skel.sh
```diff
@@ -287,6 +287,28 @@ pkg_upd_list() {
...
+    echo "x86"
+  fi
+}
+
+compare_arch()
+{
+  #check if the user is trying to install the correct bundle (x64 vs. x86) 
+  echo "Checking host architecture ..."
+  AR=$(get_arch)
+  case $1 in
```
niroyb: Follow the switch case format convention within this file.

installer/bundle/bundle_skel.sh
```diff
@@ -287,6 +287,28 @@ pkg_upd_list() {
...
+    *"$AR".sh) ;;
+    *)         echo "You are trying to install the wrong bundle on a ${AR} bit machine"
```
niroyb: Maybe rephrase to something more passive : "Cannot install bundle $1 on an ${AR} platform."

niroyb:
Signed off

PR #145, opened 2016-08-05
StatsD aggregator for FluentD

robbiezhang:
Implement the statsd aggregator and create an output plugin to buffer the data and aggregate in a configured interval. The aggregated data will be emitted to the FluentD engine again to be sent to the OMS by the out_oms plugin.
Add unit test to cover the statsd aggregation logic.
@Microsoft/omsagent-devs


niroyb:
Signed off

PR #150, opened 2016-08-11
Support -d|--domain optional parameter in the shell bundle

robbiezhang:
Support -d in omsadmin.sh for onboarding and heartbeat
Customer can specify the top level domain in the parameter
to onboard the omsagent agent in FairFax, MoonCake, BlackForest

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -60,6 +60,9 @@ usage()
     echo
     echo "  -w id, --id id         Use workspace ID <id> for automatic onboarding."
     echo "  -s key, --shared key   Use <key> as the shared key for automatic onboarding."
+    echo "  -d dmn, --domain dmn   Use <dmn> as the OMS domain for onboarding. Optional."
+    echo "                         default: opinsights.azure.com"
+    echo "                         ex: opinsights.azure.us (for FairFax)"
     echo "  -p conf, --proxy conf  Use <conf> as the proxy configuration."
     echo "                         ex: -p [protocol://][user:password@]proxyhost[:port]"
     echo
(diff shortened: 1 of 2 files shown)


robbiezhang:
@Microsoft/omsagent-devs @varunkumta

installer/bundle/bundle_skel.sh
```diff
@@ -573,13 +572,6 @@ if [ "$ONBOARD_ERROR" -ne 0 ]; then
     exit 1
 fi
 
-if [ "$onboardINT" -ne 0 ]; then
-    if [ -z "$onboardID" -o -z "$onboardKey" ]; then
```
varunkumta: add this check below, since if -d was specified it means they want to onboard and must specify the id and key

varunkumta:
Signed off

installer/bundle/bundle_skel.sh
```diff
@@ -60,6 +60,7 @@ usage()
     echo
     echo "  -w id, --id id         Use workspace ID <id> for automatic onboarding."
     echo "  -s key, --shared key   Use <key> as the shared key for automatic onboarding."
+    echo "  -d dmn, --domain dmn   Use <dmn> as the domain for onboarding. Optional."
```
agup006: Can we specify it is an Azure domain

installer/scripts/omsadmin.sh
```diff
@@ -68,7 +68,7 @@ usage()
     echo
     echo "Maintenance tool for OMS:"
     echo "Onboarding:"
-    echo "$basename -w <workspace id> -s <shared key>"
+    echo "$basename -w <workspace id> -s <shared key> -d <top level domain>"
```
niroyb: If the -d flag is optional it should be in square brackets like in man pages

niroyb:
Signed off

installer/bundle/bundle_skel.sh
robbiezhang: It should be OMS domain, e.g. opinsights.azure.com, opinsights.azure.us, etc...

PR #152, opened 2016-08-11
Update the ObjectName of the StatsD Perf Data

robbiezhang:
Use user friendly name "StatsD Timer", "StatsD Counter",
"StatsD Set", and "StatsD Gauge" as the ObjectName

@Microsoft/omsagent-devs


agup006:
sign off

PR #155, opened 2016-08-12
Fix the StatsD unit test failure

robbiezhang:
Fix the unit test failure introduced by commit 77317ce790a8bb64f7faf6c2bd91583db7c8fda1
    Update the ObjectName of the StatsD Perf Data
    Use user friendly name "StatsD Timer", "StatsD Counter",
    "StatsD Set", and "StatsD Gauge" as the ObjectName

@Microsoft/omsagent-devs


sw47:
signed off

PR #157, opened 2016-08-12
Bug fix in omsadmin.sh

robbiezhang:
Valid the workspace id is a valid GUID
Delete the certficate if onboarding failed

@Microsoft/omsagent-devs

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -360,6 +360,11 @@ onboard()
         clean_exit 1
     fi
 
+    if [[ ! $WORKSPACE_ID =~ ^[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}$ ]]; then
+        log_error "The Workspace ID is not valid"
+        clean_exit 1
+    fi
+
     PREV_WID=`grep WORKSPACE_ID $CONF_OMSADMIN 2> /dev/null | cut -d= -f2`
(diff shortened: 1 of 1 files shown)


sw47:
signed off

installer/scripts/omsadmin.sh
```diff
@@ -427,6 +433,7 @@ onboard()
     
     if [ $error -ne 0 ]; then
         log_error "Error during the onboarding request. Check the correctness of the workspace ID and shared key or run omsadmin.sh with '-v'"
+        rm "$FILE_CRT" "$FILE_KEY" > /dev/null 2>&1 || true
```
jeffaco: What if the error is transient (server outage, internet outage)? How do we keep from deleting good keys due to a transient problem?

jeffaco:
I'm concerned about transient outages causing potential problems here. Other than that, this looks fine.

Are transient outages a concern?

installer/scripts/omsadmin.sh
robbiezhang: This is the onboard method. Although the cert/key is good, but it's not registered to the OMS serve if there is a server outage, the client should not reuse the cert/key.

jeffaco:
Okay, signed off.

installer/scripts/omsadmin.sh
robbiezhang: Please note that, it's not the renew certificate scenario, where we will backup the old certificate, and overwritten when the renew succeed.

PR #158, opened 2016-08-13
Added conf for MongoDBcommand_CL

sugr4:
Also edited format for MongoDBlog_CL to include messages with a hyphen as component
@Microsoft/omsagent-devs

### DIFF
--- a/installer/conf/omsagent.d/mongo.conf
+++ b/installer/conf/omsagent.d/mongo.conf
@@ -3,7 +3,10 @@
   tag oms.mongo
   host localhost
   port 27017
-  run_interval 10 
+  user mongouser
+  password mongopw
+  auth_database admin
+  run_interval 10
 </source>
 
 #for mongoDB version < 3.0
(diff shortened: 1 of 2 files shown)


installer/conf/omsagent.d/mongo.conf
```diff
@@ -19,13 +19,20 @@
   type tail
   path /var/log/mongodb/mongod.log
   tag  oms.api.MongoDBlog
-  format /(?<timestamp>[^ ]*) (?<severity>[A-Z]) (?<component>[^ ]*)  \[(?<context>[^\]]*)\] (?<message>.*)/
+  format /(?<timestamp>[^ ]*) (?<severity>[A-Z]) (?<component>(-\s*|([^ ]*)))  \[(?<context>[^\]]*)\] (?<message>.*)/
+</source>
+
+#for logs with component: COMMAND
+<source>
```
agup006: Is this re-tailing the file? Is there a way to filter for only messages that contain Command? Additionally can we change the variable to querytime_seconds and convert ms to seconds?

robbiezhang:
Signoff

PR #159, opened 2016-08-13
Fix the regex in omsadmin.sh

robbiezhang:
CentOS 5 doesn't support parentheses in regex

@Microsoft/omsagent-devs

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -360,7 +360,7 @@ onboard()
         clean_exit 1
     fi
 
-    if [[ ! $WORKSPACE_ID =~ ^[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}$ ]]; then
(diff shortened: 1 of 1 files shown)


NarineM:
Looks good. I sign off.

installer/scripts/omsadmin.sh
```diff
@@ -360,7 +360,7 @@ onboard()
         clean_exit 1
     fi
 
-    if [[ ! $WORKSPACE_ID =~ ^[0-9a-fA-F]{8}-([0-9a-fA-F]{4}-){3}[0-9a-fA-F]{12}$ ]]; then
+    if [[ ! $WORKSPACE_ID =~ ^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$ ]]; then
```
jeffaco: Interesting. This script uses /bin/sh which, on RH, is the same as Bash. All I can imagine here is that the machine you found this on has an older version of Bash. Certainly RedHat would have the same issue.

jeffaco:
Signed off

PR #160, opened 2016-08-16
Fix out_oms should batch process

robbiezhang:
Change out_oms plugin to batch data upload based on DataType+IPName instead of tag
@Microsoft/omsagent-devs @mgladi


mgladi:
Signed off

varunkumta:
Signed off

PR #161, opened 2016-08-16
Fix libmysqlclient dependency in the build system

NarineM:
Fix libmysqlclient dependency in the build system

### DIFF
--- a/build/buildRuby.sh
+++ b/build/buildRuby.sh
@@ -50,6 +50,32 @@ elevate()
     return $exit_status
 }
 
+install_libmysqlclient() {
+    MYSQL_PKG_DPKG="libmysqlclient-dev"
+	MYSQL_PKG_RPM="mysql-devel"
+	set +e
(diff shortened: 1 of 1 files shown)


jeffaco:
I do NOT approve this, and this should not be committed. You'll need to revert it.

We don't handle dependencies in our build and dev systems this way. If you need a package installed, we handle that during system creation. That is part of what our systems team does. Otherwise, you can end up with random versions of packages that we can't necessarily easily reconstruct from prior versions.

Did you get sign-off on this? How do you want to go about reverting this?

NarineM:
This is not a package, these are the client native lib  that is needed temparaly for only building the gem for given platform. It needs to be compiled runtime and it never is used later on. This fixes the build till we find better way.

NarineM:
Is there a way to include this particular lib with Jenkins build system? If there is once it is added to jenkin I will remove this part.

jeffaco:
I will discuss with Robbie shortly.

PR #162, opened 2016-08-16
Use upcase instead of upcase!

robbiezhang:
Use upcase instead of upcase! (update the original value, which is not necessary)
@niroyb @Microsoft/omsagent-devs


niroyb:
The problem was that the `upcase!` method returns `nil` if the string is unchanged and we don't want the possibility of null keys.
Signed off

robbiezhang:
Good catch!

PR #163, opened 2016-08-17
Remove the duplicate omi perf counter for MySQL

robbiezhang:
The omi_mapping defines the "InnoDB Buffer Pool Use Pct" twice.
It causes the Test_Marshell method in nxOMSAgent resource always return false if the "InnoDB Buffer Pool Use Pct" is in the configuration.

@Microsoft/omsagent-devs


niroyb:
Signed off

PR #166, opened 2016-08-17
Fix for genRubyInstaller script and new documentation

NarineM:
Fix genRubyinstaller script and added new documentation on how to add a
new gem to omsagent bundle.


installer/scripts/genRubyInstaller.sh
```diff
@@ -28,16 +28,16 @@ substitute_arch()
...
 }
 
-# This script will generate a "starting point" for an InstallBuilder data file.
-# For example, this generated ruby.data in the omsagent project. This is useful
-# for a package that has an install process (i.e. make install), but we need to
-# generate an install data file for it.
+# This script will generate a file that will contain the new
+# ruby  changes for ruby.data file based on SOURCE_DIR content to be included in ruby install builder 
+# since building ruby has an install process
+# and requires generating an install ruby data file for it.
```
jeffaco: That's a heck of a run-on sentence, and is hard to understand.

How about replacing lines 31-34 with something like:

# This script will generate a Ruby.data file, the file that is consumed
# by InstallBuilder, to build a native RPM or DEB installation package.
#
# It is necessary to regenerate this file if new packages have been
# installed into Ruby, and those new packages should be deployed
# on customer systems during OMS installation.

I think this makes more sense. What do you think?

installer/scripts/genRubyInstaller.sh
```diff
@@ -28,16 +28,16 @@ substitute_arch()
...
 
-OUTPUT_RESULTS="/tmp/installBuilder-$USER"
+OUTPUT_RESULTS="/tmp/installBuilder-ruby-data-$USER"
 rm -f $OUTPUT_RESULTS
 
 # Evaluate any wildcards in the filename ...
-SOURCE_DIR=`stat -c "%n" ~/dev/work/oms/intermediate/*/ruby`
+SOURCE_DIR=`stat -c "%n" ../intermediate/*/100/ruby`
```
jeffaco: This is making an assumption about your current working directory. If someone doesn't know anything about this script, it will fail for no obvious reason. Before this line, add something like:

```
if [ ! -f configure -o ! -d ../build ]; then
    echo "Must be in OMS build directory for this script to run"
    exit 1
fi

if [ -z "$SOURCE_DIR" ]; then
    echo "Intermediate SSL_100 ruby version does not exist"
    echo "Please build OMS configured with --enable-ulinux"
    exit 1
fi
```

Double check that these 'if' statements do what I expect.

jeffaco:
Did you re-run genRubyInstaller.sh with these changes? Is it's generated output identical to what is currently checked in? If not, what's different?

docs/How-to-add-new-gem-to-omsagent-bundle.md
```diff
@@ -0,0 +1,49 @@
+## How to add a new gem to omsagent install bundle
+
+In order to add a new Ruby gem into OMSAgent bundle when you have the new_gem_name.gem file on you local system you need to follow these steps:
+1. Update the private branch:
```
jeffaco: The output of this is weird if you look at it displayed as markdown. Please correct.

docs/How-to-add-new-gem-to-omsagent-bundle.md
```diff
@@ -0,0 +1,49 @@
...
+* Clone Microsoft/Build-OMS-Agent-for-Linux github repo
```
jeffaco: You could just give the command to do this ...

docs/How-to-add-new-gem-to-omsagent-bundle.md
```diff
@@ -0,0 +1,49 @@
...
+* Copy the new_gem_name.gem to Build-OMS-Agent-for-Linux//omsagent/source\ext\gems folder.
```
jeffaco: Use consistent `/` formats that will actually work on Linux please. And rather than text like this, I prefer a concrete example. You can use your sample file, for example. Give REAL commands, that will help the reader to understand.

docs/How-to-add-new-gem-to-omsagent-bundle.md
```diff
@@ -0,0 +1,49 @@
...
+* Navigate to Build-OMS-Agent-for-Linux//omsagent/build and put modified commands in buildRuby.sh
+(Example ${RUBY_DESTDIR}/bin/gem install ${BASE_DIR}/source/ext/gems/new_gem_name.gem)
+* Run buildRuby.sh 100
```
jeffaco: This assumes a lot, like that you're in the build directory, you ran configure, etc. You may want to just suggest that they configure and build OMS before following these instructions.

docs/How-to-add-new-gem-to-omsagent-bundle.md
```diff
@@ -0,0 +1,49 @@
...
+* It will build the new changes for ruby under intermediate/*/100/Ruby.
+* Run ../installer/scripts/genRubyInstaller.sh.
+* It will create a new file containing changes for ruby.data in /tmp/installBuilder-ruby-data-$USER
+* Diff /tmp/installBuilder-ruby-data-$USER with the rub.data file from Build-OMS-Agent-for-Linux/omsagent/installer/datafiles/.
+* Observe only the new gem related records are added. Replace installer/datafiles/ruby.data with the content of installBuilder-ruby-data-$USER file
+
```
jeffaco: I definitely prefer a real-life example, like I commented on above. I understand this because I described it to you. But imagine someone comes along with no clue, you and I have moved on, and the reader has nobody to turn to. Giving a real example that they can look at, with output, would be a lot more helpful.

docs/How-to-add-new-gem-to-omsagent-bundle.md
```diff
@@ -0,0 +1,49 @@
...
+* Clone Microsoft/Build-OMS-Agent-for-Linux github repo
+* Find the newly installed new_gem_name.gem file from /opt/microsoft/omsagent/ruby path and copy it to Build-OMS-Agent-for-Linux//omsagent/source/ext/gems.
+* Navigate to Build-OMS-Agent-for-Linux/omsagent/installer/scripts and update genRubyInstaller.sh with SOURCE_DIR to be /opt/microsoft/omsagent/ruby
+* Run genRubyInstaller.sh. 
+* It will create a new file containing changes for ruby.data in /tmp/installBuilder-ruby-data-$USER
+* Diff /tmp/installBuilder-ruby-data-$USER with the rub.data file from Build-OMS-Agent-for-Linux/omsagent/installer/datafiles/.
+* Observe only the new gem related records are added. Replace installer/datafiles/ruby.data with the content of installBuilder-ruby-data-$USER file
+* Uninstall the agent (required to be able to build the agent from your private github depot)
+* Put  modified commands to Build-OMS-Agent-for-Linux/omsagent/build/buildRuby.sh
+(Example ${RUBY_DESTDIR}/bin/gem install ${BASE_DIR}/source/ext/gems/new_gem_name.gem)
```
jeffaco: This is very long, lots of potential places to get confused. You have a real life example, I suggest you use it and give a nice, readable page, explaining (in real life) how you might add the MySQL gem into OMS.

Real life examples that they can read and follow, with real life screen output, is much easier to understand and less likely to confuse.

Also, your markdown formatting is hard to understand. You might want to use headers (bigger bolded text) to differentiate between sections. I didn't even understand you were using a numbered list for a while there.

If you want a numbered list, you can do that, but you probably want a table of contents for screen output and details. I've done that in a bunch of places (most pages, actually), in ostc-docs.

docs/How-to-add-new-gem-to-omsagent-bundle.md
```diff
@@ -0,0 +1,49 @@
...
+
+2. Building the new changes privately:
+* Navigate to Build-OMS-Agent-for-Linux/omsagent/build to configure the build environment with --enable-ulinux
+* make
+3. Observe that the ruby generated in intermediate folder contains the new gem and it's required files in
+
+Note that if the new gem you are adding is dependent on other gems, you need to checkin dependent gems as well.
```
jeffaco: You don't just need to check in dependent gems as well, you need to add them to buildRuby.sh in non-dependent order. I'm not sure you'll have a `vendor/cache` directory in this case since you're not building the GEM itself (I'm honesty not sure). But you can certainly look at the output, I'd think, and add the gems in reverse order, then rebuild and make sure that nothing was downloaded to satisfy dependencies.

Honestly, neither of us ever tried this. It might be useful, for an example, to add some random GEM file with dependencies, just so you can document this section properly. And by doing that, you have a different real-life example that you can explain on the page with real-life screen output.

docs/How-to-add-new-gem-to-omsagent-bundle.md
```diff
@@ -0,0 +1,49 @@
...
+(Example ${RUBY_DESTDIR}/bin/gem install ${BASE_DIR}/source/ext/gems/new_gem_name.gem)
+
+2. Building the new changes privately:
+* Navigate to Build-OMS-Agent-for-Linux/omsagent/build to configure the build environment with --enable-ulinux
+* make
+3. Observe that the ruby generated in intermediate folder contains the new gem and it's required files in
+
+Note that if the new gem you are adding is dependent on other gems, you need to checkin dependent gems as well. 
+As a reference you can look at how fluentd installs the dependant gems from vendor/cache
+If your gem is depenendant on native libraries, the section "Dependencies to build a native package" in Build-OMS-Agent-for-Linux has to be updated accordingly.
```
jeffaco: This is a little unclear. I needed to read it multiple times to understand that you mean a GEM might need an RPM or DEB file to be installed. Explain yourself a little more clearly here.

Also, rather than tell them to see the section, why not just have a link to that section, so they can click to see it?

docs/How-to-add-new-gem-to-omsagent-bundle.md
```diff
@@ -0,0 +1,49 @@
...
+
+
+
+
+
+
```
jeffaco: Eliminate needless blank lines.

jeffaco:
I'd like to see another iteration, please. When I documented how to create a tag and branches and stuff, I spent about 1/2 day gathering output and putting it in consumable, easy to understand form in ostc-docs. You should plan on the same.

docs/How-to-add-new-gem-to-omsagent-bundle.md
NarineM: I see. shortening the name. Thanks.

docs/How-to-add-new-gem-to-omsagent-bundle.md
NarineM: Ok. Adding real life example instead of a placeholder.

installer/scripts/genRubyInstaller.sh
NarineM: Since SOURCE_DIR is using relative path from the build location, it will fail at line # 50 if the script is lunch not from the build location. But I agree that changing the echo to be more descriptive would be really useful. Changing it now.

# Verify validity of source directory

if [ -z "$SOURCE_DIR" -o ! -d $SOURCE_DIR ]; then
    echo "Source directory ($SOURCE_DIR) does not exist" >& 2
    exit 1
fi

installer/scripts/genRubyInstaller.sh
NarineM: Agreed. Thanks!

installer/scripts/genRubyInstaller.sh
NarineM: Yes, I re-run genRubyInstaller with these changes. It is generating the output identical to what currently is checked in.

docs/How-to-add-new-gem-to-omsagent-bundle.md
NarineM: I think I am going to leave the dependant gems peace out of this how to doc. It makes the doc too complicated. Let us keep it simple. We can certainly have a separate how to once we have such implementation done at least once.

jeffaco: My concern here is that WE (you and I) understand, at least in theory, how this works. Someone else would not, and might not fully understand things. This leads to the possibility of not "fixing things" properly, creating a scenario where the build actually downloads "random" dependencies.

I think it's best to do a how-to of adding some Gem (pick one almost at random) that has 1 or 2 dependencies, so we can document how it works. Yes, it takes longer to do this. But on the bright side, WE understand it, so we should explain it.

Talk to Robbie. If he's okay skipping this, that's his call, and I'll accept his guidance. But I'd include it if I were you ... (unless it turned out to take a week of extra time or something).

NarineM:
@jeffaco 
Fixed the markdown and added more details to the how to in the second commit.

installer/scripts/genRubyInstaller.sh
```diff
@@ -52,14 +52,20 @@ if [ -z "$SOURCE_DIR" -o ! -d $SOURCE_DIR ]; then
     exit 1
```
jeffaco: I'm after better diagnostics here (which is what I had done in prior reviews).

Imagine an error like "This script must be run from the OMS build directory" rather than "Source directory does not exist".

One tells you just what to do, another requires that you look at a script and diagnose the problem.

I'm after two basic checks (which I sent in a prior review).

1. Are we starting off in the correct directory? If not, we'll fail,
2. Are we doing a universal build, and has the build been done? If not, we'll fail.

I'd like those to be checked for since those are both very common mistakes.

docs/How-to-add-new-gem-to-omsagent-bundle.md
```diff
@@ -0,0 +1,42 @@
...
+* Observe only mysql2 gem related records are added. 
+* Replace /installer/datafiles/ruby.data in build location with the content of installBuilder-ruby-data-$USER file
+* Navigate to /omsagent/build/ folder and put modified commands in buildRuby.sh
+  Example, `${RUBY_DESTDIR}/bin/gem install ${BASE_DIR}/source/ext/gems/mysql_2.0.gem`
+* Uninstall omsagent to be able to build it privately.
+* Navigate to build location. Run `./configure --enable-ulinux`. And build the bundle using make.
+  
+Then install the new bundle and validate the ruby folder under /opt/microsoft/omsagent/ruby has all the new changes.
+  
+Note that if your gem has dependencies on native libraries, the section [Dependencies to build a native package](https://github.com/Microsoft/Build-OMS-Agent-for-Linux)has to be updated accordingly.
```
jeffaco: Add a space between the ) and "has".

You can link directly to the section you want with # afterwards to point to that specific entry on the page.

At first I thought the link was broken. Make it clearer of what you want with something like "If you new .gem file adds dependencies on native packages that must be installed, [update the README.md](https://github.com/Microsoft/Build-OMS-Agent-for-Linux#dependencies-to-build-a-native-package) file.

jeffaco:
This comment got buried because of new commits, and it's important, so I'm resurrecting it here.

Regarding .gem files with dependencies:

My concern here is that WE (you and I) understand, at least in theory, how this works. Someone else would not, and might not fully understand things. This leads to the possibility of not "fixing things" properly, creating a scenario where the build actually downloads "random" dependencies.

I think it's best to do a how-to of adding some Gem (pick one almost at random) that has 1 or 2 dependencies, so we can document how it works. Yes, it takes longer to do this. But on the bright side, WE understand it, so we should explain it.

Talk to Robbie. If he's okay skipping this, that's his call, and I'll accept his guidance. But I'd include it if I were you ... (unless it turned out to take a week of extra time or something).

jeffaco:
@robbiezhang 

At this point, I'd like Robbie's guidance on how to handle the docs for .gem files with dependencies. I think it should be updated to discuss that and how to resolve it properly, but I'll accept Robbie's guidance on this.

Otherwise, I'm fine with this (assuming other changes are made) and I sign off. In particular, please be sure to improve the genRubyInstaller.sh script to give meaningful error messages so you don't need to debug the script for common problems.

NarineM:
@jeffaco 
I see your point of making sure someone does not accidentally make the build system download non existing dependent gems without knowing about dependencies. Although, we assume that someone know what he/she is doing :) Regardless if the build process changes in the future or not, I agree that it would be beneficial to document that here since we still use the existing processes. I will update this document to warn about dependent gem install even though I did not plan to cover it here in this how to.

installer/scripts/genRubyInstaller.sh
NarineM: #1 is already covered in existing code.  I did not see the need for #2 as in my scenario I used /opt/Microsoft/ path for getting the ruby changes from (which is covered in the second part of the how to doc) where you do not have to have the make ran already prior to running genRubyInstaller.sh.

docs/How-to-add-new-gem-to-omsagent-bundle.md
NarineM: Sure.

NarineM:
@jeffaco 
Thanks Jeff for code review. I have updated the commit to address your comments. The only remaining part is to add a real world example for dependent gems. Now the doc captures the dependent gem install process but with no example. I will get back to it for adding an example for that scenario also as soon as time permits.

NarineM:
Merged.

PR #167, opened 2016-08-17
Removed mysql pre-req of mysql-devel for the end user

NarineM:
Removed mysql pre-req of mysql-devel for the end user since this
dependancy on native lib is needed for building the omsagent bundle only
and not during the bundle installation. Added the lib dependancy to
build-oms-agent document instead.

@Microsoft/omsagent-devs


jeffaco:
Signed off

robbiezhang:
Signoff

NarineM:
Merged #167

PR #168, opened 2016-08-17
Remove unit for MongoDBlog metric querytime

sugr4:
@Microsoft/omsagent-devs


robbiezhang:
The type of the property is till string.

sugr4:
Changed it from string to float.

robbiezhang:
Nice, signoff

PR #170, opened 2016-08-18
Fix the unit test failure because of dup omi perf counter

robbiezhang:
Removed the dup perf counter ("MySQL Server InnoDB Buffer Pool Use Pct") in the unit test.

@Microsoft/omsagent-devs


jeffaco:
Signed off. You did test to make sure the units tests now pass, yeah?

robbiezhang:
I did it this time. I under-estimated the coverage of the unit tests. :)
sorry about that.

PR #177, opened 2016-08-26
Replace bash regex matching with grep

sugr4:
@Microsoft/omsagent-devs


jeffaco:
I approve this. However, please update the commit message. The summary line is fine, but I'd like additional detail on why this change was made please (as a multi-line commit message).

Thanks.

niroyb:
Signed off

PR #178, opened 2016-08-29
PerfCounter name change to include KB

lagalbra:
Goal: add units label to "Used Memory" CounterName in Perf Process object
Implementation: Extended omi_mapping.json schema, added logic in oms_omi_lib.rb to support user configuration compatibility, updated unit tests

### DIFF
--- a/installer/conf/omi_mapping.json
+++ b/installer/conf/omi_mapping.json
@@ -219,7 +219,8 @@
             },
             {
                 "CimPropertyName": "UsedMemory",
-                "CounterName": "Used Memory"
+                "CounterName": "Used Memory",
+                "DisplayName": "Used Memory kBytes"
             },
             {
                 "CimPropertyName": "VirtualSharedMemory",
--- a/source/code/plugins/oms_omi_lib.rb
+++ b/source/code/plugins/oms_omi_lib.rb
@@ -73,7 +73,13 @@ class OmiOms
     cimproperties.each { |maps|
       cim_name = maps["CimPropertyName"]
       oms_name = maps["CounterName"]
-      cim_to_oms[cim_name] = oms_name
+      # Map cim_name to both CounterName and DisplayName
+      if maps.has_key?("DisplayName")
+        display_counter_name = maps["DisplayName"]
+        cim_to_oms[cim_name] = [oms_name, display_counter_name]
+      else
+        cim_to_oms[cim_name] = oms_name
+      end
     }
     return cim_to_oms
   end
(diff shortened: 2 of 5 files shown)


lagalbra:
@Microsoft/omsagent-devs

jeffaco:
The commit message is too long. Please shorten it to 70 bytes or less, or make it a multi-line commit.

This otherwise looks fine to me, and I sign off. However, Nicolas understands limitations around the OMI mapping more than I. I'll ask him to look at this.

@niroyb

test/code/plugins/omi_lib_test.rb
```diff
@@ -186,7 +186,7 @@ def test_transform_class_returns_valid_record_all_counters
...
-		expected_process_record = [{"Host":"testhost", "ObjectName":"Process","InstanceName":"omsagent","Collections":[{"CounterName":"Pct User Time","Value":"0"},{"CounterName":"Pct Privileged Time","Value":"0"},{"CounterName":"Used Memory","Value":"40208"},{"CounterName":"Virtual Shared Memory","Value":"5216"}]}]
+		expected_process_record = [{"Host":"testhost", "ObjectName":"Process","InstanceName":"omsagent","Collections":[{"CounterName":"Pct User Time","Value":"0"},{"CounterName":"Pct Privileged Time","Value":"0"},{"CounterName":"Virtual Shared Memory","Value":"5216"}]}]
```
niroyb: Where did the "Used Memory" property go?

installer/conf/omi_mapping.json
```diff
@@ -219,7 +219,8 @@
             },
             {
                 "CimPropertyName": "UsedMemory",
-                "CounterName": "Used Memory"
+                "CounterName": "Used Memory",
```
niroyb: I don't understand the use of DisplayName. Why not just change the counter name?

source/code/plugins/oms_omi_lib.rb
```diff
@@ -73,7 +73,13 @@ def get_cim_to_oms_mappings(cimproperties)
     cimproperties.each { |maps|
       cim_name = maps["CimPropertyName"]
       oms_name = maps["CounterName"]
-      cim_to_oms[cim_name] = oms_name
+      # Map cim_name to both CounterName and DisplayName
+      if maps.has_key?("DisplayName")
+        display_counter_name = maps["DisplayName"]
+        cim_to_oms[cim_name] = [oms_name, display_counter_name]
```
niroyb: I don't like that the cim_to_oms hash now has strings and arrays as values and we use the type to determine the behavior.

test/code/plugins/omi_lib_test.rb
lagalbra: When I ran unit tests, this property was dropped. I was told these tests were deprecated, and I should work around it, which is why I removed the "Used Memory" property from this test file.

installer/conf/omi_mapping.json
lagalbra: In order to maintain backward and forward user compatibility; since users may have searches that query on the "Used Memory" CounterName, I needed a way to keep the same returned data for them, but still show on the portal "Used Memory kBytes".

source/code/plugins/oms_omi_lib.rb
lagalbra: I didn't like it too much either, but it was the easiest way to preserve the current dataflow in that file while still enabling the functionality that the DisplayName field represents.

lagalbra:
Shortened the commit message.

niroyb:
Thanks for the replies. Signed off. I think it would be better to do the mapping on the server side so that a client update is not required for these kinds of changes. It would also be possible to be backwards compatible with the queries.

test/code/plugins/omi_lib_test.rb
robbiezhang: this unit test is for the omi_lib, instead of the new oms_omi_lib. 
We need to do some clean up to remove the non-used plugins and files.

lagalbra:
@robbiezhang Can you sign off on the omi_mapping.json path updates?

robbiezhang:
Sign off

lagalbra:
Change is now merged and pushed to master

PR #179, opened 2016-08-29
Fix error when checking the arch if bundle file is renamed

sugr4:
@Microsoft/omsagent-devs

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -304,12 +304,12 @@ compare_arch()
     #check if the user is trying to install the correct bundle (x64 vs. x86) 
     echo "Checking host architecture ..."
     AR=$(get_arch)
-
-    case $1 in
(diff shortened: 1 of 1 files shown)


jeffaco:
Care to elaborate why the bundle file is renamed in this way?

Currently the name is something like: `omsagent-1.2.0-38.universal.x64.sh`. What are you trying to rename, and why? The .deb and .rpm changes shouldn't affect this part, and changing this could adversely affect using this kit in OM (unless something is happening here that I don't know about).

sugr4:
@jeffaco This PR is to resolve an issue where if the shell bundle is renamed by the customer after download, s/he is unable to onboard because the compare_arch is looking at the bundle name for checking if the bundle arch matches the processor arch. I changed the code so that the processor arch is checked against the value of OMS_PKG and not the name of the bundle.

jeffaco:
Signed off.

PR #183, opened 2016-08-31
Add correct collectd dir for rpm systems

sugr4:
@Microsoft/omsagent-devs

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -28,7 +28,6 @@ ONBOARD_FILE=/etc/omsagent-onboard.conf
 DPKG_CONF_QUALS="--force-confold --force-confdef"
 OMISERV_CONF="/etc/opt/omi/conf/omiserver.conf"
 OLD_OMISERV_CONF="/etc/opt/microsoft/scx/conf/omiserver.conf"
-COLLECTD_DIR="/etc/collectd/collectd.conf.d/"
 
 # These symbols will get replaced during the bundle creation process.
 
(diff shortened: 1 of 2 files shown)


installer/bundle/bundle_skel.sh
```diff
@@ -550,10 +549,10 @@ do
             ;;
 
         --collectd)
-            if [ -d "${COLLECTD_DIR}" ]; then
+            if [ -f /etc/collectd.conf ] || [ -f /etc/collectd/collectd.conf ]; then
```
jeffaco: In shell, that's not the way you do that.

This should be something like:

`if [ -f /etc/collectd.conf  -o -f /etc/collectd/collectd.conf ]; then`

I'm not even sure the || construct will work reliably, as that's a pipe thing in this context.

installer/bundle/bundle_skel.sh
```diff
@@ -681,8 +680,11 @@ if [ "$installMode" = "R" -o "$installMode" = "P" ]; then
             rmdir /etc/opt /var/opt > /dev/null 2> /dev/null || true
         fi
     fi
-    if [ -f ${COLLECTD_DIR}/oms.conf ]; then
-        rm ${COLLECTD_DIR}/oms.conf > /dev/null 2> /dev/null
+    if [ -f /etc/collectd.d/oms.conf ]; then
+        rm /etc/collectd.d/oms.conf > /dev/null 2> /dev/null
+    fi
+    if [ -f /etc/collectd/collectd.conf.d/oms.conf ]; then
+        rm /etc/collectd/collectd.conf.d/oms.conf > /dev/null 2> /dev/null
```
jeffaco: You could just replace all this with two lines:

```
rm -f /etc/collectd.d/oms.conf > /dev/null 2> /dev/null
rm -f /etc/collectd/collectd.conf.d/oms.conf > /dev/null 2> /dev/null
```

The -f says force remove, and don't consider it an error if the file doesn't exist. That eliminates the pre-checks.

jeffaco:
Fix the "or" handling in the shell scripting, otherwise looks fine to me, signed off.

sugr4:
Made changes based on comments, built an tested new bundles.

PR #185, opened 2016-09-02
Updated omsagent installation guide for post install onboarding

NarineM:
Updated oms agent installation guide for post install onboarding to use service_control enable to register omsagent and restart the service altogether. Per current design, oms agent service restart no longer works since at the time it is called (right after onboarding using cmdline or a file) the agent is not registered as a service.
@Microsoft/omsagent-devs

### DIFF
--- a/docs/OMS-Agent-for-Linux.md
+++ b/docs/OMS-Agent-for-Linux.md
@@ -171,6 +171,7 @@ Run the omsadmin.sh command supplying the workspace id and key for your workspac
 ```
 cd /opt/microsoft/omsagent/bin
 sudo ./omsadmin.sh -w <WorkspaceID> -s <Shared Key>
+sudo ./service_control enable
 ```
 
 ## Onboarding using a file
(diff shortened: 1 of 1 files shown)


docs/OMS-Agent-for-Linux.md
```diff
@@ -181,8 +182,8 @@ sudo ./omsadmin.sh -w <WorkspaceID> -s <Shared Key>
 WORKSPACE_ID=<WorkspaceID>
 SHARED_KEY=<Shared Key>
 ```
-3.	Restart the omsagent:
-`sudo service omsagent restart`
+3.	Register the omsagent service and restart:
+`sudo ./service_control enable`
```
niroyb: Maybe specify the full path of the script

niroyb:
Signed off

docs/OMS-Agent-for-Linux.md
jeffaco: I think this is wrong. You still want to use the restart command.

The restart command will register the service, enable the service (or it should upon registration), and then start the service. This is what you want.

If you simply enable it, that will enable restarting the service at reboot time, which is a subset of what you want.

Another iteration please, or just throw this away since existing code should be fine. If not, then another iteration please.

jeffaco:
NOT signed off, see my comments.

NarineM:
As per discussion, this is a doc change to be in sync with the current design since it is used by end user and current steps do not work. The user gets "service does not exist " error. Once the product is changed, the user documentation would need to be updated to be in sync with the new changes.

PR #186, opened 2016-09-02
Fix building test ruby that fails due to ruby clean up leaving test ruby corrupted

NarineM:
We have seen instances (in pBuild and dev environments) where make does not properly rebuild the local ruby if there was a previous build break or failure which causes subsequent make command or running make with unit tests fail. Not an issue in clean VM setup where the bundle was never built or was always built successfully. When it repros local ruby folder is left corrupted and the workaround is to manually delete it then rerun the make. 
Looking at the code it seems weird to see removing versions of local ruby being done from the local ruby folder while that folder itself is expected to be deleted.  This PR is about moving the localruby cleanup logic be performed in ruby cleanup stage(used in distclean) to be consistent with the code that prepares the system to install ruby as well as be consistent with the makefile. This should help to resolve the issue we are seeing with local ruby being corrupted and make not being able to properly rebuild it.
@Microsoft/omsagent-devs ;

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -193,7 +193,6 @@ $(RUBY_TESTING_DIR) :
 	# Warning: This step will clean out checked out files from both Ruby and fluentd directories
 	#
 	@$(ECHO) "========================= Performing Building Ruby for testing"
-	sudo $(RMDIR) /usr/local/ruby-2.2.0*
 	$(BASE_DIR)/build/buildRuby.sh test
 
 # Build the version of Ruby that we distribute
--- a/build/configure
+++ b/build/configure
@@ -389,7 +389,7 @@ ruby_configure_quals=(
      )
 
 ruby_config_quals_sysins="--prefix=/opt/microsoft/omsagent/ruby"
-ruby_config_quals_testins="--prefix=/usr/local/ruby-2.2.0d"
+ruby_config_quals_testins="--prefix=${base_dir}/intermediate/${BUILD_CONFIGURATION}/local_ruby-2.2.0d"
 
 if [ "$ULINUX" = "1" ]; then
     ssl_098_dirpath=/usr/local_ssl_0.9.8
--- a/build/buildRuby.sh
+++ b/build/buildRuby.sh
@@ -235,8 +235,9 @@ if [ $RUNNING_FOR_TEST -eq 0 ]; then
 
     # Pacify Make (Make doesn't know that the generated Ruby directory can vary)
     mkdir -p ${BASE_DIR}/intermediate/${BUILD_CONFIGURATION}/ruby
-else 
-    echo "Installing MySQL gem into Test Ruby..." 
+else
+    mkdir -p ${RUBY_DESTDIR}
+    echo "Installing MySQL gem into Test Ruby path: ${RUBY_DESTDIR}..." 
     elevate ${RUBY_DESTDIR}/bin/gem install mysql2
 fi
 


build/Makefile
```diff
@@ -126,6 +126,7 @@ clean : clean-status
 clean-ruby : clean
 	sudo $(RMDIR) $(BASE_DIR)/intermediate
 	sudo $(RMDIR) $(RUBY_DIR)/.ext
+	sudo $(RMDIR) /usr/local/ruby-2.2.0*
```
jeffaco: The problem with this sort of approach is that we don't always do a 'make clean' to clean things up. We almost exclusively just use git to clean things up (i.e. `git clean` with options, which is what `git nuke` does).

If this is your intent, I think a much better way is to simply eliminate the copy of Ruby in /usr/local entirely, instead building Ruby in the users local directory (i.e. intermediate/*/local_ruby). This way, `git nuke` would clean it up as well.

I still believe that the better way is to save a checksum, and if something key changes, then rebuild ruby. This would get rid of the notion of a letter in the directory; instead, we'd recognize that it would need to be rebuilt and just rebuild it.

But if you don't want to do that, then building Ruby on a per-user basis in the intermediate directory is a better way to go. Then `git nuke` would clean it up, which is what the build system generally does anyway.

jeffaco:
See comments above. I don't think this approach will ultimately do what you wish ...

build/Makefile
NarineM: @jeffaco 
Moved local ruby for testing to be built under intermediate. and having $(RMDIR) $(BASE_DIR)/intermediate handle clean up of local ruby automatically .

NarineM: Renamed the description of this PR because the scope is changed.

build/configure
```diff
@@ -389,7 +389,7 @@ ruby_configure_quals=(
      )
 
 ruby_config_quals_sysins="--prefix=/opt/microsoft/omsagent/ruby"
-ruby_config_quals_testins="--prefix=/usr/local/ruby-2.2.0c"
+ruby_config_quals_testins="--prefix=${base_dir}/intermediate/local_ruby-2.2.0c"
```
jeffaco: No need to have the letter separator. Just call it local-ruby-2.2.0 at that point, since any chances to it would just rebuild it anyway. Also, rather than putting it in a fixed dir (ignoring debug vs. release), shouldn't this be based on $BUILD_CONFIGURATION, which is what (I believe) the subdirectory is called under intermediate?

This will likely require "sudo" to do a "git nuke", right, since Ruby will be "installed" into the intermediate_dir with root permissions? This will be slightly less convenient for OMS devs that use 'git nuke', but it looks like the build system uses sudo for that already.

jeffaco:
Squash to a single commit that is <= 70 bytes in length, as per git guidelines.

Since you're not going to do the checksum thing, please at least make this sensitive to debug/release (via the `$BUILD_CONFIGURATION` setting, I believe) so it'll be put where everything else is put for the build.

I do kind of wonder if we should just bag building Ruby for test purposes anyway, and just use something like [rbenv](https://github.com/rbenv/rbenv) instead. This would give you a version of your choice (important for being tied to the same version of Ruby we run with), give you independence from the system version of Ruby (which, today, needs to be enough to bootstrap building Ruby, which can be just about anything), and would get you out of the business of building the damn thing every time (reduce build time by 30%, essentially). But that's a bigger change including testing, so it might not be worth it at this time.

Minimally, if you put the test version of Ruby in the "proper" intermediate directory (based on `$BUILD_CONFIGURATION`) and test that it's indeed in the proper place, then I'll sign off on this. No need for another review as long as it's in the same directory as all other build artifacts. Otherwise, let's discuss.

NarineM:
@jeffaco 
Thanks Jeff. I actually had the exact same thought about test ruby and whether we really need to bother the make and build it.  There are certainly areas that need to be refactored, based on the priority of course. Althought the refactoring will make the system optimal and shorten the overall build time. I consider fixing pBuild and dev environments as a priority and this should do the thing. Later we can come back to existing make process and look into it more.

NarineM:
Just realized that did not merge this PR.  Re-opened to check it in.

PR #188, opened 2016-09-06
Nit fix for sending timestamps only for valid dates

vrdmr:
Earlier, timestamp for each installed or available
update was being sent, which did not have clean
data and thus were misrepreseted in the reporting.

This fix makes this field optional and not send
if the underlying resource does not send correct
format or does not send at all.

Please review.

@Microsoft/omsagent-devs @stasku


NarineM:
Looks good. Signoff.

robbiezhang:
signoff

vrdmr:
Thanks

stasku:
sign-off

niroyb:
signed off

PR #193, opened 2016-09-09
Add option for adding OMSCloudID and AzureResourceID

sugr4:
to http headers on onboarding

@Microsoft/omsagent-devs

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -44,28 +44,29 @@ usage()
 {
     echo "usage: $1 [OPTIONS]"
     echo "Options:"
-    echo "  --extract              Extract contents and exit."
-    echo "  --force                Force upgrade (override version checks)."
-    echo "  --install              Install the package from the system."
-    echo "  --purge                Uninstall the package and remove all related data."
-    echo "  --remove               Uninstall the package from the system."
-    echo "  --restart-deps         Reconfigure and restart dependent service(s)."
-    echo "  --source-references    Show source code reference hashes."
-    echo "  --upgrade              Upgrade the package in the system."
-    echo "  --version              Version of this shell bundle."
-    echo "  --version-check        Check versions already installed to see if upgradable."
-    echo "  --debug                use shell debug mode."
-    echo "  --collectd             Enable collectd."
+    echo "  --extract                  Extract contents and exit."
+    echo "  --force                    Force upgrade (override version checks)."
+    echo "  --install                  Install the package from the system."
+    echo "  --purge                    Uninstall the package and remove all related data."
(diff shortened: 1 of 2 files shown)


installer/scripts/omsadmin.sh
```diff
@@ -416,8 +422,40 @@ onboard()
 
     set_proxy_setting
 
-    # Send the request to the registration server
-    RET_CODE=`curl --header "x-ms-Date: $REQ_DATE" \
+    if [ -n "$AZURE_RESOURCE_ID" ]; then
```
robbiezhang: you don't need to add them into this header

installer/scripts/omsadmin.sh
```diff
@@ -416,8 +422,40 @@ onboard()
...
+        if [ -f /sys/devices/virtual/dmi/id/chassis_asset_tag ]; then
+            ASSET_TAG=$(cat /sys/devices/virtual/dmi/id/chassis_asset_tag)
+            case "$ASSET_TAG" in
+                77*) OMSCLOUD_GUID=$ASSET_TAG ;;       # If the asset tag begins with a 77 this the azure guid
```
robbiezhang: always get the GUID and store it into the omsadmin.conf no matter the AzureResourceId is specified or not.

installer/scripts/omsadmin.sh
```diff
@@ -116,6 +118,7 @@ save_config()
     echo URL_TLD=$URL_TLD >> $CONF_OMSADMIN
     echo DSC_ENDPOINT=$DSC_ENDPOINT >> $CONF_OMSADMIN
     echo OMS_ENDPOINT=https://$WORKSPACE_ID.ods.$URL_TLD/OperationalData.svc/PostJsonDataItems >> $CONF_OMSADMIN
+    echo AZURE_RESOURCE_ID=$AZURE_RESOURCE_ID >> $CONF_OMSADMIN
```
robbiezhang: also save the OMSCloudID

installer/scripts/omsadmin.sh
robbiezhang: can you also get the UUID here?

sugr4:
Made changes based on comments

robbiezhang:
Signoff

installer/bundle/bundle_skel.sh
```diff
@@ -558,6 +559,11 @@ do
             shift 1
             ;;
 
+        -a|--id)
```
varunkumta: --id is also used for the workspace id -w|--id so we should probably use another string here

robbiezhang: good catch.

PR #199, opened 2016-09-12
Make IPName optional in the out_oms plugin

robbiezhang:
Make IPName an optional field for the data out_oms plugin
@Microsoft/omsagent-devs

### DIFF
--- a/source/code/plugins/out_oms.rb
+++ b/source/code/plugins/out_oms.rb
@@ -91,8 +91,9 @@ module Fluent
       datatypes = {}
       unmergable_records = []
       chunk.msgpack_each {|(tag, record)|
-        if record.has_key?('DataType') and record.has_key?('IPName')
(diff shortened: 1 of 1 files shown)


vrdmr:
Sign off

source/code/plugins/out_oms.rb
```diff
@@ -91,8 +91,9 @@ def write(chunk)
       datatypes = {}
       unmergable_records = []
       chunk.msgpack_each {|(tag, record)|
-        if record.has_key?('DataType') and record.has_key?('IPName')
-          key = "#{record['DataType']}.#{record['IPName']}".upcase
+        if record.has_key?('DataType')
```
varunkumta: if this change is for the ADM issue .. shouldn't we instead fix the case-sensitivity issue? i.e. accept both IPName and IPNAME .. but I still think we should have exactly one of those 2

PR #200, opened 2016-09-13
Add azureResourceID and OMScloudID in the httpRequest headers

sugr4:
for all output plugins

@Microsoft/omsagent-devs

### DIFF
--- a/source/code/plugins/oms_common.rb
+++ b/source/code/plugins/oms_common.rb
@@ -677,6 +677,16 @@ module OMS
       def create_ods_request(path, record, compress, extra_headers=nil, serializer=method(:parse_json_record_encoding))
         headers = extra_headers.nil? ? {} : extra_headers
 
+        azure_resource_id = OMS::Configuration.azure_resource_id
+        if !azure_resource_id.to_s.empty?
+          headers[OMS::CaseSensitiveString.new("x-ms-AzureResourceId")] = OMS::Configuration.azure_resource_id
+        end
+
+        omscloud_id = OMS::Configuration.omscloud_id
+        if !omscloud_id.to_s.empty?
+          headers[OMS::CaseSensitiveString.new("x-ms-OMSCloudId")] = OMS::Configuration.omscloud_id
+        end
+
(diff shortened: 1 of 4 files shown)


source/code/plugins/oms_common.rb
```diff
@@ -674,9 +674,18 @@ def create_ods_http(ods_uri, proxy={})
       #   serializer: method. serializer of the record
       # returns:
       #   HTTPRequest. request to ODS
-      def create_ods_request(path, record, compress, extra_headers=nil, serializer=method(:parse_json_record_encoding))
+      def create_ods_request(path, omsadmin_conf_path, record, compress, extra_headers=nil, serializer=method(:parse_json_record_encoding))
```
robbiezhang: please use the oms_configuration file to load all configurations in a batch.

robbiezhang:
please use the oms_configuration file to load all configurations in a batch. looking forward to another iteration.

source/code/plugins/oms_common.rb
```diff
@@ -674,9 +674,18 @@ def create_ods_http(ods_uri, proxy={})
...
         headers = extra_headers.nil? ? {} : extra_headers
 
+        File.open(omsadmin_conf_path).each_line do |line|
```
varunkumta: we shouldn't read the conf file for every request .. instead have it static in memory and read the file only if it is not initialized (first time after service is started/restarted)

sugr4:
Used oms_configuration.rb to load values for azureResourceID and omscloudID

source/code/plugins/oms_configuration.rb
```diff
@@ -99,6 +101,15 @@ def load_configuration(conf_path, cert_path, key_path)
           Log.warn_once("Found more than one AGENT_GUID setting in #{conf_path}, will use the first one.")
         end
 
+        File.open(conf_path).each_line do |line|
```
robbiezhang: please move the section down after parsing the AgentId below.

robbiezhang:
Signoff

PR #205, opened 2016-09-16
Fix bugs in onboarding script and service_control

sugr4:
Changes made are:
- service_control enable/start/restart will warn if agent is not onboarded
- omsadmin.sh will register omsagent as a service if onboarding is successful
- added double quotes to if statement for checking if dmidecode exists to
  get rid of the error [: =: unary operator expected

@Microsoft/omsagent-devs


agup006:
+ @jeffaco -- if you have time to review

installer/scripts/omsadmin.sh
```diff
@@ -682,8 +682,11 @@ main()
     fi
 	
     # If we reach this point, onboarding was successful, we can remove the
-    # onboard conf to prevent accidentally re-onboarding
+    # onboard conf to prevent accidentally re-onboarding 
     [ "$ONBOARD_FROM_FILE" = "1" ] && rm "$FILE_ONBOARD" > /dev/null 2>&1 || true
+  
+    # Register omsagent as a service and start the agent
+    /opt/microsoft/omsagent/bin/service_control enable
```
jeffaco: I don't think this is correct. We can onboard (up at line 669), and that can abort if there is a problem (or continue if not).

There's no conditional here. So even if we did NOT onboard, it seems like we would still enable the service, which is not correct here. Why would you not do this only if $ONBOARDING = 1 ?

installer/scripts/service_control
```diff
@@ -45,6 +46,13 @@ is_omsagent_running()
     fi
 }
 
+warn_if_agent_not_onboarded()
+{
+    if [ ! -f $OMSADMIN_CONF ]; then
+       echo "Warning: Agent is not onboarded. omsagent cannot be registered as a service."
+    fi
+}
```
jeffaco: I think a warning here is not enough. If the agent isn't onboarded, you should not continue with code execution.

Thus, you should return a value here and, when calling, abort code execution if we're not onboarded.

test/installer/scripts/prep_omsadmin.sh
```diff
@@ -22,6 +22,9 @@ sed -i s,AGENT_GROUP=.*,AGENT_GROUP=`id -gn`,1 $OMSADMIN
 sed -i s,METACONFIG_PY=.*,METACONFIG_PY=$TESTDIR/not_a_file,1 $OMSADMIN
 sed -i s,PROCESS_STATS=.*,PROCESS_STATS=$TESTDIR/process_stats,1 $OMSADMIN
 
+# Don't run service_control in a test
+sed -i s,/opt/microsoft/omsagent/bin/service_control[.]*,,1 $OMSADMIN
```
jeffaco: Oh, I don't like this one bit, and it's a bad idea:

- It's confusing that we change a shipping file from what is checked out,
- Nothing changes it back (that I see), possibly causing staging problems and shipping the wrong version

If service_control is being run as part of a test, then I think running a test should set an environment variable and whoever is calling service_control shouldn't do that if the "OMSAGENT_TESTING_IN_PROGRESS" is set. Or something like that.

test/installer/scripts/prep_omsadmin.sh
```diff
@@ -22,6 +22,9 @@ sed -i s,AGENT_GROUP=.*,AGENT_GROUP=`id -gn`,1 $OMSADMIN
...
+
 # Don't run omi in a test since it may not be installed.
 # The output generated by omi is mocked in process_stats.
 sed -i s,/opt/omi/bin/omicli.*,,1 $OMSADMIN
```
jeffaco: Same here. Please do not edit shipping files as part of unit tests.

installer/scripts/service_control
```diff
@@ -93,6 +101,7 @@ start_omsagent_process()
 
 start_omsagent()
 {
+    warn_if_agent_not_onboarded
```
jeffaco: Don't allow agent to try and start if we're not onboarded, an error will just result.

installer/scripts/service_control
```diff
@@ -143,6 +152,8 @@ restart_omsagent()
         return
     fi
 
+    warn_if_agent_not_onboarded
```
jeffaco: Why not put this up at line 149? And: abort code flow if not onboarded, an error will just result.

installer/scripts/service_control
```diff
@@ -191,7 +202,9 @@ find_systemd_dir()
 
 enable_omsagent_service()
 {
-    if [ ! -f /etc/opt/microsoft/omsagent/conf/.service_registered ] && [ -f /etc/opt/microsoft/omsagent/conf/omsadmin.conf ]; then
+    warn_if_agent_not_onboarded
```
jeffaco: Since we have a test just below, shouldn't this just be in the `else` case for the `if` at line 207?

sugr4:
@jeffaco made changes based on your comments

installer/scripts/service_control
```diff
@@ -45,6 +46,14 @@ is_omsagent_running()
     fi
 }
 
+exit_if_agent_not_onboarded()
+{
+    if [ ! -f $OMSADMIN_CONF ]; then
+        echo "Warning: Agent is not onboarded. omsagent cannot be registered as a service."
+        exit 1
```
jeffaco: I'm not sure this is a good idea. Did you build a shell bundle, and does this work properly?

The concern is that this can trigger an installation failure (since we're exiting the script with an error) even though this is not actually an error, it's a perfectly normal condition (and frankly, and expected one).

Did you consider emitting the warning and exiting zero? Or, better yet, just return (like 'return') from the routine and trap it from the caller? (Not sure if that's useful, though).

installer/scripts/service_control
```diff
@@ -93,6 +102,7 @@ start_omsagent_process()
 
 start_omsagent()
 {
+    exit_if_agent_not_onboarded
```
jeffaco: Because of prior implementation, you can be in a situation where we are onboarded with no service registered (although that was never released to customers). Did you consider just registering the service prior to a "start" or "restart" request if we are onboarded but not registered?

installer/scripts/service_control
```diff
@@ -215,7 +227,9 @@ enable_omsagent_service()
         fi
         touch  /etc/opt/microsoft/omsagent/conf/.service_registered
         start_omsagent
-    fi
+    else
+        exit_if_agent_not_onboarded
```
jeffaco: We know we're not onboarded at this point, yes? Thus, no need to do the 'if' test again. Not a biggie, just a nit.

installer/scripts/service_control
sugr4: The onboarding file in this commit should always register the service after onboarding, unless there is an error during registration itself. But if we ever do get into that situation we could do away with the enable API and merge it with "start" instead. Then "restart" already includes ''start" so it would also register if we are onboarded but not registered.

installer/scripts/service_control
sugr4: I did build and test a shell bundle and it works find but you are right, emitting 0 would be better.

sugr4: *exiting

jeffaco: Okay, please change to exit 0, it's not an error so you shouldn't return one.

installer/scripts/service_control
jeffaco: Prefer that you verify that the service is enabled before start and restart. So, that is, upon start:

- Register the service if we are onboarded and if service is not yet registered
- Start the service

Verify that restart has similar logic, or doesn't need it if it relies on 'start'.

Enable should still exist, as it makes sense to call that from the post installation scripts. So don't do away with it, just include it if we say "start" and the service isn't enabled.

sugr4: On doing so, both "enable" and "start" would have the exact same steps:
1. Register the service if onboarded and not registered.
2. Start the service
Did you want "enable" to only have the service registration and not service start?

jeffaco: Yes. Start is a subset of enable, but enable should JUST enable and not start.

By the same token, disable is a superset of stop (or, said another way, stop is a subset of disable).

It is incorrect for enable to start the service. Enable should simply perform service registration, but not start the service.

PR #209, opened 2016-09-20
Added two new counter names present in Mongo v2.6

sugr4:
@Microsoft/omsagent-devs

### DIFF
--- a/source/code/plugins/mongostat_lib.rb
+++ b/source/code/plugins/mongostat_lib.rb
@@ -4,36 +4,39 @@ module MongoStatModule
     require_relative 'oms_common'
 
     # sample mongostat record
-    #$ mongostat --all
-    # insert query update delete getmore command flushes mapped  vsize   res non-mapped faults qr|qw ar|aw netIn netOut conn     time
-    # *0    *0     *0     *0       0     2|0       0  80.0M 358.0M 25.0M     278.0M      0   0|0   0|0  158b    20k    2 16:24:07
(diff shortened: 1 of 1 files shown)


source/code/plugins/mongostat_lib.rb
```diff
@@ -4,36 +4,48 @@ class MongoStat
...
       else
-        @values = arr
+        @values = record.split
       end
 
-      if @values and @values.length > 0
+      if @counters && @values && @values.length > 0
+        if @counters.include? 'locked'
+          i = @counters.index('locked')
+          @counters.delete_if {|item| item =~ /locked|db/ }
```
robbiezhang: It's tricky here. Any column called "db" will be removed.

source/code/plugins/mongostat_lib.rb
```diff
@@ -4,36 +4,48 @@ class MongoStat
...
+          @counters.insert(i, "locked db")
+        end
+
+        if @counters.include? 'idx'
+          j = @counters.index('idx')
+          @counters.delete_if {|item| item =~ /idx|miss|%/ }
```
robbiezhang: same here. Any column named "idx" or "miss" or "%" will be removed. Since you already get the index of "idx", can you check whether the @counters[j+1] == "miss" and @counters[j+2] == "%", and merge them when all the check returns true?

sugr4: Moved the checks to when @counters is a string, before it is split to an array

robbiezhang:
Signoff

PR #211, opened 2016-09-21
Include omsagent process stats in telemetry only if onboarded

sugr4:
successfully, to prevent onboarding failure with error 400

@Microsoft/omsagent-devs


robbiezhang:
Signoff

PR #214, opened 2016-09-22
Clarify necessary permissions for log files

lagalbra:
Explicitly state that user needs to be able to "execute" directories
This is necessary to read contained log files for omsagent


lagalbra:
@Microsoft/omsagent-devs

robbiezhang:
Signoff

PR #219, opened 2016-09-23
remove mysql ruby plugin from the bundle

NarineM:
Temp removing MySQL gem and the plugin from the bundle till the issue with CentOS 5's libmysqlclient compatability withUbuntu 14/16 is resolved.
@robbiezhang


robbiezhang:
@Microsoft/omsagent-devs

robbiezhang:
remove the removed ruby and config files from the 
installer/datafiles/base_omsagent.data

robbiezhang:
Remove the install mysql gem in build/buildRuby.sh

NarineM:
Canceling this. Anurag suggested to keep it the way it is.

agup006:
We should keep this work item open, but not have as a deliverable for
Ignite release

PR #222, opened 2016-09-26
create build share param for signing script

NarineM:
Updated the signing script to take the build share path as a parameter and default it to OSTC share so that the change does not require updating OSTC Jenkins configurations. Since OMS Jenkins is dropping builds in a different location, we need to parameterize the build share.


NarineM:
@Microsoft/omsagent-devs

robbiezhang:
Signoff

PR #230, opened 2016-09-28
Move telemetry from onboarding script to separate Ruby script

sugr4:
The append_telemetry logic is moved from the shell script to a separate Ruby script such that we now have a Telemetry object with strongly typed attributes. This is to prevent empty strings from being sent as telemetry attributes in agent topology requests.

@Microsoft/omsagent-devs


installer/scripts/omsadmin.sh
```diff
@@ -29,6 +29,7 @@ INSTALL_INFO=/etc/opt/microsoft/omsagent/sysconf/installinfo.txt
 # Ruby helpers
 RUBY=/opt/microsoft/omsagent/ruby/bin/ruby
 AUTH_KEY_SCRIPT=/opt/microsoft/omsagent/bin/auth_key.rb
+TELEMETRY_SCRIPT=/opt/microsoft/omsagent/bin/telemetry.rb
```
robbiezhang: We won't be able to update the files in the folder from DSC. I think we need to move it to the plugin folder. What's your opinion?

installer/scripts/omsadmin.sh
```diff
@@ -410,7 +364,11 @@ onboard()
     echo "   <FullyQualfiedDomainName>${FQDN}</FullyQualfiedDomainName>" >> $BODY_ONBOARD
```
robbiezhang: can you compose the whole AgentTopologyRequest object in the Ruby? not just the Telemetry part.

installer/scripts/omsadmin.sh
sugr4: Ok. We can use a _script suffix. So telemetry_script.rb

installer/scripts/telemetry.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  File.open(os_info).each_line do |line|
+    telemetry.OSName = line.sub("OSName=","").strip if line =~ /OSName/
+    telemetry.OSManufacturer = line.sub("OSManufacturer=","").strip if line =~ /OSManufacturer/
+    telemetry.OSVersion = line.sub("OSVersion=","").strip if line =~ /OSVersion/
+  end
+end
+        
+def to_xml(telemetry, path)
+  File.open(path, 'a') do |f|
```
robbiezhang: Can you investigate using some library to convert the ruby class into the xml string?

installer/scripts/telemetry.rb
```diff
@@ -0,0 +1,92 @@
+class Telemetry
```
robbiezhang: can you investigate how to generate a strong typed ruby class from an XSD?

installer/scripts/telemetry.rb
```diff
@@ -0,0 +1,92 @@
...
+def to_xml(telemetry, path)
+  File.open(path, 'a') do |f|
+    f << "   <OperatingSystem>\n"
+    f << "      <Name>#{telemetry.OSName}</Name>\n"
+    f << "      <Manufacturer>#{telemetry.OSManufacturer}</Manufacturer>\n"
+    f << "      <ProcessorArchitecture>x64</ProcessorArchitecture>\n"
+    f << "      <Version>#{telemetry.OSVersion}</Version>\n"
+    f << "      <InContainer>#{telemetry.InContainer}</InContainer>\n"
+  end
+  if ENV['TEST_WORKSPACE_ID'].nil? && ENV['TEST_SHARED_KEY'].nil? && File.exist?('/etc/opt/microsoft/omsagent/conf/omsadmin.conf')
```
robbiezhang: make it an input parameter to determine whether to include the telemetry data

installer/scripts/omsadmin.sh
robbiezhang: don't forget to add the file into the base_omsagent.data

installer/scripts/omsadmin.sh
```diff
@@ -410,7 +364,11 @@ onboard()
...
     echo "   <EntityTypeId>$AGENT_GUID</EntityTypeId>" >> $BODY_ONBOARD
     echo "   <AuthenticationCertificate>${CERT_SERVER}</AuthenticationCertificate>" >> $BODY_ONBOARD
-    append_telemetry $BODY_ONBOARD
+
+    # append telemetry to $BODY_ONBOARD
+    `$RUBY $TELEMETRY_SCRIPT "$BODY_ONBOARD" "$OS_INFO" "$PROCESS_STATS"`
+    [ $? -ne 0 ] && return 1
```
niroyb: I don't think we want a telemetry failure to break the whole on-boarding.
Maybe just log an error.

installer/scripts/telemetry.rb
```diff
@@ -0,0 +1,92 @@
...
+def get_telemetry_data(telemetry, os_info, process_stats)
+  if File.exist?('/root/.dockerenv') || File.exist?('/root/.dockerinit')
+    telemetry.InContainer = "True"
+  else
+    telemetry.InContainer = "False"
+  end
+
+  if ENV['TEST_WORKSPACE_ID'].nil? && ENV['TEST_SHARED_KEY'].nil? && File.exist?('/etc/opt/microsoft/omsagent/conf/omsadmin.conf')
+    stats=`/opt/omi/bin/omicli wql root/scx "SELECT PercentUserTime, PercentPrivilegedTime, UsedMemory, PercentUsedMemory FROM SCX_UnixProcessStatisticalInformation where Name like 'omsagent'" | grep =`
+    File.write(process_stats, stats)
```
niroyb: I don't think this has to be written to disk here. It made work easier in bash.

installer/scripts/agent_topology_request.rb
```diff
@@ -0,0 +1,124 @@
...
+    os.Version = line.sub("OSVersion=","").strip if line =~ /OSVersion/
+  end
+end
+
+def obj_to_hash(obj)
+  hash = {}
+  obj.instance_variables.each { |var|
+    val = obj.instance_variable_get(var)
+    if val.is_a? AgentTopologyRequestOperatingSystemTelemetry
+      telemetry_hash = {"Telemetry" => "", :attributes! => {"Telemetry" => obj_to_hash(val) }}
```
robbiezhang: what does :attributes! mean? does it mean all the properties in Telemetry class should be serialized as attribute in xml? if so, I'd prefer to investigate the property decorator for ruby similar to the "Attribute" in C#.

installer/scripts/agent_topology_request.rb
```diff
@@ -0,0 +1,126 @@
+require 'optparse'
+require 'gyoku'
```
sugr4: No, it isn't I will have to add it to the gemlist. It has a dependency called Builder, both are MIT licensed.

installer/scripts/agent_topology_request.rb
sugr4: Yes you are right about the :attributes! part, but it is part of Gyoku's syntax and is not in-built Ruby code. The properties of Telemetry get serialized as xml attributes once this hash is serialized by Gyoku.

sugr4:
Made changes based on comments, added unit tests, added new gems Builder and Gyoku to ruby.data.

build/buildRuby.sh
```diff
@@ -202,6 +202,12 @@ export PATH=${RUBY_DESTDIR}/bin:$PATH
 echo "Installing Bundler into Ruby ..."
 elevate ${RUBY_DESTDIR}/bin/gem install ${BASE_DIR}/source/ext/gems/bundler-1.10.6.gem
 
+echo "Installing Builder into Ruby ..."
+elevate ${RUBY_DESTDIR}/bin/gem install ${BASE_DIR}/source/ext/gems/builder-3.2.2.gem
```
robbiezhang: did you send these 2 gems to @agup006 for review?

sugr4: Yes, he approved.

source/code/plugins/agent_topology_request_script.rb
```diff
@@ -0,0 +1,143 @@
...
+      if (value != 'x64' && value != 'x86')
+        raise ArgumentError, \"Invalid data for ProcessorArchitecture.\"
+      end
+      @#{name}=value
+    end")
+  end
+end
+
+class AgentTopologyRequestOperatingSystemTelemetry < StrongTypedClass
+  strongtyped_accessor :PercentUserTime, Integer
```
robbiezhang: if these numbers are nil or not assigned, how will it be caught?

sugr4: If they are nil or not assigned, telemetry will be nil, but OperatingSystem  info will still go through to the topology request

robbiezhang:
Approve. Make sure it pass the pbuild.

PR #231, opened 2016-09-28
moved an existing parameter to be second

NarineM:
Moved TestSigningDir to take the value of the second parameter since the first parameter of the signing script is used to refer to the build share and it is no longer an empty string. 
@Microsoft/omsagent-devs

### DIFF
--- a/build/signDSCModules.sh
+++ b/build/signDSCModules.sh
@@ -1,5 +1,13 @@
 #!/bin/bash
 
+# This shell script takes two parameters (BUILD_BASEDIR and TestSigningDir).
+# The first parameter (BUILD_BASEDIR) points to the build share where signed package is dropped, 
+# the second parameter (TestSigningDir) points to the path where signing certs are located.
+# These are parameter combinations currently supported by this script:
+#                    Blank (No parameter)
+#                    /mnt/ostcdata/OSTCData/Builds/omsagent/develop (only BUILD_BASEDIR is specified)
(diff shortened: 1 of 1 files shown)


build/signDSCModules.sh
```diff
@@ -64,7 +69,7 @@ DscModuleIntermediateDir="$IntermediateDir/merging_dsc_modules"
 DscModuleTargetDir="$TargetDir/dsc_signed"
 X64Dir="$BUILD_BASEDIR/Linux_ULINUX_1.0_x64_64_Release/dsc"
 X86Dir="$BUILD_BASEDIR/Linux_ULINUX_1.0_x86_32_Release/dsc"
-TestSigningDir=$1
+TestSigningDir=$2
```
jeffaco: Line 23 (now line 28 due to your new comments) is a weird way to handle parameters, and won't really do what you want.

If you want the notion of two separate optional parameters (which is probably correct here), then you should allow them independently of one another. That is:

- No parameters need be specified if you don't need to override,
- Only the build share can be specified,
- Only the test signing dir can be specified,
- Both the build share and test signing dir can be specified.

The way the code is currently structured, this isn't the case today. Please fix.

NarineM: Updated the comments section to specify what is currently supported. Updated line#28 to be consistent with the rest of the script implementation (the way it validate the input parameter). Although existing style can be improved I am not going to re-write the script in this PR since it is only used by test tools. Definitely can come back to the script in the future.

jeffaco: No need to rewrite, what's here is good enough.

PR #236, opened 2016-09-30
Use collectd record's host field in perf,unless if value is localhost

sugr4:
@Microsoft/omsagent-devs

### DIFF
--- a/source/code/plugins/collectd_lib.rb
+++ b/source/code/plugins/collectd_lib.rb
@@ -27,13 +27,15 @@ module CollectdModule
       data_items = []
       data_info = {}
       data_info["Timestamp"] = record["Timestamp"]
-      data_info["Host"] = hostname
+
+      if (record["host"].nil? || record["host"] == "localhost" || record["host"] == "127.0.0.1")
+        data_info["Host"] = hostname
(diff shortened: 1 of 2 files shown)


source/code/plugins/collectd_lib.rb
```diff
@@ -27,7 +27,14 @@ def transform(record, hostname)
       data_items = []
       data_info = {}
       data_info["Timestamp"] = record["Timestamp"]
-      data_info["Host"] = hostname
+
+      host = record["host"] if !record["host"].nil?
```
NarineM: Why "host" starts with lower case in record["host"]  but upper case in data_info? Are these case sensitive?

sugr4: The automatically generated collectd record has keys starting with lowercase. In the perf blob, we just use the convention that the field names start with a capital letter.

NarineM:
I can approve this PR if there is no additional fix or change needed for the parameter 'hostname' when it is localhost or 127.0.0.1.

sugr4:
Deleted the converted properties

robbiezhang:
signoff

PR #237, opened 2016-10-04
Support default Apache access and error logs

lagalbra:


lagalbra:
FluentD plugins to collect schematized Apache HTTP server logs

docs/Apache-HTTP-Server-Logs-Collection.md
```diff
@@ -0,0 +1,40 @@
+# Apache HTTP Server Log Monitoring Solution for Operations Management Suite
+
+1. Set up a supported Linux machine and install an [Apache HTTP Server](http://httpd.apache.org/docs/2.2/install.html).
```
jeffaco: The Apache provider project supports Apache 2.2 and Apache 2.4, FWIW.

docs/Apache-HTTP-Server-Logs-Collection.md
```diff
@@ -0,0 +1,40 @@
...
+  sudo chown omsagent:omiusers /etc/opt/microsoft/omsagent/conf/omsagent.d/apache_logs.conf
+  ```
+
+4. Specify the machine's log file paths in the configuration file `/etc/opt/microsoft/omsagent/conf/omsagent.d/apache_logs.conf` and comment out the other options with `#`:
+
+  ```
+  # Apache Access Log
+  <source>
+    ...
+    path /usr/local/apache2/logs/access_log  #/var/log/apache2/access.log /var/log/httpd/access_log /var/log/apache2/access_log
```
jeffaco: This information is available via the [Apache provider](https://github.com/Microsoft/Apache-Provider#enumeration-of-apache_httpdvirtualhost). Isn't it possible that you could get it from there? That way, the administrator doesn't need to specify it twice, and keep them in sync. If this isn't possible, then it isn't possible. But if it is, that would be a nicer user experience for sure, and you'd automatically pick up changes to it.

Also, this information is on a per-virtual host basis. Given that multiple virtual hosts are common, you should probably clarify how to specify this for multiple virtual hosts.

docs/Apache-HTTP-Server-Logs-Collection.md
lagalbra: Link updated to clarify support.

docs/Apache-HTTP-Server-Logs-Collection.md
lagalbra: The support does not need to go beyond the default scenario for the current scope. Even if the users do not specify their log path, all default system paths listed in the current configuration will be monitored.

PR #238, opened 2016-10-04
bug fix on failure type

keikhara:
@arjun-urs  @amitsara @Microsoft/omsagent-devs 
fixed bug on the parser. Please review. 
-failure type 
-added more filter to remove unnecessary information into OMS


keikhara:
consolidated with #239

PR #240, opened 2016-10-05
Revert "Fix test ruby cleanup to make it reliable"

NarineM:
This reverts commit 84ea4c0fb6178dcc731962f531fd9001f3048eef till all combinations (ruby and its cleanup being built from root or non root users, or both) are tested.


lagalbra:
@Microsoft/omsagent-devs

PR #242, opened 2016-10-06
Improve onboarding failure HTTP403 error message

sugr4:
@Microsoft/omsagent-devs


sugr4:
@niroyb Added the reason in the server response to the error message

PR #243, opened 2016-10-06
mysql ruby plugin improvements

NarineM:
Updated MySQL ruby plugin perf counter names to be consistent with the names used in legacy omi provider. Redirected errors from MySQL gem install to console instead of wrapping it into a custom error.
@Microsoft/omsagent-devs


robbiezhang:
signoff

PR #245, opened 2016-10-17
Docker bundle 1.0.0-14: fix segment fault

robbiezhang:
Fix the invalid memory access issue in docker provider

@Microsoft/omsagent-devs


amitsara:
Approved.

PR #246, opened 2016-10-18
dsc release signing script

NarineM:
Automation to sign DSC modules using azure storage account SAS URLs. 
@Microsoft/omsagent-devs

### DIFF
--- a/build/relsignDSCModules.sh
+++ b/build/relsignDSCModules.sh
@@ -1,0 +1,284 @@
+#!/bin/bash
+
+# Script to sign DSC modules from a Linux box and drop off signed packages (in zip format) into build share.
+# It merges x86 and x 84 DSC modules, then generates a checksum file for it, uploads to azure storage account using SAS URI.
+# Once the checksume file is signed as *.asc file, it downloads and 
+# serializes the signed file into the zip package for given module.
+# Script assumes that source and destination containers are created in azure storage account with proper write and read permissions
+# 
+# Supported parameters
+# $1 - SAS URI of source container which contains the file to be signed
+# $2 - SAS URI of destination container where the signed file is dropped
+# $3 - Build version. Example 1.2.0-118
+# $4 - Build share location to download given build . Example /mnt/omsdata/OMSData/Builds/omsagent/develop
+
+# If no arguments, print usage statement & exit from function
+[ $# -eq 0 ] && { echo "Usage: $0 <sas_source_uri> <sas_destination_uri> <build_version> <build_share_location>"; exit 1; }
+
+StorageVersion="2014-02-14"
+# Max number of attempts to wait for signing worklow to finish 
+MaxSignWaitAttempts=30
+
+# read_blob <URI> <REMOTE_FILE> <DEST_FILE>
+read_blob() {
+
+    [ ! -z "$1" ] || error "read_blob(): expected remote URI" 1
+    [ ! -z "$2" ] || error "read_blob(): expected remote filename" 1
+    [ ! -z "$3" ] || error "read_blob(): expected destination file" 1
+
+    url="$1"
+    remotefile="$2"
+    localfile="$3"
+
+    container=$( echo "$1" | awk -F \?  '{container=index($0,"?"); print $1}' )
+	sas=$( echo "$url" | awk -F \? '{container=index($0,"?"); print $2}' )
+    url="${container}/${remotefile}?${sas}"
+
+    timestamp=$(date +%Y-%m-%dT%H:%M:%SZ)
+    output=$(curl --silent --write-out %{http_code} --request "GET" \
+                  --header "x-ms-date    : $timestamp" \
+                  --header "x-ms-version : $StorageVersion" \
+                  "$url" )
+    
+    code=${output: (-3)}	
+	if [ "$code" != 200 ]; then
+	    echo "read_blob(): error retrieving $remotefile from container $container, server returned an http error code $code"
+        return 1
+    fi
+
+    output=${output: :(-3)}
+    echo -n "$output" > "$localfile"
+   
+    return 0
+}
+
+# write_blob <SASURL> <FILE>
+write_blob() {
+
+    [ ! -z "$1" ] || error "write_blob(): expected remote URI" 1
+    [ ! -z "$2" ] || error "write_blob(): expected filename to upload" 1
+    [ -f "$2" ] || error "write_blob(): file $2 does not exist" 1
+
+    url="$1"
+    localfile="$2"
+    remotefile=$(basename "$localfile")
+
+    container=$( echo "$1" | awk -F \?  '{container=index($0,"?"); print $1}' )
+    sas=$( echo "$url" | awk -F \? '{container=index($0,"?"); print $2}' )
+    url="${container}/${remotefile}?${sas}"
+
+    timestamp=$(date +%Y-%m-%dT%H:%M:%SZ)
+    output=$(curl --silent --write-out %{http_code} --request "PUT" \
+                  --header "x-ms-date      : $timestamp" \
+                  --header "x-ms-blob-type : BlockBlob" \
+                  --header "x-ms-version   : $StorageVersion" \
+                  --data-binary "@${localfile}" "$url" )
+    
+    code=${output: (-3)}
+	if [ "$code" != 201 ]; then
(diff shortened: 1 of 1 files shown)


build/relsignDSCModules.sh
```diff
@@ -0,0 +1,309 @@
...
+echo " SAS destination container URL is : $2"
+# Build info
+BUILD_VERS=$3
+BUILD_BASEDIR=$4
+
+if [ "$BUILD_BASEDIR" = "" ]; then
+
+    echo "No build share path is specified. Default build share is mounted at: /mnt/omsdata"
+
+    BUILD_BASEDIR=/mnt/omsdata/OMSData/Builds/omsagent/develop
```
jeffaco: So this won't work on OSTC build servers, right, since the path is wrong?

This is intended once the OMS team takes over the OMS builds?

build/relsignDSCModules.sh
```diff
@@ -0,0 +1,309 @@
...
+delete_blob() {
+
+    [ ! -z "$1" ] || error "delete_blob(): expected remote URI" 1
+    [ ! -z "$2" ] || error "delete_blob(): expected remote filename" 1
+   
+    url="$1"
+    remotefile="$2"
+    
+    container=$( echo "$1" | awk -F \?  '{container=index($0,"?"); print $1}' )
+	echo "container is $container"
```
jeffaco: Inconsistent indentation due to tabs. I'm okay with this (it shows up properly in Emacs or some reasonable editor). But just FYI, we often tell our editors to space indent to avoid indentation issues across a variety of editors (particularly this on Windows).

build/relsignDSCModules.sh
```diff
@@ -0,0 +1,309 @@
...
+    sas=$( echo "$url" | awk -F \? '{container=index($0,"?"); print $2}' )
+    url="${container}/${remotefile}?${sas}"
+
+    timestamp=$(date +%Y-%m-%dT%H:%M:%SZ)
+    output=$(curl --silent --write-out %{http_code} --request "GET" \
+                  --header "x-ms-date    : $timestamp" \
+                  --header "x-ms-version : $StorageVersion" \
+                  "$url" )
```
jeffaco: I don't understand the protocol, so I won't be reviewing this for "correctness" in signing - I'll leave that to someone else more familiar.

build/relsignDSCModules.sh
```diff
@@ -0,0 +1,309 @@
...
+
+
+
+if [ ! -d $BUILD_BASEDIR ]; then
+
+    echo "Not finding Build share mounted at $BUILD_BASEDIR ..."
+
+    exit 1
+
+fi
```
jeffaco: You don't really need all these blank lines, and just makes the script longer. If I wrote this, lines 129, 131, and 133 wouldn't exist (here and in similar cases below.

build/relsignDSCModules.sh
```diff
@@ -0,0 +1,309 @@
...
+
+
+
+zip -L >/dev/null 2>&1
+
+if [ $? != 0 ]; then
```
jeffaco: These kinds of tests can be done on a single line. This would allow the script to run with error detection, which I think tends to make for a more reliable script (with less "unintended allowed errors). As written, you couldn't enable error detection.

build/relsignDSCModules.sh
```diff
@@ -0,0 +1,309 @@
...
+
+    exit 1
+
+fi
+
+
+
+unzip -Z >/dev/null 2>&1
+
+if [ $? != 0 ]; then
```
jeffaco: Ditto with above, consider a single line if test for this and enabling error detection for the script.

build/relsignDSCModules.sh
```diff
@@ -0,0 +1,309 @@
...
+
+fi
+
+# Working directory used for merging both x64 and x86 versions of a DSC module.
+DscModuleIntermediateDir="tmp/merging_dsc_modules"
+# Working directory where the signed package is dropped.
+DscModuleTargetDir="tmp/dsc_signed"
+# Locations of dsc modules that need to be signed
+X64Dir="$BUILD_BASEDIR/Linux_ULINUX_1.0_x64_64_Release/dsc"
+X86Dir="$BUILD_BASEDIR/Linux_ULINUX_1.0_x86_32_Release/dsc"
```
jeffaco: Blech (choke, cough). There are issues that I don't fully understand here with signing, but it would be VERY good if 32-bit and 64-bit packages could be signed independently.

That would allow you to sign packages during the regular build process (when you have access to the packages from the current working directory).

The advantage here is much simpler build host setup (no need to mount the build publishing disk to re-read the builds). That's compelling, if you can possibly swing it.

build/relsignDSCModules.sh
```diff
@@ -0,0 +1,309 @@
+#!/bin/bash
+
+# Script to sign DSC modules and drop off signed packages (in zip format) into build share.
+# It generates a checksum file for each DSC module, uploads to azure storage account using SAS_SOURCE_URI,
+# then downloads the signed *.asc file from SAS_DEST_URI and 
+# serializes it into the zip package for given DSC module.
```
jeffaco: What's the usage here? You never say how the script should be run and what, if any, parameters should be passed in.

You also never say how this is intended to be called. Is this called directly from Jenkins? Is it embedded elsewhere?

build/relsignDSCModules.sh
```diff
@@ -0,0 +1,309 @@
...
+
+# Working directory used for merging both x64 and x86 versions of a DSC module.
+DscModuleIntermediateDir="tmp/merging_dsc_modules"
+# Working directory where the signed package is dropped.
+DscModuleTargetDir="tmp/dsc_signed"
+# Locations of dsc modules that need to be signed
+X64Dir="$BUILD_BASEDIR/Linux_ULINUX_1.0_x64_64_Release/dsc"
+X86Dir="$BUILD_BASEDIR/Linux_ULINUX_1.0_x86_32_Release/dsc"
+
+set -e
```
jeffaco: Nice that you run with error detection. But it would be much nicer if this were done at the very top, and the entire script ran that way. Easy to do as long as your `if` tests were coded properly for testing, for example, if `zip` was installed.

build/relsignDSCModules.sh
NarineM: we (oms ) are calling it from jenkins, but the script is written in such a way it can be called from anywhere since it  does not have any GitHub project hierarchical dependencies whatsoever. I will add more details in the description if it would help.

build/relsignDSCModules.sh
NarineM: Will fix the spaces for consistency.

build/relsignDSCModules.sh
NarineM: This is just a default. The default really is not needed, but just sake of convenience. It can be used for OSTC builds as well as long as you pass parameter $4.

build/relsignDSCModules.sh
NarineM: Ah, will remove the blank lines. thanks

build/relsignDSCModules.sh
NarineM: Ok, will remove redirecting the error. good point. thanks.

build/relsignDSCModules.sh
NarineM: Merging x64 and x86 is now done because of the deployment constraints. Most likely can be removed in the future.

build/relsignDSCModules.sh
jeffaco: Definitely, you need usage here. It should be trivial to look at a script to understand usage, or run a script with `-h` or `-?` to get usage.

Does this script replace the existing signing script? If not, when is this used vs. the existing signing script? If this replaces the existing script, then shouldn't you remove that as part of this pull request?

jeffaco: Definitely, you need usage here. It should be trivial to look at a script to understand usage, or run a script with `-h` or `-?` to get usage.

Does this script replace the existing signing script? If not, when is this used vs. the existing signing script? If this replaces the existing script, then shouldn't you remove that as part of this pull request?

build/relsignDSCModules.sh
NarineM: I was told that the current req is to merge the modules , and then sign. If this constraints change in the future, we will need to update the automation. I will add a comment here in the script.

PR #248, opened 2016-10-19
updated default path for signing

NarineM:
Updated default path for signing to keep automation changes minimal. This was the initial path I had, so bringing it back.  
@Microsoft/omsagent-devs

### DIFF
--- a/build/relsignDSCModules.sh
+++ b/build/relsignDSCModules.sh
@@ -5,6 +5,7 @@
 # Once the checksume file is signed as *.asc file, it downloads and 
 # serializes the signed file into the zip package for given module.
 # Script assumes that source and destination containers are created in azure storage account with proper write and read permissions
+# Script is expected to run from Build-OMS-Agent-for-Linux path
 # 
 # Supported parameters
 # $1 - SAS URI of source container which contains the file to be signed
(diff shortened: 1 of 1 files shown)


build/relsignDSCModules.sh
```diff
@@ -155,10 +155,10 @@ if ! unzip -Z 1>/dev/null; then
 fi
 
 # Working directory used for merging both x64 and x86 versions of a DSC module.
-DscModuleIntermediateDir="/tmp/merging_dsc_modules"
+DscModuleIntermediateDir="tmp/merging_dsc_modules"
```
jeffaco: It's not clear where this is run from, so it's not clear where this directory is created. It's important that this is either created in the intermediate directory, or that this directory be added to .gitignore for the project.

I'd like to see documentation on where this should be run from (up top), and I'd like to see verification that you're actually running from there. For example, if the intent is that you're running from the build directory, then verify that you really are running from the build directory.

If the intent is that this is run from the build directory, then these paths definitely need to be changed so they're created in the intermediate directory, or otherwise cleaned up (via make clean/git nuke/whatever).

NarineM: Intent was to implement the script in such a way it can be run from any location. The output is created under tmp in script's current location. Is there any extra cleanup step I need to have here for nuke/clean or it will be picked up automatically?

NarineM:
@jeffaco 
After giving it a second thought, I do not see anything wrong assuming that the working dir is the build dir, since the script is located in .\build. I can update it to be ./intermediate/tmp if you agree with this. Let me know which one you like the most.

jeffaco:
The problem is that it creates "junk" it a git repository. The "git status" command will show these files and won't understand what they are, thinking maybe they should be added?

Best to put them in the intermediate directory. Everything under there is considered "junk" from a source code perspective.

NarineM:
Ok, just checkin the change that moves the output dir to be under intermediate. It also does not require any changes in Jenkins' current signing jobs. Which is convenient :)

PR #249, opened 2016-10-19
local ruby cleanup fix

NarineM:
This change is for making local ruby installation reliable. Before the change the local ruby was being built under system local path which caused unexpected results when 2 different people run buddy build on pBuild systems, or sometimes 'make distclean' would not be able to delete the path in clean up phase leaving the ruby corrupted even on local dev boxes.

@Microsoft/omsagent-devs

### DIFF
--- a/build/configure
+++ b/build/configure
@@ -389,7 +389,7 @@ ruby_configure_quals=(
      )
 
 ruby_config_quals_sysins="--prefix=/opt/microsoft/omsagent/ruby"
-ruby_config_quals_testins="--prefix=/usr/local/ruby-2.2.0e"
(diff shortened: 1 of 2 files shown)


build/configure
```diff
@@ -389,7 +389,7 @@ ruby_configure_quals=(
      )
 
 ruby_config_quals_sysins="--prefix=/opt/microsoft/omsagent/ruby"
-ruby_config_quals_testins="--prefix=/usr/local/ruby-2.2.0e"
+ruby_config_quals_testins="--prefix=${base_dir}/intermediate/${BUILD_CONFIGURATION}/local_ruby-2.2.0e"
```
jeffaco: I suggest getting rid of "e" at the end, you really don't need it. That's noise from a system installation, and is no longer relevant here.

PR #250, opened 2016-10-19
vmware: fixed regex, skipping logs from blacklisted processes.

amitsara:
@Microsoft/omi-devs 
@keikhara 

Fixes:
1. Removed '[12345]' from the process name.
2. Fixed the ESXIFailure regex to match exact data.
3. Skipping the logs generated from 'crond', 'syslog', 'root', 'slpd', 'omsagent' and 'systemd' processes.


keikhara:
Look good sign-off.

PR #251, opened 2016-10-20
Relocate test version of Ruby to the user's ~/bin directory

jeffaco:
It appears that 'gem' has an issue with very long paths. If Jenkins
builds a version of Ruby in the intermediate directory, the path can
be about 120 bytes long, and 'gem' crashes (even when invoked with
no parameters).

We modify configure to place it in ~/bin and configure will clean it
out when run (so it is always rebuild when we run configure).


jeffaco:
This was pair-programmed with Narine; she signs off 😄

PR #260, opened 2016-11-02
Bug Fix: Omsagent changes to Split Json for different change tracking…

gauravga:
Omsagent changes and tests for corresponding changes in powershell DSC
to split json for change tracking. 

Bug fix for change tracking.
After inventory mof for file change tracking was introduced, the json produced
by filter_changetracking either contains service and package changes together or only
file changes. This confuses the back end into thinking that there are changes
across all types. The fix is to split the json for each change type

@robbiezhang @vrdmr 
@Microsoft/omsagent-devs

### DIFF
--- a/installer/conf/omsagent.d/change_tracking.conf
+++ b/installer/conf/omsagent.d/change_tracking.conf
@@ -1,13 +1,13 @@
 <source>
   type exec
-  tag oms.changetracking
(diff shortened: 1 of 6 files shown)


gauravga:
Starting a new pull request with my changes merged with varad's changes as we are changing same files and had merge conflicts

PR #262, opened 2016-11-04
(Merged) Moved the XML to JSON transformation from filter to source to

gauravga:
This pull request contains merge of two changes
1) Changes to moved the XML to JSON transformation from filter to source to
resolve memory issues. (Varad's fix)

2) Changes to solve the bug which introduced changes in the UI, even when there
weren't any changes. These changes have already been approved in a seperate pull request (my fix)

since these changes had merge conflicts and affected each other, we combined them in a single pull request.


gauravga:
@robbiezhang @vrdmr @Microsoft/omsagent-devs

PR #263, opened 2016-11-05
Fix the fluentd reference to v0.12.24

robbiezhang:
The fluentd repo was reverted back to the v0.12.14 in the previous commit.
Revert it back to reference v0.12.24.


robbiezhang:
@Microsoft/omsagent-devs

vrdmr:
Sign off. Thanks Robbie.

PR #269, opened 2016-11-08
Add single script to download install and onboard the agent

lagalbra:
This script will execute both commands necessary to get the OMSAgent for Linux:
wget ....
sudo sh omsagent-*....
The main page README has been updated to reference this quicker method

### DIFF
--- a/installer/scripts/onboard_agent.sh
+++ b/installer/scripts/onboard_agent.sh
@@ -1,0 +1,80 @@
+#!/bin/sh
+
+#
+# Easy download/install/onboard script for the OMSAgent for Linux
+#
+
+
+# Values to be updated upon each new release
+GITHUB_RELEASE="https://github.com/Microsoft/OMS-Agent-for-Linux/releases/download/OMSAgent-201610-v1.2.0-148/"
+BUNDLE_X64="omsagent-1.2.0-148.universal.x64.sh"
+BUNDLE_X86="omsagent-1.2.0-148.universal.x86.sh"
+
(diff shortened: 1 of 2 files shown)


lagalbra:
@robbiezhang @sugr4

installer/scripts/onboard_agent.sh
```diff
@@ -0,0 +1,79 @@
...
+done
+
+
+# Parameter checks
+if [ -z "$onboardID" -o -z "$onboardKey" ]; then
+    echo "Must specify both workspace ID (--id) and key (--shared) to onboard" 1>& 2
+    exit 1
+fi
+
+bundleParameters="--upgrade -w $onboardID -s $onboardKey"
```
sugr4: upgrade or install?

lagalbra: Robbie has recommended upgrade rather than install for all scenarios. The current copyable commands in the portal are upgrade.

installer/scripts/onboard_agent.sh
```diff
@@ -0,0 +1,79 @@
...
+         *)
+            echo "Unknown argument: '$1'" >&2
+            echo "Use -h or --help for usage" >&2
+            exit 1
+            ;;
+    esac
+done
+
+
+# Parameter checks
```
lagalbra: You're right; I removed the checks and only passed the workspace/key/domain parameters that had been specified by the user, so "-w -s" won't be passed.

PR #270, opened 2016-11-08
Fix hostname for containerized omsagent

amitsara:
@Microsoft/omsagent-devs 
@keikhara 

# Issue:
        When omsagent runs inside a container, gethostname returns the hostname of the container (random name) not the actual machine hostname.
        One way to solve this problem is to set the container hostname same as machine name, but this is not possible when host-machine is a private VM inside a cluster.

# Solution:
        Share/mount ‘/etc/hostname’ file with container and omsagent will read hostname from shared file.


source/code/plugins/oms_common.rb
```diff
@@ -577,6 +577,24 @@ def get_os_version(conf_path = "/etc/opt/microsoft/scx/conf/scx-release")
...
+        # Issue:
+        #   When omsagent runs inside a container, gethostname returns the hostname of the container (random name)
+        #   not the actual machine hostname.
+        #   One way to solve this problem is to set the container hostname same as machine name, but this is not
+        #   possible when host-machine is a private VM inside a cluster.
+        # Solution:
+        #   Share/mount ‘/etc/hostname’ file with container and omsagent will read hostname from shared file.
+
+        begin
+          fileHostName = '/etc/hostname'
```
robbiezhang: please add the unit test

amitsara: Added unit test for this change.

test/code/plugins/oms_common_test.rb
```diff
@@ -172,11 +172,17 @@ def test_get_os_version_missing_field()
 
     def test_get_hostname
       $log = MockLog.new
+      file_hostname = '/etc/hostname'
```
robbiezhang: you need to have at least 2 test cases: with the /etc/hostname, and w/o /etc/hostname.

amitsara: added tests to verify scenarios when host file exist and when doesn't.
please note that the container file host path is changed to /etc/containerhostname

PR #271, opened 2016-11-09
Add hdinsight filter plugin

duoxu:
This is the filter plugin for HDInsight clusters. When the filter is
used, it will add cluster information as part of json output.

### DIFF
--- a/source/code/plugins/filter_hdinsight.rb
+++ b/source/code/plugins/filter_hdinsight.rb
@@ -1,0 +1,44 @@
+module Fluent
+  require_relative 'omslog'
+  require_relative 'oms_common'
+  require 'json'
+  require 'open3'
+
+
+  class HdinsightFilter < Filter
+    Fluent::Plugin.register_filter('filter_hdinsight', self)
+
+    BASE_DIR = File.dirname(File.expand_path('..', __FILE__))
+    RUBY_DIR = BASE_DIR + '/ruby/bin/ruby '
+    SCRIPT = BASE_DIR + '/bin/hdinsightmanifestreader.rb '
+	
+	attr_accessor :command
+
+    def configure(conf)
+      super
+      @hostname = OMS::Common.get_hostname or "Unknown host"
+      @command = "sudo " << RUBY_DIR << SCRIPT
+      @clustername = ""
+    end
+
+    def start
+      super
+      Open3.popen3(@command) {|stdin, stdout, stderr, wait_thr|
+          pid = wait_thr.pid
+          stdin.close
+          @clustername = stdout.read
+          wait_thr.value
+      }	  
+    end
+
+    def shutdown
+      super
+    end
+
+    def filter(tag, time, record)
+      record["ClusterName"] = @clustername
+      record["HostName"] = @hostname
+      record
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


duoxu:
@robbiezhang  @agup006 

Please take a look. Thanks!

robbiezhang:
@Microsoft/omsagent-devs

source/code/plugins/filter_hdinsight.rb
```diff
@@ -0,0 +1,42 @@
...
+    def start
+      super
+    end
+
+    def shutdown
+      super
+    end
+
+    def filter(tag, time, record)
+      record["ClusterName"] = @clustername
```
NarineM: Should not there be a check on stdout.read nil? in case the clustername is returned as null , should not it skip sending the record?

source/code/plugins/filter_hdinsight.rb
```diff
@@ -0,0 +1,42 @@
...
+      record["HostName"] = @hostname
+      record
+    end
+  end
+end
```
NarineM: Can you add some basic tests for this plugin?

source/code/plugins/filter_hdinsight.rb
duoxu: The clustername is got from our python API which read a manifest file on the same VM. So it should not return null.

source/code/plugins/filter_hdinsight.rb
NarineM: you can run individual test, but it is good to run the new tests with the rest of collateral to make sure it is not breaking other tests, >make test to run all unit tests altogether.

duoxu: I looked at my filter again. In configure step, it calls our Python APIs to get cluster name and in filter step it appends the clustername to the record. So the actual logic is in our Python APIs, I do not know what kind of test I can write here for this filter. Any ideas?

@robbiezhang @agup006 @NarineM

NarineM: Add a test case that validates the cluster info is part of the record. If Open3.popen3 works only in cluster deployment, may be mock this interface?

duoxu: I looked at all the existing tests, there are no filter tests at all. All the existing tests are library tests, which the filters use. HDInsight filter calls a API from our python library. The filter just appends the output to the record. So I would suggest no test is needed for this filter. 

@robbiezhang @NarineM @agup006

NarineM: @robbiezhang 
I am going to leave it to Robbie to approve if he is OK with not having any tests for this plugin.

robbiezhang: You can mock the call to python script in the unit test and verify the logic of enriching the data.
You can reference test/installer/scripts/tailfilereader_test.rb to see how it test the tailfilereader.rb

duoxu: I have added a unit test. Please take a look.

duoxu:
@robbiezhang @NarineM 

Thanks for the review. Could you merge this into master branch? I do not have write access to this repo.

PR #274, opened 2016-11-10
Add new mongostat field names for Ubuntu distro

sugr4:
Modify conf for mongo < v3.0 to include query details

@Microsoft/omsagent-devs

### DIFF
--- a/source/code/plugins/mongostat_lib.rb
+++ b/source/code/plugins/mongostat_lib.rb
@@ -27,6 +27,8 @@ module MongoStatModule
       if record.start_with?('insert')
         record.sub!("locked db", "locked-db") if record.include? "locked db"
         record.sub!("idx miss %", "idx-miss-%") if record.include? "idx miss %"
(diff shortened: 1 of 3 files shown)


source/code/plugins/mongostat_lib.rb
```diff
@@ -64,19 +66,21 @@ def transform_data(record)
...
+              local, replicated = command.split("|")
+              transformed_rec["Local Commands/sec"] = local
+              transformed_rec["Replicated Commands/sec"] = replicated
+            else
+              transformed_rec["Commands/sec"] = command
+            end
           end
 
           #version >= 3 rec has the counternames locked, dirty, used, non-mapped, flushes
-          transformed_record["% Time Global Write Lock"] = rec["locked"] if rec.has_key?("locked")
```
robbiezhang: interesting. no test case caught this? can you add one?

sugr4: Added test case

PR #278, opened 2016-11-15
Log name of file being tailed by in_sudo_tail plugin

sugr4:
@Microsoft/omsagent-devs

### DIFF
--- a/source/code/plugins/in_sudo_tail.rb
+++ b/source/code/plugins/in_sudo_tail.rb
@@ -4,6 +4,7 @@ require 'fluent/event'
 require 'fluent/config/error'
 require 'fluent/parser'
 require 'open3'
+require_relative 'omslog'
 
 module Fluent
   class SudoTail < Input
@@ -12,6 +13,7 @@ module Fluent
     def initialize
       super
       @command = nil
+      @paths = []
     end
 
     attr_accessor :command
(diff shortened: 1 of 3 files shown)


source/code/plugins/in_sudo_tail.rb
```diff
@@ -101,6 +101,10 @@ def receive_log(line)
     end
  
     def run_periodic
+      @paths = @path.split(' ').each {|path|
+        $log.info "Following tail of #{path}"
```
niroyb: If you don't want redundant repetitive logs, consider using something like warn_once in omslog.rb but for info level.

robbiezhang: Ideally, it will be only called when agent restarts, which is the same as using warn_once. So, it should be fine to use $log.info. But, yes, it's better to de-dup the paths if there are duplicates.

source/code/plugins/in_sudo_tail.rb
```diff
@@ -58,7 +58,7 @@ def configure(conf)
  
       @parser = Plugin.new_parser(conf['format'])
       @parser.configure(conf)
-
+      @path.gsub!("," , " ")       #in case multiple paths are separated by commas
```
robbiezhang: edge case: the file path can contain , or space. although they are really rare.
do we allow wild card in the path? /etc/*.log

sugr4: I've updated changes to log each filename for the case when path includes *
I am currently assuming that in case of multiple paths, they are separated by comma and the file path does not contain commas (as in the original tail plugin). We could change the "path" field in this input plugin to an array instead of a string like so: path ["/var/log/messages*", "/var/log/audit/audit.log"]

PR #281, opened 2016-11-17
Adding NPMD solution input plugin and conf file.

chparimi:
This pull request contains two files related to Network Performance Monitoring and Diagnostics solution for Linux.
- File npmd.conf is the configuration file
- File in_npmd.rb is the input plugin file

Functionality of the input plugin
1. It serves as a control center to start/stop/restart the npmd_agent binary on premise
2. It communicates with DSC python scripts over unix socket and also communicates with npmd_agent binary over the same
3. It converts the configuration provided by DSC into a valid XML which the agent understands
4. It pushes the data uploaded by the agent to ODS after appending certain fields
5. Error logs are pushed up to HEALTH_ASSESSMENT_BLOB right now

Note: Intention is to download this plugin via DSC channel when NPMD Solution is selected on OMS

### DIFF
--- a/installer/conf/omsagent.d/npmd.conf
+++ b/installer/conf/omsagent.d/npmd.conf
@@ -1,0 +1,9 @@
+<source>
+    type npmd
+    tag oms.npmd
+    location_unix_endpoint  "/var/opt/microsoft/omsagent/run/npmdagent.sock"
+    location_control_data   "/etc/opt/microsoft/omsagent/conf/npmd_agent_config.xml"
+    location_agent_binary   "/opt/microsoft/omsagent/npmd_agent"
+    location_uuid_file      "/etc/opt/microsoft/omsagent/conf/npmd_agent_guid.txt"
+</source>
+
--- a/source/code/plugins/in_npmd_server.rb
+++ b/source/code/plugins/in_npmd_server.rb
@@ -1,0 +1,453 @@
+# The npmd input plugin to fluentd
+module Fluent
+    class NPM < Input
+        Fluent::Plugin.register_input('npmd', self)
+
+        unless method_defined?(:router)
+            define_method("router") {Fluent::Engine}
+        end
+
+        def initialize
+            super
+            require 'socket'
+            require 'fileutils'
+            require 'json'
+            require 'securerandom'
+
+            require_relative 'npmd_config_lib'
+        end
+
+        config_param :location_unix_endpoint, :string, :default => "/var/opt/microsoft/omsagent/run/npmdagent.sock"
+        config_param :location_control_data,  :string, :default => "/etc/opt/microsoft/omagent/conf/npmd_agent_config.xml"
+        config_param :location_agent_binary,  :string, :default => "/opt/microsoft/omsagent/npmd_agent"
+        config_param :location_uuid_file,     :string, :default => "/etc/opt/microsoft/omsagent/conf/npmd_agent_guid.txt"
+        config_param :tag, :string, :default => "oms.npmd"
+        def configure(conf)
+            super
+
+            unless @location_unix_endpoint
+                raise ConfigError, "'location_unix_endpoint' is needed for agent communication"
+            end
+
+            unless @location_control_data
+                raise ConfigError, "'location_control_data' is needed to send config to agent"
+            end
+
+            unless @location_agent_binary
+                raise ConfigError, "'location_agent_binary' is needed to invoke the agent"
+            end
+
+            unless @location_uuid_file
+                raise ConfigError, "'location_uuid_file' is needed for storing the agent GUID"
+            end
+        end
+
+        attr_accessor :agentId
+        attr_accessor :binary_presence_test_string
+        attr_accessor :binary_invocation_cmd
+        attr_accessor :npmdClientSock
+        attr_accessor :num_path_data, :num_agent_data
+        attr_accessor :num_config_sent
+        attr_accessor :is_purged
+
+        CMD_START         = "StartNPMD"
+        CMD_STOP          = "StopNPMD"
+        CMD_CONFIG        = "ConfigNPMD"
+        CMD_PURGE         = "PurgeNPMD"
+        CMD_LOG           = "ErrorLog"
+
+        NPMD_CONN_CONFIRM = "NPMDAgent Connected!"
+
+        STOP_SIGNAL       = "SIGKILL"
+
+        RETRY_START_BACKOFF_TIME_SECS   = 900 # 15 minutes
+        RETRY_START_WAIT_TIME_SECS      = 5   # 5 seconds
+        RETRY_START_ATTEMPTS_PER_BATCH  = 5
+
+        CMD_ENUMERATE_AGENT_PROCESSES = "ps aux | grep npmd"
+
+        def start
+            @binary_presence_test_string = "npmd_agent" if @binary_presence_test_string.nil?
+            @binary_invocation_cmd = @location_agent_binary if @binary_invocation_cmd.nil?
+            kill_all_agent_instances()
+            check_and_update_binaries()
+            setup_endpoint()
+            @fqdn = get_fqdn()
+            check_and_get_guid()
+            @server_thread = Thread.new(&method(:server_run))
+            @npmdIntendedStop = false
+            @stop_sync = Mutex.new
+            # Initialize to true to prevent wait on first start
+            @npmdExitHandled = true
+            @wait_to_collect_stderror = ConditionVariable.new
+            start_npmd_async_retry_thread()
+        end
+
+        def shutdown
+            Logger::logInfo "Received shutdown notification"
+            stop_npmd()
+            unless @npmdClientSock.nil?
+                @npmdClientSock.close()
+                @npmdClientSock = nil
+            end
+            Thread.kill(@server_thread)
+            File.unlink(@location_unix_endpoint) if File.exist?@location_unix_endpoint
+            File.unlink(@location_agent_binary) if File.exist?@location_agent_binary and (!@do_purge.nil? and @do_purge)
+            File.unlink(@location_uuid_file) if File.exist?@location_uuid_file and (!@do_purge.nil? and @do_purge)
+            @is_purged = true unless @is_purged.nil?
+        end
+
+        def setup_endpoint
+            begin
+                _dirname = File.dirname(@location_unix_endpoint)
+                unless File.directory?(_dirname)
+                    FileUtils.mkdir_p(_dirname)
+                end
+                if File.exists?(@location_unix_endpoint)
+                    File.unlink(@location_unix_endpoint)
+                end
+                @server_obj = UNIXServer.new(@location_unix_endpoint)
+            rescue StandardError => e
+                log_error "Got error #{e}", Logger::resc
+            end
+        end
+
+        def get_fqdn
+            _ip = nil
+            _name = nil
+            Socket.ip_address_list.each do |addrinfo|
+                next unless addrinfo.ip?
+                if addrinfo.ipv4?
+                    next if addrinfo.ipv4_loopback? or
+                        addrinfo.ipv4_multicast?
+                else
+                    next if addrinfo.ipv6_linklocal? or
+                        addrinfo.ipv6_loopback? or
+                        addrinfo.ipv6_multicast?
+                end
+                _ip = addrinfo.ip_address
+                _name = addrinfo.getnameinfo.first
+                break if _ip != _name
+            end
+            _name
+        end
+
+        def check_and_get_json(text)
+            begin
+                _json = JSON.parse(text)
+                _json
+            rescue JSON::ParserError => e
+                Logger::logInfo "Json parsing failed for #{text} because of #{e}", Logger::resc
+                nil
+            end
+        end
+
+        def log_error(msg, depth=0)
+            Logger::logError(msg, depth + 1)
+            package_and_send_error_log(msg)
+        end
+
+        def kill_all_agent_instances
+            _resultStr = `#{CMD_ENUMERATE_AGENT_PROCESSES}`
+            return if _resultStr.nil?
+            _lines = _resultStr.split("\n")
+            _lines.each do |line|
+                if line.include?@binary_presence_test_string
+                    _staleId = line.split()[1]
+                    Process.kill(STOP_SIGNAL, _staleId.to_i)
+                end
+            end
+        end
+
+        def npmd_reader
+            begin
+                begin
+                    _line = @npmdClientSock.gets
+                    if _line.nil? and !is_npmd_seen_in_ps()
+                        @npmdClientSock = nil
+                        Logger::logInfo "Exiting reader thread as npmdAgent found stopped"
+                        break
+                    end
+                    next if _line.nil?
+                    _json = check_and_get_json(_line.chomp)
+                    unless !_json.nil?
+                        Logger::logWarn "Sent string to plugin is not a json string", Logger::loop
+                    else
+                        unless _json.key?("DataItems") and !_json["DataItems"].nil? and _json["DataItems"] != ""
+                            Logger::logWarn "No valid data items found in sent json #{_json}", Logger::loop
+                        else
+                            _uploadData = _json["DataItems"].reject {|x| x["SubType"] == "ErrorLog"}
+                            _errorLogs  = _json["DataItems"].select {|x| x["SubType"] == "ErrorLog"}
+                            _uploadData.each do |item|
+                                if item.key?("SubType")
+                                    # Append FQDN to path data
+                                    if !@fqdn.nil? and item["SubType"] == "NetworkPath"
+                                        @num_path_data += 1 unless @num_path_data.nil?
+                                        item["Computer"] = @fqdn
+                                        # Append agent Guid to agent data
+                                    elsif !@agentId.nil? and item["SubType"] == "NetworkAgent"
+                                        @num_agent_data += 1 unless @num_agent_data.nil?
+                                        item["AgentId"] = @agentId
+                                    end
+                                end
+                            end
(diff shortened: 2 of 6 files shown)


installer/conf/omsagent.d/npmd.conf
```diff
@@ -0,0 +1,14 @@
+<source>
+    type npmd
+</source>
+
+<match npmd.oms.blob>
```
gauravga: Have you tested that this works? where is this tag defined? is this an existing tag?

gauravga:
Where are the tests?

gauravga:
The commit is incomplete, you also need to add these files to the right module, so they can land on agents that were released before.

dineshkarora:
Adding @Microsoft/omsagent-devs  for review of this PR

source/code/plugins/in_npmd.rb
```diff
@@ -0,0 +1,676 @@
+class Logger
```
NarineM: This file seems too big... can you split the business logic from configuration and helpers?

source/code/plugins/in_npmd.rb
```diff
@@ -0,0 +1,676 @@
...
+        _methodname = getMethodname(depth)
+        _message = "[#{_methodname}]:#{msg}"
+        $log.error "[NPMD]:#{_message}"
+    end
+    def self.logInfo(msg, depth=0)
+        _methodname = getMethodname(depth)
+        $log.info "[NPMD]:[#{_methodname}]:#{msg}"
+    end
+    def self.logWarn(msg, depth=0)
+        _methodname = getMethodname(depth)
```
NarineM: getMethodname (depth) calls are duplicated, can you remove that?

source/code/plugins/in_npmd.rb
```diff
@@ -0,0 +1,676 @@
...
+        Cmd_log           = "ErrorLog"
+
+        Npmd_conn_confirm = "NPMDAgent Connected!"
+
+        Loc_unix_endpoint = "/var/opt/microsoft/omsagent/run/npmdagent.sock"
+        Loc_config_file   = "/etc/opt/microsoft/omsagent/conf/npmdAgentConfig.conf"
+        Loc_npmd_binary   = "/opt/microsoft/omsagent/Daemon"
+
+        Stop_signal       = "SIGKILL"
+        Tag               = "npmd.oms.blob"
```
NarineM: Why this tag is hardcoded within the plugin?

source/code/plugins/in_npmd.rb
```diff
@@ -0,0 +1,676 @@
...
+
+        Retry_delay_long  = 900 # 15 minutes
```
NarineM: please append min or sec keyword to the parameter names so that it is clear which one is which

installer/conf/omsagent.d/npmd.conf
```diff
@@ -0,0 +1,14 @@
+<source>
+    type npmd
```
robbiezhang: please expose the tag in the conf, and set it to oms.npmd

installer/conf/omsagent.d/npmd.conf
robbiezhang: you don't need a special output plugin in the configuration. it can be processed by the one in omsagent.conf

source/code/plugins/in_npmd.rb
robbiezhang: Please use ALL CAPITAL for constants

source/code/plugins/in_npmd.rb
```diff
@@ -0,0 +1,676 @@
...
+                _h[KeyNetworks] = getNetworkHashFromJson(_config.elements[NetworkInfoTag].text())
+                _h[KeySubnets]  = getSubnetHashFromJson(_config.elements[SubnetInfoTag].text())
+                _h[KeyAgents]   = getAgentHashFromJson(_config.elements[AgentInfoTag].text())
+                _h[KeyRules]    = getRuleHashFromJson(_config.elements[RuleInfoTag].text())
+                _h = nil if (_h[KeyNetworks].nil? or _h[KeySubnets].nil? or _h[KeyAgents].nil? or _h[KeyRules].nil?)
+                return _h
+
+            rescue REXML::ParseException => e
+                Logger::logError "Got XML parse exception at #{e.line()}, #{e.position()}", Logger::resc
+                raise Exception.new("Got XML parse exception at #{e.line()}, #{e.position()}")
```
robbiezhang: throw exception here will make fluentd stop working

source/code/plugins/in_npmd.rb
```diff
@@ -0,0 +1,676 @@
...
+
+    # This class holds the methods for parsing
+    # a config sent via DSC into a hash
+    class UIConfigParser
+        public
+
+        # Only accessible method
+        def self.parse(string)
+            begin
+                _doc = REXML::Document.new(string)
```
robbiezhang: we suspect the REXML class has some memory issue. can you check your memory usage?

source/code/plugins/in_npmd.rb
```diff
@@ -0,0 +1,676 @@
...
+        Cmd_start         = "StartNPMD"
+        Cmd_stop          = "StopNPMD"
+        Cmd_config        = "ConfigNPMD"
+        Cmd_restart       = "RestartNPMD"
+        Cmd_log           = "ErrorLog"
+
+        Npmd_conn_confirm = "NPMDAgent Connected!"
+
+        Loc_unix_endpoint = "/var/opt/microsoft/omsagent/run/npmdagent.sock"
+        Loc_config_file   = "/etc/opt/microsoft/omsagent/conf/npmdAgentConfig.conf"
```
robbiezhang: don't use ".conf" file extension. it's reserved for fluentd configuration

source/code/plugins/in_npmd.rb
```diff
@@ -0,0 +1,676 @@
...
+
+        Cmd_start         = "StartNPMD"
+        Cmd_stop          = "StopNPMD"
+        Cmd_config        = "ConfigNPMD"
+        Cmd_restart       = "RestartNPMD"
+        Cmd_log           = "ErrorLog"
+
+        Npmd_conn_confirm = "NPMDAgent Connected!"
+
+        Loc_unix_endpoint = "/var/opt/microsoft/omsagent/run/npmdagent.sock"
```
robbiezhang: this setting should be configurable for multi-homing

source/code/plugins/in_npmd.rb
robbiezhang: it should also be configurable

source/code/plugins/in_npmd.rb
```diff
@@ -0,0 +1,676 @@
...
+            _record["DataItems"] = dataitems
+            router.emit(Tag, Engine.now, _record)
+        end
+
+        def wait_for_agent_id
+            if @agentId.nil?
+                begin
+                    @agentId = OMS::Configuration.agent_id()
+                    sleep(0.1)
+                end while @agentId.nil?
```
robbiezhang: why?

source/code/plugins/in_npmd.rb
chparimi: @gauravga

source/code/plugins/in_npmd.rb
robbiezhang: Firstly, the agent id is not necessary in the data uploaded from the agent. It could be enriched in the ODS after authentication.
Secondly, if you really want to insert agent id into the data, you need to initialize the OMS::Configuration in this plugin?

source/code/plugins/in_npmd.rb
robbiezhang: I don't see a reason why DSC module cannot copy multiple files.

source/code/plugins/in_npmd.rb
```diff
@@ -0,0 +1,676 @@
...
+        end
+
+        def server_run
+            unless defined?@server_obj and !@server_obj.nil?
+                Logger::logInfo "Server obj was not created properly, Exiting"
+            else
+                @fqdn = OMS::Common.get_fully_qualified_domain_name()
+                @agentId = OMS::Configuration.agent_id()
+                # Wait for agentId to be set to valid value
+                wait_for_agent_id()
```
robbiezhang: The plugin should never depend on any other plugin. otherwise, it will lose the flexibility and hard to test and diagnose.

dineshkarora: NPMD agent needs a GUID to handle the aggregation of data. The original idea was to re-use the one generated from OMS::Configuration, but based on the dependency issue highlighted, it is easy for us to maintain our own GUID. We will make necessary change and avoid plugin dependency.

source/code/plugins/in_npmd.rb
dineshkarora: Currently , multi-homing is not planned to be supported.

source/code/plugins/in_npmd.rb
dineshkarora: Agent GUID is required for NPMD specific purpose. As captured before, we will now generate our own GUID to keep it independent of OMS::Configuration

source/code/plugins/in_npmd.rb
chparimi: Removing dependency on both plugins. Now we create our own guid and we have our own FQDN function.

installer/conf/omsagent.d/npmd.conf
chparimi: Can we have the separate match to output plugin as it offers better control?

chparimi: Heartbeat lib is also having similarly

chparimi:
Uploaded modified commit. Now we have
1. Separate files for configuration parsing and npmd c++ binary control
2. Unit tests for both separately

source/code/plugins/in_npmd.rb
NarineM: If depth does not change, you can have getMethodname (depth) as glbal param . Function call will be done only once, then throughout the code just use the param value instead of calling the same function multiple times

source/code/plugins/in_npmd_server.rb
```diff
@@ -0,0 +1,428 @@
...
+            if File.exist?@location_uuid_file
+                f = File.new(@location_uuid_file, "r")
+                @agentId = f.read
+                f.close
+
+                create_new_guid = false if (!@agentId.nil? and !@agentId.empty?)
+            end
+
+            if create_new_guid
+                @agentId = SecureRandom.uuid
```
robbiezhang: why don't you introduce another uuid instead of using the agent id?

source/code/plugins/in_npmd_server.rb
```diff
@@ -0,0 +1,428 @@
...
+                        # Transform the UI XML configuration to agent configuration
+                        _agentConfig = NPMDConfig::GetAgentConfigFromUIConfig(_uiXml)
+                        if _agentConfig.nil? or _agentConfig == ""
+                            Logger::logWarn "Agent configuration transformation returned empty"
+                            return
+                        end
+
+                        @npmdClientSock.puts _agentConfig
+                        @npmdClientSock.flush
+                        @num_config_sent += 1 unless @num_config_sent.nil?
```
robbiezhang: you need to initialize the num_config_sent. if it was nil, it would never be assigned a value.

source/code/plugins/in_npmd_server.rb
```diff
@@ -0,0 +1,428 @@
...
+            rescue StandardError => e
+                log_error "Error when sending config: #{e}", Logger::resc
+            end
+        end
+
+        def server_run
+            unless defined?@server_obj and !@server_obj.nil?
+                Logger::logInfo "Server obj was not created properly, Exiting"
+            else
+                # Wait for agentId to be set to valid value
```
robbiezhang: is the comment still valid?

source/code/plugins/in_npmd_server.rb
```diff
@@ -0,0 +1,428 @@
...
+
+                        @npmdClientSock.puts _agentConfig
+                        @npmdClientSock.flush
+                        @num_config_sent += 1 unless @num_config_sent.nil?
+
+                    else
+                        Logger::logWarn "NPMD client socket not connected yet!"
+                    end
+                end
+            rescue Exception => e
```
robbiezhang: don't catch Exception, especially before the other error or exceptions. it will shortcut the others.

source/code/plugins/npmd_config_lib.rb
```diff
@@ -0,0 +1,335 @@
...
+                _h[KeyNetworks] = getNetworkHashFromJson(_config.elements[NetworkInfoTag].text())
+                _h[KeySubnets]  = getSubnetHashFromJson(_config.elements[SubnetInfoTag].text())
+                _h[KeyAgents]   = getAgentHashFromJson(_config.elements[AgentInfoTag].text())
+                _h[KeyRules]    = getRuleHashFromJson(_config.elements[RuleInfoTag].text())
+                _h = nil if (_h[KeyNetworks].nil? or _h[KeySubnets].nil? or _h[KeyAgents].nil? or _h[KeyRules].nil?)
+                return _h
+
+            rescue REXML::ParseException => e
+                Logger::logError "Got XML parse exception at #{e.line()}, #{e.position()}", Logger::resc
+                raise Exception.new("Got XML parse exception at #{e.line()}, #{e.position()}")
```
robbiezhang: don't use Exception.new, you can just raise <string> to raise an RuntimeError.

source/code/plugins/npmd_config_lib.rb
```diff
@@ -0,0 +1,335 @@
...
+        end
+
+    end
+
+    # Only function needed to be called from this module
+    def self.GetAgentConfigFromUIConfig(uiXml)
+        begin
+            _uiHash = UIConfigParser.parse(uiXml)
+            _agentXml = AgentConfigCreator.createXmlFromUIConfigHash(_uiHash)
+        rescue Exception => e
```
robbiezhang: don't catch exception. in this case, you can let the caller to handle.

installer/conf/omsagent.d/npmd.conf
NarineM: Your latest commit still does not have the change for this plugin and the conf file be deployed during the omsagent install. You need to add them to base.data.

chparimi: The intention is not to deploy these during omsagent install. For this solution to actually start, one would need to enable it in the web portal explicitly (NPMD Solution) and then via DSC we would be copying these over to the respective directories and restarting omsagent.

installer/conf/omsagent.d/npmd.conf
```diff
@@ -0,0 +1,19 @@
...
+    tag oms.npmd
+    location_unix_endpoint  "/var/opt/microsoft/omsagent/run/npmdagent.sock"
+    location_control_data   "/etc/opt/microsoft/omsagent/conf/npmd_agent_config.xml"
+    location_agent_binary   "/opt/microsoft/omsagent/npmd_agent"
+    location_uuid_file      "/etc/opt/microsoft/omsagent/conf/npmd_agent_guid.txt"
+</source>
+
+<match npmd.oms.blob>
```
chparimi: Yes we agree that it can be handled by default output plugin. We are basically asking if we can have this so as to have more control over the event transmission. Similar to how "oms_heartbeat" and "monitor_agent" have redefined this in "monitor.conf"

source/code/plugins/in_npmd_server.rb
chparimi: We are introducing a new UUID rather than the omsagent id. We plan to have one UUID right now for the NPMD solution over multiple restarts unless there is a purge.

source/code/plugins/in_npmd_server.rb
chparimi: This is just needed for the unit testing scenario. Hence it is fine to not initialize here.

source/code/plugins/in_npmd_server.rb
chparimi: Understood. I would remove the StandardError altogether below.

source/code/plugins/in_npmd_server.rb
chparimi: Sorry it needs to removed. Will remove it.

source/code/plugins/npmd_config_lib.rb
chparimi: Any specific reason for having RuntimeError?

source/code/plugins/npmd_config_lib.rb
```diff
@@ -0,0 +1,335 @@
...
+                _h[KeySubnets]  = getSubnetHashFromJson(_config.elements[SubnetInfoTag].text())
+                _h[KeyAgents]   = getAgentHashFromJson(_config.elements[AgentInfoTag].text())
+                _h[KeyRules]    = getRuleHashFromJson(_config.elements[RuleInfoTag].text())
+                _h = nil if (_h[KeyNetworks].nil? or _h[KeySubnets].nil? or _h[KeyAgents].nil? or _h[KeyRules].nil?)
+                return _h
+
+            rescue REXML::ParseException => e
+                Logger::logError "Got XML parse exception at #{e.line()}, #{e.position()}", Logger::resc
+                raise Exception.new("Got XML parse exception at #{e.line()}, #{e.position()}")
+            rescue StandardError => e
```
chparimi: Remove this rescue as well.

source/code/plugins/npmd_config_lib.rb
chparimi: OK

installer/conf/omsagent.d/npmd.conf
NarineM: I see. Sounds good.

NarineM: I see. Sounds good.

PR #283, opened 2016-11-18
Add OMS-Auditd-Plugin (auoms)

taglines:
See also corresponding pull-requests for Build-OMS-Agent-for-Linux and PowerShell-DSC-for-Linux as well as code committed to master branch of OMS-Auditd-Plugin (LCA approved, but not yet public)

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -9,6 +9,7 @@ OMI_ROOT := $(shell cd ../../omi/Unix; pwd -P)
 PAL_DIR := $(shell cd ../../pal; pwd -P)
 SCX_DIR := $(shell cd ../../opsmgr; pwd -P)
 DSC_DIR := $(shell cd ../../dsc; pwd -P)
+AUOMS_DIR := $(shell cd ../../auoms; pwd -P)
 
 PF_POSIX := 1
 include $(BASE_DIR)/build/config.mak
@@ -21,6 +22,7 @@ endif
 
 SCX_INTERMEDIATE_DIR := $(SCX_DIR)/intermediate/$(BUILD_CONFIGURATION)
 DSC_TARGET_DIR := $(DSC_DIR)/release
+AUOMS_TARGET_DIR := $(AUOMS_DIR)/target/$(BUILD_CONFIGURATION)
 
 RUBY_DIR := $(BASE_DIR)/source/ext/ruby
 # This Ruby version number refers only to the major/minor version (not teeny)
@@ -115,7 +117,7 @@ SYSTEST_CONFIG := $(BASE_DIR)/test/config/systest.conf
 .PHONY: all clean clean-ruby distclean clean-status kit
 .PHONY: tests test omstest unittest systemtest systemtestrb systemtestsh
 
-all : $(SCX_INTERMEDIATE_DIR) $(DSC_TARGET_DIR) $(RUBY_TESTING_DIR) $(RUBY_DEST_DIR) $(IN_PLUGINS_LIB) kit
+all : $(SCX_INTERMEDIATE_DIR) $(DSC_TARGET_DIR) $(RUBY_TESTING_DIR) $(RUBY_DEST_DIR) $(IN_PLUGINS_LIB) $(AUOMS_TARGET_DIR)/auoms-bundle-test.sh kit
         # After a successful make, undo the Ruby patches for sequential builds from old branches
 	@echo "Cleaning patched Ruby files..."
 	cd $(BASE_DIR)/source/ext/ruby; git checkout -- ext/openssl/ossl_ssl.c ext/openssl/extconf.rb
@@ -157,7 +159,13 @@ clean-scx : clean
 	@$(ECHO) "!!!!! PLEASE RE-RUN THE ./configure SCRIPT !!!!!"
 	@$(ECHO) "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!"
 
-distclean : clean clean-ruby clean-dsc clean-scx
+clean-auoms : clean
+	-make -C $(AUOMS_DIR)/build distclean
+	@$(ECHO) "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!"
+	@$(ECHO) "!!!!! PLEASE RE-RUN THE ./configure SCRIPT !!!!!"
+	@$(ECHO) "!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!"
+
+distclean : clean clean-ruby clean-dsc clean-scx clean-auoms
 	-$(RMDIR) $(OMI_ROOT)/output*
 
 	cd $(BASE_DIR)/source/ext/fluentd; git checkout -- lib/fluent/env.rb
@@ -194,6 +202,12 @@ ifeq ($(ULINUX),1)
 	$(MAKE) -C $(DSC_DIR) dsckit100
 	cd $(OMI_ROOT); $(RMDIR) output
 endif
+
+#--------------------------------------------------------------------------------
+# Build AUOMS project
+
+$(AUOMS_TARGET_DIR)/auoms-bundle-test.sh : 
+	$(MAKE) -C $(AUOMS_DIR)/build all
 
 #--------------------------------------------------------------------------------
 # Build the version of Ruby that we test with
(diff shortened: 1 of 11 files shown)


robbiezhang:
@Microsoft/omsagent-devs

build/Makefile
```diff
@@ -196,6 +204,12 @@ ifeq ($(ULINUX),1)
 endif
 
 #--------------------------------------------------------------------------------
+# Build AUOMS project
+
+$(AUOMS_TARGET_DIR)/auoms-bundle-test.sh :
```
NarineM: What is the meaning of test in the auoms bundle installer name?

installer/bundle/bundle_skel.sh
```diff
@@ -817,6 +834,21 @@ case "$installMode" in
                     [ $TEMP_STATUS -ne 0 ] && BUNDLE_EXIT_STATUS="$TEMP_STATUS"
                 fi
             done
+            for i in bundles/*-bundle-test.sh; do
+                # If filespec didn't expand, break out of loop
+                [ ! -f $i ] && break
+
+                # It's possible we have a test file without bundle; if so, ignore it
+                BUNDLE=`basename $i -bundle-test.sh`
```
NarineM: What is the  test t file used for? And where the test bundle sh is being built?

installer/bundle/bundle_skel.sh
```diff
@@ -817,6 +834,21 @@ case "$installMode" in
...
+                [ ! -f bundles/${BUNDLE}-*universal.*.sh ] && continue
+
+                ./$i
+                if [ $? -eq 0 ]; then
+                    ./bundles/${BUNDLE}-*universal.*.sh --install $FORCE $restartDependencies
```
NarineM: I do not seem to find $restartDependencies  defined anywhere, did you miss it during merge?

installer/bundle/bundle_skel.sh
```diff
@@ -885,6 +917,21 @@ case "$installMode" in
...
+            # It's possible we have a test file without bundle; if so, ignore it
+            BUNDLE=`basename $i -bundle-test.sh`
+            [ ! -f bundles/${BUNDLE}-*universal.*.sh ] && continue
+
+            ./$i
+            if [ $? -eq 0 ]; then
+                ./bundles/${BUNDLE}-*universal.*.sh --upgrade $FORCE $restartDependencies
+                TEMP_STATUS=$?
+                [ $TEMP_STATUS -ne 0 ] && BUNDLE_EXIT_STATUS="$TEMP_STATUS"
+            fi
```
NarineM: do you need to echo something when it falls in else block?

installer/conf/sudoers
```diff
@@ -31,6 +31,9 @@ omsagent ALL=(ALL) NOPASSWD: /opt/microsoft/omsagent/ruby/bin/ruby /opt/microsof
 #run hdinsightmanifestreader.rb
 omsagent ALL=(ALL) NOPASSWD: /opt/microsoft/omsagent/ruby/bin/ruby /opt/microsoft/omsagent/bin/hdinsightmanifestreader.rb
 
+# Enable nxOMSAuditdResource
+omsagent ALL=(ALL) NOPASSWD: /opt/microsoft/omsconfig/Scripts/OMSAuditdPlugin.sh
```
NarineM: per oms agent naming convention the bundle names start with lower case (it is actually all lower case , example omsagent-1.3.0-25.universal.deb.x64.sh ).

build/configure
```diff
@@ -5,6 +5,7 @@ base_dir=`(cd ..; pwd -P)`
 scxomi_dir=`(cd ../../omi/Unix; pwd -P)`
 scxpal_dir=`(cd ../../pal; pwd -P)`
 scxcim_dir=`(cd ../../opsmgr; pwd -P)`
+auoms_dir=`(cd ../../auoms; pwd -P)`
```
NarineM: you mentioned that there are two packages that need to sit under /data folder . I do not see them being installed through configure or it is done somewhere else?

source/code/plugins/auditd_plugin_lib.rb
```diff
@@ -0,0 +1,57 @@
...
+
+            for ridx in 1..record["record-count"]
+                rname = "record-data-"+(ridx-1).to_s
+                if !record.has_key?(rname)
+                    @log.error "Transformation of Auditd Plugin input failed; Missing field '" + rname + "'"
+                    return nil
+                end
+                rdata = Yajl::Parser.parse(record[rname])
+                rdata["Timestamp"] = OMS::Common.format_time(record["AuditID"].to_f)
+                rdata["AuditID"] = record["AuditID"]
```
NarineM: You might want to check for nill value here. I remember there was a limitation on oms service side on processing data with null entities

source/code/plugins/auditd_plugin_lib.rb
```diff
@@ -0,0 +1,57 @@
+require 'yajl'
```
NarineM: Please add unittests. I could not find any. you can look into other ruby plugin unittests that use mock objects for testing.

build/Makefile
taglines: auoms-bundle-test.sh is not the bundle. It is the script that checks to see if the bundle can be installed. It serves the same purpose as the *-oss-test.sh scripts.

installer/bundle/bundle_skel.sh
taglines: The value of restartDependencies is set at line 482.

installer/bundle/bundle_skel.sh
taglines: The test script is used at line 845.

installer/conf/sudoers
taglines: OMSAuditdPlugin.sh is NOT a bundle! It is the script used by the nxOMSAUditdPlugin DSC resource to complete the enable/disable "SET" operation.

build/configure
taglines: See lines 45-65 of auoms/build/configure.

installer/bundle/bundle_skel.sh
taglines: No. The "BUNDLE_EXIT_STATUS" value is handled later. And, the bundle itself will report failure to the console.

source/code/plugins/auditd_plugin_lib.rb
```diff
@@ -0,0 +1,57 @@
...
+            end
+
+            wrapper = {
+                "DataType"=>"LINUX_AUDITD_BLOB",
+                "IPName"=>"Security",
+                "DataItems"=>records
+            }
+
+            @log.info "Audit Event Processed"
+
```
sugr4: Was this for debugging? do you want this message to be logged everytime an audit event is processed?

source/code/plugins/auditd_plugin_lib.rb
sugr4: second on unit tests

source/code/plugins/filter_auditd_plugin.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    def shutdown
+        super
+        # This method is called when Fluentd is shutting down.
+        # Use it to free up resources, etc.
+    end
+
+    def filter(tag, time, record)       
+        # Create Security Baseline and Security Baseline Summary blobs based on omsbaseline tool scan & assessment results
```
sugr4: This comment does not seem to fit.

build/Makefile
NarineM: I see,  having a comment will clear it out since it has the word "bundle" in the name.

installer/bundle/bundle_skel.sh
NarineM: yes, but the bundle is for handling the error for all the components altogether.

installer/conf/sudoers
NarineM: That is fine, I meant the naming convention used in agent sh and bundle names in general.

NarineM: I see that other scripts for dsc resources also use upper case convention so it is fine if the purpose is to relate the script with the resources names.

NarineM:
Looks good except the missing tests.

build/Makefile
NarineM: I see, I interpreted it as a bundle for installing auoms for testing or mocking. I think having comments would really help. are other oss test scripts have the word "bundle" in the name?

installer/bundle/bundle_skel.sh
NarineM: I still do not understand the meaning of this comment" It's possible we have a test file without bundle; if so, ignore it " , can you rephrase it if the test file is about the bundle-test.sh itself and not some other test input data?

NarineM: ok, I thought test file here means some other input file that is used by the script which is going to look for it :) Nevermind.

installer/bundle/bundle_skel.sh
NarineM: This is fine, seems like other kits like oss handle it same way.

source/code/plugins/auditd_plugin_lib.rb
taglines: Added unit test

source/code/plugins/filter_auditd_plugin.rb
taglines: The dangers of copy-n-paste. Will remove.

PR #287, opened 2016-11-24
fix in_npmd_server_plugintest.rb test cases

chparimi:
1. Send relative fake endpoint path to prevent endpoint path too long error
2. Ignore Errno:ESRCH as before killing of fake instance it might have exited
3. Give time between binary recreation as so that mtime comparision remains
   valid.
4. Kill all stale instances during shutdown as well if any.

### DIFF
--- a/source/code/plugins/in_npmd_server.rb
+++ b/source/code/plugins/in_npmd_server.rb
@@ -64,7 +64,7 @@ module Fluent
         RETRY_START_WAIT_TIME_SECS      = 5   # 5 seconds
         RETRY_START_ATTEMPTS_PER_BATCH  = 5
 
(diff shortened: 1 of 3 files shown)


chparimi:
@Microsoft/omsagent-devs  for reviewing this PR

chparimi:
Verified this on Vanilla VM and also on pbuild.
The 32 bit one is failing with "OMSAgent already installed error"
The 64 bit one is passing.

NarineM:
The error means that the agent got installed under /opt/Microsoft path . first try uninstalling it with --purge option, then go back to your working dir and kick off a new make.

NarineM:
Please checkin this fix. It is blocking agent builds right now. thanks.

chparimi:
pbuild succeeded for 32 bit as well after fixing the OMSAgent already installed error.

PR #289, opened 2016-12-01
adding shell script for npmd solution 

varastog:
adding shell script for npmd solution to sudoers list to let it run without password


vrdmr:
Signoff.

Thanks,
Varad

PR #295, opened 2016-12-07
fix subnetmask representation in config output for IPV6

chparimi:
Subnet mask was completely expanded for IPV6 cases. Agent is expecting compressed format.


chparimi:
@Microsoft/omsagent-devs  for review

chparimi:
pbuild succeeded

source/code/plugins/npmd_config_lib.rb
```diff
@@ -57,9 +57,6 @@ class ::IPAddr
         def getNetMaskString
             _to_string(@mask_addr)
         end
-        def getIPString
```
robbiezhang: update the comment above

chparimi: Done.

PR #298, opened 2016-12-08
Include hdinsight filter in agent installer

duoxu:
I forgot to modify this file in the previous checkin. The new filter does not exist after agent is installed.


duoxu:
@robbiezhang Please take a look. If everything looks good, could you help merge it?

Thanks,
Duo

PR #299, opened 2016-12-12
fix connection from DSC to socket and add recovery for agent

chparimi:
DSC connection is not done by root user but via omsagent user
Add watchdog to make sure agent is restarted if hung

Say if agent hangs and does not send data to plugin for over
an hour then restart the agent.


chparimi:
@Microsoft/omsagent-devs

chparimi:
pbuild has passed

PR #301, opened 2016-12-13
Add Remove, RemoveAll, and List Functions in omsadmin.sh

robbiezhang:
Add the following parameters for omsadmin:
-l: list all workspaces
-x <workspace id>: remove the specified workspace
-X: remove all workspaces

Update the system tests for multi-homing onboarding,
so that the configuration from different workspaces won't corrupt.
Add system tests for onboard 2 workspaces, and verified
the multi-onboarding, and list/remove/removeall functions

Fixed a bug in disable service
Make the proxy.conf a global file for all workspaces
@Microsoft/omsagent-devs @lagalbra

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -29,7 +29,7 @@ CONF_OMSADMIN=$CONF_DIR/omsadmin.conf
 CONF_OMSAGENT=$CONF_DIR/omsagent.conf
 
 # Omsagent proxy configuration
-CONF_PROXY=$CONF_DIR/proxy.conf
+CONF_PROXY=$ETC_DIR/proxy.conf
 
 # File with OS information for telemetry
 OS_INFO=/etc/opt/microsoft/scx/conf/scx-release
@@ -77,6 +77,15 @@ usage()
     echo "Maintenance tool for OMS:"
     echo "Onboarding:"
     echo "$basename -w <workspace id> -s <shared key> [-d <top level domain>]"
+    echo
+    echo "List Workspaces:"
+    echo "$basename -l"
+    echo
+    echo "Remove Workspace:"
+    echo "$basename -x <workspace id>"
+    echo
+    echo "Remove All Workspaces:"
+    echo "$basename -X"
     echo
     echo "Define proxy settings ('-u' will prompt for password):"
     echo "$basename [-u user] -p host[:port]"
@@ -181,7 +190,7 @@ parse_args()
 {
     local OPTIND opt
 
-    while getopts "h?s:w:d:vp:u:a:c" opt; do
+    while getopts "h?s:w:d:vp:u:a:clx:X" opt; do
         case "$opt" in
         h|\?)
             usage
@@ -214,6 +223,16 @@ parse_args()
         c) 
             COLLECTD=1
             echo "Setting up collectd for OMS..."
+            ;;
+        l)
+            LIST_WORKSPACES=1
+            ;;
+        x)
+            REMOVE=1
+            WORKSPACE_ID=$OPTARG
+            ;;
+        X)
+            REMOVE_ALL=1
             ;;
         esac
     done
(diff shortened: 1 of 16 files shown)


lagalbra:
If I understand the proxy changes correctly, all workspaces connected to a single machine will use the same proxy (if proxy is provided), correct? Is this the intended design?

installer/scripts/omsadmin.sh
```diff
@@ -301,11 +320,7 @@ onboard()
         clean_exit 1
     fi
 
-
-    # If a test is not in progress then create workspace-specific directories
-    if [ -z "$TEST_WORKSPACE_ID" -a -z "$TEST_SHARED_KEY" ]; then
-        create_workspace_directories $WORKSPACE_ID 
-    fi
+    create_workspace_directories $WORKSPACE_ID
```
lagalbra: Did you run the unit and system tests with this change? Some were failing before I added the surrounding if. (If you intend to bulk-groom the tests later, this is fine.

installer/scripts/omsadmin.sh
```diff
@@ -412,13 +427,12 @@ onboard()
...
     save_config
 
-    # If a test is not in progress then register omsagent as a service and start the agent 
-    # If a test is in progress then omsagent conf/symlinks do not need to be updated
-    if [ -z "$TEST_WORKSPACE_ID" -a -z "$TEST_SHARED_KEY" ]; then
-        copy_omsagent_conf
+    copy_omsagent_conf
 
-        update_symlinks
+    update_symlinks
```
lagalbra: Same testing concern here

source/code/plugins/out_oms_blob.rb
```diff
@@ -29,7 +29,7 @@ def initialize
     config_param :key_path, :string, :default => '/etc/opt/microsoft/omsagent/certs/oms.key'
     config_param :blob_uri_expiry, :string, :default => '00:10:00'
     config_param :url_suffix_template, :string, :default => "custom_data_type + '/00000000-0000-0000-0000-000000000002/' + OMS::Common.get_hostname + '/' + OMS::Configuration.agent_id + '/' + suffix + '.log'"
-    config_param :proxy_conf_path, :string, :default => '/etc/opt/microsoft/omsagent/conf/proxy.conf'
+    config_param :proxy_conf_path, :string, :default => '/etc/opt/microsoft/omsagent/proxy.conf'
```
lagalbra: This change also needs to be added to source/code/plugins/in_heartbeat_request.rb and installer/conf/omsagent.d/heartbeat.conf.
Currently, heartbeat.conf has proxy_path listed as %CONF_DIR_WS%/proxy.conf ; for minimal changes, this should simply be removed and change in_heartbeat_request.rb to have the same config param as the out_oms_* files now do.

test/installer/scripts/omsadmin_systest.rb
```diff
@@ -94,15 +94,42 @@ def test_reonboard
...
+
+    output_removeall = `#{@omsadmin_script} -X`
+    assert_equal(0, $?.to_i, "Remove all workspace should succeed: #{output_removeall}")
 
-    # With current multihoming support, the GUID is re-used when onboarding to an additional workspace
-    # TODO Robbie change this when AGENT_GUID logic has been fully converted with multihoming
-    assert_equal(old_guid, new_guid, "The GUID should not change when reonboarding to another workspace ID")
-    assert_match(/Reusing/, output, "Should be reusing GUID")
+    output_list = `#{@omsadmin_script} -l`
+    assert_match(/No Workspace/, output_list, "Did not find workspace 2 in list")
```
lagalbra: Change message in case of non-match: "Output list should contain no workspaces: #{output_list}"

installer/scripts/omsadmin.sh
robbiezhang: Yes, the unittest and systemtest all passed with the new folder structure for multi-homing.

robbiezhang:
I chatted with Kris a while back. We agreed that the proxy should be an agent-wise setting. We think the customers regard the omsagent as a single agent, no matter how many workspaces it is attached to. So, it might be weird that they use different proxy to communicate to different workspace. 
BTW, DSC agent will also follow this guideline.

test/installer/scripts/omsadmin_systest.rb
robbiezhang: good catch

PR #304, opened 2016-12-16
npmd: limit emit chunk size to 5MB of data

chparimi:
If emitted chunk size is large causing size of data to be more
than buffer_chunk_limit, warnings mentioning possible errors to
output plugins is given by the frameworks.

Hence limiting the emit size to 5000 dataitems which would amount
to 5MB of data.


chparimi:
@Microsoft/omsagent-devs

varastog:
pbuild succeeded

PR #305, opened 2016-12-16
Updating OMS heartbeat to collect workspace-specific process usage

lagalbra:
Query OMI's omicli executable with PID instead of process name to support multihoming
Update existing system tests and unit tests
Add unit tests for helper telemetry detector

This change is intended to be merged into the FB-multihoming feature branch.
@Microsoft/omsagent-devs @robbiezhang


test/code/plugins/agent_topology_request_script_plugintest.rb
```diff
@@ -229,15 +210,21 @@ def test_valid_data_empty_process_stats
...
+    IO.write(PID_FILE, CONSTANT_PID)
 
-    atr.get_telemetry_data(OS_INFO, CONF_OMSADMIN)
+    atr.get_telemetry_data(OS_INFO, CONF_OMSADMIN, PID_FILE)
+    verify_complete_xml(atr, expect_telemetry = false)
+ 
+    flexmock(AgentTopologyRequest).flexmock_teardown
+  end    
 
+  def verify_complete_xml(atr, expect_telemetry = true)
```
lagalbra: Git's diff has parsed this a little strangely; just know that the new file's lines 223-271 are a single new helper method called verify_complete_xml

source/code/plugins/in_heartbeat_request.rb
```diff
@@ -30,14 +31,14 @@ def configure (conf)
       if !@key_path
         raise Fluent::ConfigError, "'key_path' option is required on heartbeat_request input"
       end
-      if !@proxy_path
```
robbiezhang: good catch

PR #306, opened 2016-12-17
Manage the Port for Syslog and Monitor Agent

robbiezhang:
Move the syslog fluentd configuration out of the
omsagent.conf into omsagent.d/syslog.conf, and make it
a template for different workspace and port.
Make monitor.conf a template too.

In onboarding phase, find a free port to the syslog and
monitor agent, and customize the port in the conf.
Port for syslog starts from 25224
Port for monitor agent starts from 25324

Support configure rsyslog and syslog-ng per workspace
Move the syslog configure into configure_syslog.sh
from the linux.data. The configuration will now happen
in the onboarding phase instead of installation phase.

Configure the rsyslog or syslog-ng according to the
port assigned to the workspace in 3 different ways:
/etc/rsyslog.d/95-omsagent.conf
/etc/rsyslog.conf
/etc/syslog-ng/syslog-ng.conf

Remove the syslog configure settings when removing
the workspace.

@Microsoft/omsagent-devs @lagalbra

### DIFF
--- a/installer/conf/omsagent.d/syslog.conf
+++ b/installer/conf/omsagent.d/syslog.conf
@@ -1,0 +1,12 @@
+<source>
+  type syslog
+  port %SYSLOG_PORT%
+  bind 127.0.0.1
+  protocol_type udp
+  tag oms.syslog
+</source>
+
+<filter oms.syslog.**>
+  type filter_syslog
+</filter>
+
--- a/installer/conf/rsyslog.conf
+++ b/installer/conf/rsyslog.conf
@@ -1,18 +1,18 @@
-# OMS Syslog collection
-kern.warning              @127.0.0.1:25224
-user.warning              @127.0.0.1:25224
-daemon.warning            @127.0.0.1:25224
-auth.warning              @127.0.0.1:25224
-syslog.warning            @127.0.0.1:25224
-uucp.warning              @127.0.0.1:25224
-authpriv.warning          @127.0.0.1:25224
-ftp.warning               @127.0.0.1:25224
-cron.warning              @127.0.0.1:25224
-local0.warning            @127.0.0.1:25224
-local1.warning            @127.0.0.1:25224
-local2.warning            @127.0.0.1:25224
-local3.warning            @127.0.0.1:25224
-local4.warning            @127.0.0.1:25224
-local5.warning            @127.0.0.1:25224
-local6.warning            @127.0.0.1:25224
-local7.warning            @127.0.0.1:25224
+# OMS Syslog collection for workspace %WORKSPACE_ID%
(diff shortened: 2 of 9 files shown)


installer/scripts/omsadmin.sh
```diff
@@ -70,6 +70,12 @@ OMSCLOUD_ID=""
 UUID=""
 USER_ID=`id -u`
 
+DEFAULT_SYSLOG_PORT=25224
+DEFAULT_MONITOR_AGENT_PORT=25324
```
lagalbra: Is there any reason that we would ever try to claim more than 100 syslog ports, thus overflowing into the monitor agent ports? In this case, should the find_available_port() method simply skip past the monitor agent ports?

installer/conf/omsagent.d/syslog.conf
```diff
@@ -0,0 +1,12 @@
+<source>
+  type syslog
+  port %PORT%
```
lagalbra: Since this port and the port replaced in monitor.conf are two different types of ports (syslog vs monitor agent) could they be templated with different strings?

installer/conf/rsyslog.conf
```diff
@@ -1,18 +1,18 @@
...
+ftp.warning               @127.0.0.1:%PORT%
+cron.warning              @127.0.0.1:%PORT%
+local0.warning            @127.0.0.1:%PORT%
+local1.warning            @127.0.0.1:%PORT%
+local2.warning            @127.0.0.1:%PORT%
+local3.warning            @127.0.0.1:%PORT%
+local4.warning            @127.0.0.1:%PORT%
+local5.warning            @127.0.0.1:%PORT%
+local6.warning            @127.0.0.1:%PORT%
+local7.warning            @127.0.0.1:%PORT%
```
lagalbra: Same template string concern as above.

installer/scripts/omsadmin.sh
robbiezhang: I don't think we will use more than 100 syslog ports. However, even it overflows into the monitor agent port range, it's still fine because it will assign the port only when the port is not occupied.

installer/conf/syslog-ng.conf
```diff
@@ -1,4 +1,3 @@
-#OMS_facility = all
-filter f_warning_oms { level(warning); };
-destination warning_oms { udp("127.0.0.1" port(25224)); };
-log { source(src); filter(f_warning_oms); destination(warning_oms); };
+filter f_%WORKSPACE_ID%_oms { level(warning); };
+destination d_%WORKSPACE_ID%_oms { udp("127.0.0.1" port(%PORT%)); };
```
sugr4: %SYSLOG_PORT%

PR #307, opened 2016-12-21
NPMD: Config parsing error fix and backoff for agent retry

chparimi:
When default rule is not used in UI then the rule expects to pick the subnetId from the subnet table.
Backoff will make sure we do not retry immediately to start agent in case of exit.


varastog:
pbuild succeeded

chparimi:
@Microsoft/omsagent-devs

chparimi:
Squashed the changes and ran pbuild again which succeeded. Merging.

PR #309, opened 2016-12-27
npm onboard script

varastog:
1. This script needs to be run to onbaord to NPM solution after oms agent is installed on the machine.
2. This is being used for NPM private preview until new version of OMS agent is released. 
3. This is being pushed to OMS agent because we need it to be downloaded from a public location like 'installer/scripts/onboard_agent.sh'

### DIFF
--- a/installer/scripts/npm_startup.sh
+++ b/installer/scripts/npm_startup.sh
@@ -1,0 +1,74 @@
+# This scripts onboards NPMD solution to omsagent node
+#Usage:
+# For Ubuntu and CentOs use 'sudo sh npm_startup.sh'
+# For RHEL use 'sh npm_startup.sh'
+
+cmd="# NPMD\nomsagent ALL=(ALL) NOPASSWD: /opt/microsoft/omsconfig/Scripts/NPMAgentBinaryCap.sh"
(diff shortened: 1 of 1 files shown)


varastog:
Adding @Microsoft/omsagent-devs for review

installer/scripts/npm_startup.sh
```diff
@@ -0,0 +1,63 @@
+cmd="omsagent ALL=(ALL) NOPASSWD: /opt/microsoft/omsconfig/Scripts/NPMAgentBinaryCap.sh"
```
robbiezhang: please add comment or usage on how to use the script

installer/scripts/npm_startup.sh
```diff
@@ -0,0 +1,63 @@
...
+
+    if [ $INCLUDEDIR -eq 1 -a -d /etc/sudoers.d ]; then
+        return 0
+    else
+        return 1
+    fi
+}
+
+SudoSupportsIncludeDirective
+if [ $? -eq 0 ]; then
```
robbiezhang: In the RemoveSudoersSupport in uninstall script, we use the comment "# Begin sudo configuration for omsagent" and "# End sudo configuration for omsagent" to identity the section for OMS agent. Since your new line is out of the scope, it won't be removed when uninstall the omsagent package.

PR #317, opened 2017-01-04
sudoers changes to allow update management commands to be executed

shuliums:
Add changes in sudoers to allow update management commands to be executed. @Microsoft/omsagent-devs @robbiezhang @Atchub

### DIFF
--- a/installer/conf/sudoers
+++ b/installer/conf/sudoers
@@ -1,5 +1,6 @@
 # Begin sudo configuration for omsagent
 Defaults:omsagent !requiretty
+Defaults:omsagent lecture = never 
 
 #update and restart syslog
 omsagent ALL=(ALL) NOPASSWD: /opt/microsoft/omsconfig/Scripts/OMSRsyslog.post.sh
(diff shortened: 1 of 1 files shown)


installer/conf/sudoers
```diff
@@ -29,4 +29,8 @@ omsagent ALL=(ALL) NOPASSWD: /opt/microsoft/omsagent/ruby/bin/ruby /opt/microsof
 
 #run hdinsightmanifestreader.rb
 omsagent ALL=(ALL) NOPASSWD: /opt/microsoft/omsagent/ruby/bin/ruby /opt/microsoft/omsagent/bin/hdinsightmanifestreader.rb
+
+#run update management commands
+omsagent ALL=(ALL) NOPASSWD: /usr/bin/apt-get *
```
robbiezhang: It exposes too many permissions (install, update, upgrade, etc) to the omsagent. 
Can you give a list of commands that will be used by PatchManagement?

BTW, we need to update the release document about this change.
@agup006 @KrisBash

vrdmr: @shuliums @robbiezhang 
If we are not doing any remove/purge, and only doing install, lets only have the require params explicitly -
```
/usr/bin/apt-get install *
/usr/bin/apt-get update *
/usr/bin/apt-get upgrade *
/usr/bin/apt-get dist-upgrade *
```

ShujunLiu:
@vrdmr @robbiezhang , please review this change. Thanks!

PR #319, opened 2017-01-05
installer: scripts: npm: Update to show helper message

chparimi:
Updating the script to post helper message in case firewall was
not configured.

### DIFF
--- a/installer/scripts/npm_startup.sh
+++ b/installer/scripts/npm_startup.sh
@@ -45,6 +45,7 @@ tcp_port=8084
 release_file=/etc/os-release
 str_centOS7="CentOS Linux 7"
 str_rhel7="Red Hat Enterprise Linux Server 7.0"
+require_notify=1
 if [ -f $release_file ];
 then
         pretty_name=`cat $release_file | grep PRETTY_NAME=`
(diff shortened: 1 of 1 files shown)


chparimi:
@Microsoft/omsagent-devs

installer/scripts/npm_startup.sh
```diff
@@ -72,3 +74,7 @@ else
         echo "File $release_file not found!"
 fi
 
+if [ $require_notify -eq 1 ];
+then
+    echo "Please configure your firewall (if running), to allow connections for destination port 8084 over TCP for NPM solution to run"
```
lagalbra: Will this port number always be 8084?

chparimi: Yes it shall be always 8084. We have not yet had support for altering the port in the solution automatically. That sort of a request will require manual efforts which are not yet supported fully.

lagalbra: Ok, I see that it doesn't change in the current implementation. Could you just clean it up by printing $tcp_port instead of 8084?

chparimi: OK doing that

chparimi:
Done updating the helper line to print $tcp_port rather than 8084.

installer/scripts/npm_startup.sh
chparimi: Done

PR #323, opened 2017-01-06
npmd: fix test case 13 and 14 for when omsagent user is absent

chparimi:
When omsagent user is absent the test cases are throwing an
ArgumentError which needs to be handled.


chparimi:
@Microsoft/omsagent-devs

chparimi:
Pbuild succeeded

PR #325, opened 2017-01-07
Remove fluentd conf for OMSConsistencyInvoker upon upgrade

sugr4:
@Microsoft/omsagent-devs

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -29,6 +29,7 @@ DPKG_CONF_QUALS="--force-confold --force-confdef"
 OMISERV_CONF="/etc/opt/omi/conf/omiserver.conf"
 OLD_OMISERV_CONF="/etc/opt/microsoft/scx/conf/omiserver.conf"
 OMS_RUBY_DIR="/opt/microsoft/omsagent/ruby/bin"
+OMS_CONSISTENCY_INVOKER="/etc/cron.d/OMSConsistencyInvoker"
 
 # These symbols will get replaced during the bundle creation process.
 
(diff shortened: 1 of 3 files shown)


installer/bundle/bundle_skel.sh
```diff
@@ -842,6 +843,14 @@ case "$installMode" in
...
 		
         if [ $KIT_STATUS -eq 0 ]; then
+            # Remove fluentd conf for OMSConsistencyInvoker upon upgrade, if it exists
+            rm -f /etc/opt/microsoft/omsagent/conf/omsagent.d/omsconfig.consistencyinvoker.conf
+
+            # In case --upgrade is run without -w <id> and -s <key>
+            if [ ! -f /etc/cron.d/OMSConsistencyInvoker ]; then
+                echo "*/5 * * * * omsagent /opt/omi/bin/OMSConsistencyInvoker >/dev/null 2>&1" > /etc/cron.d/OMSConsistencyInvoker
+            fi
+            /opt/microsoft/omsagent/bin/service_control restart
```
robbiezhang: the omsagent is stopped before upgrading, you don't need to restart it.

sugr4: Removed line for omsagent restart

installer/bundle/bundle_skel.sh
```diff
@@ -625,6 +625,7 @@ cd $EXTRACT_DIR
 # Do we need to remove the package?
 set +e
 if [ "$installMode" = "R" -o "$installMode" = "P" ]; then
+    rm -f /etc/cron.d/OMSConsistencyInvoker
```
NarineM: I see that you use /etc/cron.d many places, can this be converted to a const string? Also is the cron path same for all the distros?

NarineM: Can you also add a description to this commit for what issue this fix solves ? Would this change fix the verbose message with files in use warning during uninstall too?

sugr4: Yes it's the same for all distros that we support

sugr4: This line only deletes the cron file. It doesn't fix the verbose messages you mentioned. Since this is the first step in --purge and --remove and we still get the  messages, this cron file did not contribute to those messages in the first place.

sugr4:
Made further changes to accommodate system tests

PR #327, opened 2017-01-10
npmd: fix case when username running process is too long

chparimi:
When the username associated with running process is longer than
about 8 chars, then in ps aux the uid is displayed rather than
the username.

Fixing the corner case.


chparimi:
@Microsoft/omsagent-devs

chparimi:
@lagalbra I verified this with a very long username so that the uid appears in the ps aux. It worked. Going to run pbuild

chparimi:
Pbuild succeeded

PR #328, opened 2017-01-10
Fix several bugs for old OS

robbiezhang:
Fix the issue that sed -z is not supported in some distro
Fix the issue about insserv error: service omsagent already provided
@Microsoft/omsagent-devs

### DIFF
--- a/installer/scripts/service_control
+++ b/installer/scripts/service_control
@@ -271,7 +271,7 @@ enable_omsagent_service()
             local omsagent_initd=/etc/init.d/$OMSAGENT_WS
             cp /etc/opt/microsoft/omsagent/sysconf/omsagent.ulinux $omsagent_initd
 
-            sed -i s,WORKSPACE_ID=.*,WORKSPACE_ID=$WORKSPACE_ID,1 $omsagent_initd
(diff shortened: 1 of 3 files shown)


installer/scripts/service_control
```diff
@@ -271,7 +271,7 @@ enable_omsagent_service()
             local omsagent_initd=/etc/init.d/$OMSAGENT_WS
             cp /etc/opt/microsoft/omsagent/sysconf/omsagent.ulinux $omsagent_initd
 
-            sed -i s,WORKSPACE_ID=.*,WORKSPACE_ID=$WORKSPACE_ID,1 $omsagent_initd
+            sed -i s,%WORKSPACE_ID%,$WORKSPACE_ID,g $omsagent_initd
```
lagalbra: I don't think it's a bug, but I'm curious about the different uses of sed throughout the multihoming branch; most use 1, but this uses g.

robbiezhang: 1 means replace only the first occurrence; g means replace globally.
The old one assume there is one occurrence matches the pattern. Since I change it to use symbol, and add a new place, it should become a global match.

PR #330, opened 2017-01-11
Update Ruby version to 2.3.3

lagalbra:
These changes all support the upgrade of OMSAgent Ruby to version 2.3:
 - License remains the same, but it is now marked with the new Ruby version
 - Built Ruby directory structure renamed with correct version
 - SSL certs patch removed since Ruby 2.3 contains RubyGems.org certs update
 - OpenSSL patch added for incompatibility between OpenSSL 0.9.8 and Ruby 2.3
 - CFLAG for Ruby compilation added to enable building FluentD
 - Re-generated ruby.data and Ruby's configure script
 - Change to source/ext/ruby: point to the ruby v2_3_3 tag

@Microsoft/omsagent-devs


lagalbra:
@jeffaco @agup006 Could you please review this and let me know if anything looks glaringly wrong?

build/Makefile
```diff
@@ -52,7 +50,7 @@ DEBUG_FLAGS := -g
 endif
 
 # Need to use RUBY_COMPILE_FLAGS when compiling code that uses C Ruby interfaces
-# (Note that "-Wshadow -Wredundant-decls" don't work with Ruby v2.2.0)
+# (Note that "-Wshadow -Wredundant-decls" don't work with Ruby v2.2.0 or v2.3.3)
```
jeffaco: The fact that it doesn't work with Ruby 2.2.0 is not relevant if we're not building that. I'd just change it to say `... don't work with Ruby v2.3.3`.

build/Makefile
```diff
@@ -23,9 +23,9 @@ SCX_INTERMEDIATE_DIR := $(SCX_DIR)/intermediate/$(BUILD_CONFIGURATION)
 DSC_TARGET_DIR := $(DSC_DIR)/release
 
 RUBY_DIR := $(BASE_DIR)/source/ext/ruby
+RUBY_VER := 2.3.0
```
jeffaco: I don't know if you need comments here about base system requirements (we don't have that for other base system requirements here). But they definitely do need to be documented in the `Build-OMS-Agent-for-Linux` project.

build/Makefile
```diff
@@ -512,12 +497,14 @@ endif
 #--------------------------------------------------------------------------------
 # Run unit tests
 #
+# Note: with the upgrade to ruby 2.3, rake needs the temporary PATH change to
```
jeffaco: We do this routinely all over the place, and don't generally comment on how shell works. I think, when we're launching a shell command, it's reasonable to assume that the developer understands shell, or has the resources to look up what something does. If you want to comment, go ahead, but we do this sort of thing all over the place (thus my suggestion), and have never comment "Shell behaves like xxx yyy zzz" at those times.

build/Makefile
lagalbra: Added a comment about what this version number refers to, because it can be confusing, since this line says 2.3.0 when our tagged Ruby version has a different teeny number.

build/Makefile
lagalbra: Clarified my comment about PATH.

build/Makefile
lagalbra: Done.

build/Makefile
lagalbra: Created a pull request in the Build-OMS-Agent-for-Linux project with the documentation update: [PR 19](https://github.com/Microsoft/Build-OMS-Agent-for-Linux/pull/19)
Note: autoconf does not need to be updated to successfully build omsagent.

NarineM: @jeffaco ruby 2.3 has a new compiler which does not work with older BASERUBY versions such as 1.8. This was not a case in the past.

build/Makefile
NarineM: @jeffaco I still think this case is special to highlight because it is changing system environment variables. devs can easily confuse the $PATH local parameter used in make vs system configuration. And this workaround is needed for only 2.3 (was not a case in 2.2)

PR #332, opened 2017-01-13
Travis-CI setup

sugr4:
@Microsoft/omsagent-devs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,44 @@
+language: c
+dist: trusty   # Travis build machine is Ubuntu Trusty 14.04 x64
+sudo: required # A sudo enabled, full VM is used for building instead of a docker container
+
+git:
+  submodules: false
+
+before_install:
+  - sudo apt-get update -qq
+  # Pre-reqs for building omsagent
+  - sudo apt-get install -qq pkg-config make ruby bison g++ rpm librpm-dev libpam0g-dev libssl-dev libmysqlclient-dev
+  - rvm use system                 # Use system ruby (v1.9.3p484) instead of ruby provided by rvm
+  - sed -i 's/--no-ri//' ~/.gemrc  # ri data from installed gems is required for our build
+
+# Travis-CI automatically clones the OMS-Agent-for-Linux repo since this is where this .travis.yml file is located
+# We will ignore this clone and clone the superproject Build-OMS-Agent-for-Linux as follows
+
+install:
+  - cd ../..
+  - git clone https://github.com/Microsoft/Build-OMS-Agent-for-Linux.git Microsoft/Build-OMS-Agent-for-Linux
+  - cd Microsoft/Build-OMS-Agent-for-Linux
+  # change url from ssh to https for submodules, travis-ci cannot download a submodule that isn’t on a public url
+  - sed -i 's/git@github.com:/https:\/\/github.com\//' .gitmodules 
+  - git submodule update --init --recursive dsc omi opsmgr opsmgr-kits pal
+  - git clone https://github.com/Microsoft/OMS-Agent-for-Linux.git omsagent 
+  - cd omsagent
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,44 @@
...
+sudo: required # A sudo enabled, full VM is used for building instead of a docker container
+
+git:
+  submodules: false
+
+before_install:
+  - sudo apt-get update -qq
+  # Pre-reqs for building omsagent
+  - sudo apt-get install -qq pkg-config make ruby bison g++ rpm librpm-dev libpam0g-dev libssl-dev libmysqlclient-dev
+  - rvm use system                 # Use system ruby instead of ruby provided by rvm
```
lagalbra: As per requirements with the upgrade to Ruby 2.3, this version must be at least 1.9. I believe an Ubuntu 14 build machine should pass with this, but I've asked Suju to double-check.

.travis.yml
```diff
@@ -0,0 +1,44 @@
...
+  - sed -i 's/--no-ri//' ~/.gemrc  # ri data from installed gems is required for our build
+
+# Travis-CI automatically clones the OMS-Agent-for-Linux repo since this is where this .travis.yml file is located
+# We will ignore this clone and clone the superproject Build-OMS-Agent-for-Linux as follows
+
+install:
+  - cd ../..
+  - git clone https://github.com/sugr4/Build-OMS-Agent-for-Linux.git sugr4/Build-OMS-Agent-for-Linux
```
lagalbra: Why isn't it https://github.com/Microsoft/Build-OMS-Agent-for-Linux.git ? (i.e. why is your GitHub username in it)

.travis.yml
```diff
@@ -0,0 +1,44 @@
...
+  - cd sugr4/Build-OMS-Agent-for-Linux
```
lagalbra: As part of best practice, maybe this should just be in a folder called bld-omsagent

.travis.yml
```diff
@@ -0,0 +1,44 @@
...
+  # PR validation during check-in will be setup in github in the settings section of our repo
+  - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then git fetch origin refs/pull/$TRAVIS_PULL_REQUEST/head ; git checkout FETCH_HEAD; fi 
+  - cd build
+  # pwd: /home/travis/build/sugr4/Build-OMS-Agent-for-Linux/omsagent/build
+  # Since 'build' is a keyword for travis-CI, change Makefile to substitute 'omsagent/build' instead of 'build'
+  - sed -i 's/(subst \/build,,$(CURDIR))/(subst omsagent\/build,omsagent,$(CURDIR))/' Makefile 
+
+script:
+  - sudo ./configure --enable-debug
+  - sudo make && sudo make unittest # System tests are not run because OMS-Agent-for-Linux-testconfig.git is a private repo
```
lagalbra: Should we have to run these commands with "sudo"? I thought we had resolved that issue.

lagalbra: Has this config been used to make a test run yet, or will that option only be available once this file is checked in?

.travis.yml
robbiezhang: do you need to change sugr4 to master?

NarineM: The only comment I have is that since audit plugin is now public you can update the git URL to be master branch.

.travis.yml
sugr4: ruby 1.9.3p484 (2013-11-22 revision 43786) [x86_64-linux]

.travis.yml
sugr4: Removed sudo and tested a build

sugr4: Just made a test run on forked repos, but will be able to build using the upstream repos only after being checked in

.travis.yml
sugr4: Done

PR #336, opened 2017-01-15
Enforce key checks on JSON data being sent to NPM solution

chparimi:
Enforcing content contracts for NPM solution.

### DIFF
--- a/source/code/plugins/in_npmd_server.rb
+++ b/source/code/plugins/in_npmd_server.rb
@@ -213,20 +213,23 @@ module Fluent
                         else
                             _uploadData = _json["DataItems"].reject {|x| x["SubType"] == "ErrorLog"}
                             _errorLogs  = _json["DataItems"].select {|x| x["SubType"] == "ErrorLog"}
+                            _validUploadDataItems = Array.new
                             _uploadData.each do |item|
                                 if item.key?("SubType")
                                     # Append FQDN to path data
                                     if !@fqdn.nil? and item["SubType"] == "NetworkPath"
                                         @num_path_data += 1 unless @num_path_data.nil?
                                         item["Computer"] = @fqdn
-                                        # Append agent Guid to agent data
+                                        _validUploadDataItems << item if is_valid_dataitem(item)
+                                    # Append agent Guid to agent data
                                     elsif !@agentId.nil? and item["SubType"] == "NetworkAgent"
                                         @num_agent_data += 1 unless @num_agent_data.nil?
                                         item["AgentId"] = @agentId
+                                        _validUploadDataItems << item if is_valid_dataitem(item)
                                     end
                                 end
                             end
-                            emit_upload_data_dataitems(_uploadData) if !_uploadData.nil? and !_uploadData.empty?
+                            emit_upload_data_dataitems(_validUploadDataItems) if !_validUploadDataItems.nil? and !_validUploadDataItems.empty?
                             emit_error_log_dataitems(_errorLogs) if !_errorLogs.nil? and !_errorLogs.empty?
                         end
                     end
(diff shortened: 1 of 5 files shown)


chparimi:
@Microsoft/omsagent-devs

source/code/plugins/in_npmd_server.rb
```diff
@@ -182,15 +188,49 @@ def kill_all_agent_instances
...
+                    _line = @npmdDiagSock.gets
+                    if _line.nil? and !is_npmd_seen_in_ps()
+                        @npmdDiagSock = nil
+                        Logger::logInfo "Exiting diag thread as npmdAgent found stopped"
+                        break
+                    end
+                    next if _line.nil? or _line.strip== ""
+                    _json = check_and_get_json(_line.chomp)
+                    unless !_json.nil?
+                        Logger::logWarn "Sent string to diag is not a json string", Logger::loop
```
NarineM: Please try to not abbreviate the words , sometimes it makes the logs hard to read. Same comment applies to line#191.

source/code/plugins/in_npmd_server.rb
```diff
@@ -207,27 +209,32 @@ def npmd_reader
                     _json = check_and_get_json(_line.chomp)
                     unless !_json.nil?
                         Logger::logWarn "Sent string to plugin is not a json string", Logger::loop
+                        log_error "String received not json: #{_line[0..100]}" if _line.bytesize > 50
```
varastog: bytesize>100 ?

chparimi: we need to be wary of the 17 bytes case. Anything above that is OK. So went with 50.

source/code/plugins/npmd_config_lib.rb
```diff
@@ -343,3 +343,71 @@ def self.GetAgentConfigFromUIConfig(uiXml)
...
+        elsif itemType == DATAITEM_DIAG
+            _contract = CONTRACT_DIAG_DATA_KEYS
+        end
+
+        return DATAITEM_ERR_INVALID_TYPE, nil if _contract.empty?
+
+        item.keys.each do |k|
+            return DATAITEM_ERR_INVALID_FIELDS, k if !_contract.include?(k)
+        end
+        _contract.each do |e|
```
varastog: nitpick: if count of contract and item is equal then you can skip the second loop

source/code/plugins/in_npmd_server.rb
```diff
@@ -242,19 +249,47 @@ def npmd_reader
...
+            return false if _itemType.empty?
+
+            _res, _prob = NPMContract::IsValidDataitem(item, _itemType)
+
+            return true if _res == NPMContract::DATAITEM_VALID
+            if (_res == NPMContract::DATAITEM_ERR_INVALID_FIELDS)
+                Logger::logInfo "Invalid key in #{item["SubType"]} data: #{_prob}"
+            elsif (_res == NPMContract::DATAITEM_ERR_MISSING_FIELDS)
+                Logger::logInfo "Key #{_prob} absent in #{item["SubType"]} data"
+            end
```
varastog: nitpick: although you are checking for valid itemtype [lines 253-261], it would be a good practice to add a log in case _res is DATAITEM_ERR_INVALID_TYPE because it is a possible return value for NPMContract::IsValidDataitem

source/code/plugins/npmd_config_lib.rb
chparimi: Thanks will add that

source/code/plugins/in_npmd_server.rb
chparimi: ok

PR #340, opened 2017-01-19
Merge the Array in the input in out_oms_api

robbiezhang:
Check the input of the out_oms_api plugin.
If it's a Hash, add it into the buffer.
If it's an Array, merge it with the items in the buffer.
Otherwise, ignore.

@Microsoft/omsagent-devs

### DIFF
--- a/source/code/plugins/out_oms_api.rb
+++ b/source/code/plugins/out_oms_api.rb
@@ -100,7 +100,7 @@ module Fluent
     #   tag: string. the tag of the item
     #   records: hash[]. an arrary of data
     def handle_records(tag, records)
-      @log.trace "Handling record : #{tag}"
+      @log.trace "Handling record : #{tag} x #{records.count}"
       tags = tag.split('.')
       if tags.size >= 3
         # tag should have 3 or 4 parts at least:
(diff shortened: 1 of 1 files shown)


source/code/plugins/out_oms_api.rb
```diff
@@ -164,10 +164,19 @@ def write(chunk)
...
+        if !record.to_s.empty?
+          if !datatypes.has_key?(tag)
+            datatypes[tag] = []
+          end
+
+          if record.is_a?(Array)
+            record.each do |r|
+              datatypes[tag] << r if !r.to_s.empty? and r.is_a?(Hash)
+            end
+          elsif record.is_a?(Hash)
```
lagalbra: Do we know that record is always (expected to be) a Hash or an Array?

robbiezhang: Yes, the service side expects a JObject or a JArray of JObject, but it cannot handle a JArray of JArray. Since we batch process the records, we need to make sure it always send a JArray of JObject. No simple type is allowed.

robbiezhang:
The functionality is verified on an agent. I'm waiting for the pbuild.

PR #343, opened 2017-01-20
Add cluster type field in hdinsight filter

duoxu:
Read cluster type in hdinsight cluster manifest and append to every log/metrics sent to OMS.

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -639,6 +639,7 @@ cd $EXTRACT_DIR
 set +e
 if [ "$installMode" = "R" -o "$installMode" = "P" ]; then
     rm -f "$OMS_CONSISTENCY_INVOKER" > /dev/null 2> /dev/null 
+    rm -f "$ONBOARD_FILE" > /dev/null 2> /dev/null
     if [ -f /opt/microsoft/omsagent/bin/uninstall ]; then
         /opt/microsoft/omsagent/bin/uninstall $installMode
     else


duoxu:
@Microsoft/omsagent-devs
@agup006 

Please take a look.

duoxu:
Sorry, I will fix the test.

test/code/plugins/hdinsightfilter_plugintest.rb
```diff
@@ -20,10 +20,11 @@ def create_driver(conf=CONFIG)
...
   def test_filter
     d = create_driver
-    d.instance.command = "echo 'fake_cluster_name'"
+    d.instance.command = "echo '{\"cluster_name\":\"fake_cluster_name\", \"cluster_type\":\"fake_cluster_type\"}'"
     d.run
     record = Hash.new
     emit = d.instance.filter("", "", record)
-    assert_equal("fake_cluster_name\n", emit["ClusterName"])
+    assert_equal("fake_cluster_name", emit["ClusterName"])
+	assert_equal("fake_cluster_type", emit["ClusterType"])
```
lagalbra: nit: use consistent indentation on this line

source/code/plugins/filter_hdinsight.rb
```diff
@@ -11,24 +11,27 @@ class HdinsightFilter < Filter
...
     SCRIPT = BASE_DIR + '/bin/hdinsightmanifestreader.rb '
-	
-	attr_accessor :command
+
+    attr_accessor :command
 
     def configure(conf)
       super
-      @hostname = OMS::Common.get_hostname or "Unknown host"
+      @hostname = OMS::Common.get_hostname or "unknown"
```
lagalbra: Any reason that this string changed? "unknown" is a little less clear than "Unknown host"

test/code/plugins/hdinsightfilter_plugintest.rb
duoxu: Thanks! I will fix this.

source/code/plugins/filter_hdinsight.rb
duoxu: In cluster_type, the default I set to "unknown" if this information cannot be retrieved from our manifest, maybe I should set it to "unknown cluster type", making it consistent with hostname instead.

PR #344, opened 2017-01-23
npmd: add TimeGenerated key to path and agent dataitems

chparimi:
Adding the TimeGenerated key value pair for path and agent
dataitems.


chparimi:
@Microsoft/omsagent-devs

PR #345, opened 2017-01-24
Add cluster type field in hdinsight filter

duoxu:
Read cluster type in hdinsight cluster manifest and append to every
log/metrics sent to OMS.


duoxu:
@lagalbra 

Sorry, I closed previous PR. I have squashed all the commits into one and opened up a new PR here.

PR #350, opened 2017-01-25
Add -O parameter to onboarding helper script for explicit output

lagalbra:
This unblocked one customer, and it does not impact other scenarios.
While the omsagent bundle will now be downloaded under a less explicit name, the version can still be discovered by looking at installinfo.txt after the bundle has been installed.
If desired, I can add a file rename command before installing the bundle.
@Microsoft/omsagent-devs


lagalbra:
Re-implemented so that file will get saved as bundle name.

PR #351, opened 2017-01-25
npmd: change startup script to verify for firewalld state

chparimi:
Now startup script does not verify for CentOS7 or RHEL7 release
file but rather for presence of firewalld directly. When found
it attempts to open the destination port. Incase it does not
open a port or does not find firewalld then it prints a message
requesting user to configure their own firewall for allowing
connections to destination port for NPM.


chparimi:
@Microsoft/omsagent-devs

PR #352, opened 2017-01-25
npmd: config parsing partial rejection logic and reporting

chparimi:
1. Say if all IPs of an agent are invalid, then reject agent itself
2. Say if all subnet IDs of a network are invalid, then reject network
3. Say if any subnet IDs of given pair in rule is invalid then skip
   that subnet pair in network test matrix
4. If no subnet pairs are left in network test matrix then skip
   that rule altogether
5. Adding a means to report the configuration drop errors
6. Adding a unit test to verify the configuration drop errors


source/code/plugins/npmd_config_lib.rb
```diff
@@ -64,6 +64,56 @@ def getNetMaskString
...
+            _ruleDrops=""
+
+            if @@agent_ip_drops != 0
+                _agentIpDrops = "Agent IPs=#{@@agent_ip_drops}"
+            end
+            if @@agent_drops != 0
+                _agentDrops= "Agents=#{@@agent_drops}"
+            end
+            if @@network_subnet_drops != 0
+                _networkSNDrops = "Network subnets=#{@@network_subnet_drops}"
```
NarineM: I would suggest to have error strings defined as consts in the beginning instead of having them hardcoded throughout the code in case the they need to be changed or localized.

chparimi: Updated this with error strings.

PR #353, opened 2017-01-25
Ensure patched Ruby files are cleaned up after successful make

lagalbra:
This should unblock Jenkins builds that run on an older release branch, as well as other scenarios that have not occurred yet.

@Microsoft/omsagent-devs

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -116,6 +116,9 @@ SYSTEST_CONFIG := $(BASE_DIR)/test/config/systest.conf
 .PHONY: tests test omstest unittest systemtest systemtestrb systemtestsh
 
 all : $(SCX_INTERMEDIATE_DIR) $(DSC_TARGET_DIR) $(RUBY_TESTING_DIR) $(RUBY_DEST_DIR) $(IN_PLUGINS_LIB) kit
+        # After a successful make, undo the Ruby patches for sequential builds from old branches
+	@echo "Cleaning patched Ruby files..."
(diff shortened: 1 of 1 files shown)


build/Makefile
```diff
@@ -116,6 +116,9 @@ SYSTEST_CONFIG := $(BASE_DIR)/test/config/systest.conf
 .PHONY: tests test omstest unittest systemtest systemtestrb systemtestsh
 
 all : $(SCX_INTERMEDIATE_DIR) $(DSC_TARGET_DIR) $(RUBY_TESTING_DIR) $(RUBY_DEST_DIR) $(IN_PLUGINS_LIB) kit
+        # After a successful make, undo the Ruby patches for sequential builds from old branches
+	@echo "Cleaning patched Ruby files..."
+	cd $(BASE_DIR)/source/ext/ruby; git checkout -- ext/openssl/ossl_ssl.c ext/openssl/extconf.rb
```
NarineM: This one is to clean up the workaround for ssl issue , correct? if this step is specific to a ruby version, it would be good to have a comment about it, it might no longer be needed if base os is moved to cent6 or ruby is upgraded.

NarineM: please put the version info here as well.

lagalbra: This change is specific to the patches that were added for the Ruby version and the SSLv3 issues. Even with future Ruby versions, the patches are still necessary as long as we support OpenSSL 0.9.8 and the Ubuntu 16 as install scenarios.

NarineM: This patching was already implemented in ruby's tplunk branch, so it might not be needed if we upgrade to a newer version of ruby. can you update the description of this PR , or comment section to mention the version that patching was implemented? The description is very broad.

PR #354, opened 2017-01-26
Ensure that proxy.conf location is backwards-compatible

lagalbra:
For backwards-compatibility with agents upgraded to a bundle with the new proxy.conf location

Behavior will be to check the new location, then the old location, then assume no proxy is configured.
Plugins and the omsadmin script have been updated.

@Microsoft/omsagent-devs @KrisBash

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -30,6 +30,7 @@ CONF_OMSAGENT=$CONF_DIR/omsagent.conf
 
 # Omsagent proxy configuration
 CONF_PROXY=$ETC_DIR/proxy.conf
+OLD_CONF_PROXY=$DF_CONF_DIR/proxy.conf
 
 # File with OS information for telemetry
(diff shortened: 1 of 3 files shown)


installer/scripts/omsadmin.sh
```diff
@@ -30,6 +30,7 @@ CONF_OMSAGENT=$CONF_DIR/omsagent.conf
 
 # Omsagent proxy configuration
 CONF_PROXY=$ETC_DIR/proxy.conf
+OLD_CONF_PROXY=/etc/opt/microsoft/omsagent/conf/proxy.conf
```
sugr4: use $ETC_DIR/conf/proxy.conf

lagalbra: Changed to use $DF_CONF_DIR/proxy.conf instead since that is the convention [here](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/scripts/omsadmin.sh#L652)

PR #360, opened 2017-01-30
npmd: start watchdog timer after first configuration is sent to agent

chparimi:
The agent halts for first configuration without sending any data. If configuration does not arrive within an hour then the waiting agent will be killed by the watchdog thread.

To prevent this scenario, we start the watchdog timer after the first configuration is sent to agent.

### DIFF
--- a/source/code/plugins/in_npmd_server.rb
+++ b/source/code/plugins/in_npmd_server.rb
@@ -95,7 +95,7 @@ module Fluent
             @agent_restart_count = 0
             @last_npmd_start = nil
             start_npmd()
-            @watch_dog_thread = Thread.new(&method(:watch_dog_wait_for_pet))
+            @watch_dog_thread = nil
             @watch_dog_sync = Mutex.new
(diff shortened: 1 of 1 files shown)


chparimi:
@Microsoft/omsagent-devs

source/code/plugins/in_npmd_server.rb
```diff
@@ -595,6 +598,12 @@ def server_run
 
         def watch_dog_wait_for_pet
             _sleepInterval = WATCHDOG_PET_INTERVAL_SECS
+
+            # Pet right after first agent configuration is sent
+            @watch_dog_sync.synchronize do
```
varastog: Add a small threshold on line 615 so to make sure that we go into the expected if condition as _diffTime could be = or > WATCHDOG_PET_INTERVAL_SECS

chparimi: Actually we are OK with having something close to an hour.

PR #364, opened 2017-01-31
Populate Package Name in UpdateRunProgress record

shuliums:
@stasku @Atchub @Microsoft/omsagent-devs

### DIFF
--- a/source/code/plugins/patch_management_lib.rb
+++ b/source/code/plugins/patch_management_lib.rb
@@ -293,7 +293,7 @@ class LinuxUpdates
             # "OSType": "Linux",
             # "UpdateId": "8579fbee-d418-43e7-ac67-5fcc0b11cbb7",
             # "UpdateRunName": "LinuxUpdateRun",
-            # "UpdateTitle": "Security Update for git-1.0.0.0",
+            # "PackageName": "vim-common",
             # "Status": "Succeeded",
             # "Computer": "LinuxXenial",
             # "StartTime": "2016-10-21 04:25:51Z",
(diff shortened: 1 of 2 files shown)


source/code/plugins/patch_management_lib.rb
```diff
@@ -411,10 +413,22 @@ def process_yum_record(record, tag, host, time)
...
+        if !package_name_sections.nil? && package_name_sections.length > 0
+            package_name = package_name_sections[0][/[\A\d:+]*(\S*)/, 1]
+        end
+        
 		update_run_record["Computer"] = host
 		update_run_record["OSType"] = "Linux"
 		update_run_record["UpdateRunName"] = getUpdateRunName()
-		update_run_record["UpdateTitle"] = record["package-name"]
+		update_run_record["UpdateTitle"] = ""
+        update_run_record["PackageName"] = (package_name.nil?)?record["package_name"] : package_name
```
lagalbra: nit: fix indentation here

source/code/plugins/patch_management_lib.rb
```diff
@@ -411,10 +413,22 @@ def process_yum_record(record, tag, host, time)
...
+        package_name = nil;
+        if !package_name_sections.nil? && package_name_sections.length > 0
+            package_name = package_name_sections[0][/[\A\d:+]*(\S*)/, 1]
+        end
+        
 		update_run_record["Computer"] = host
 		update_run_record["OSType"] = "Linux"
 		update_run_record["UpdateRunName"] = getUpdateRunName()
-		update_run_record["UpdateTitle"] = record["package-name"]
+		update_run_record["UpdateTitle"] = ""
```
stasku: Could you use nil instead of empty string?

shuliums: There is no String with a value of nil. Because nil is simply not a String.  nil is actually also an object in Ruby. So I will keep initialize it with empty string.

PR #367, opened 2017-02-02
Fix bug: OMS agent logs are filling the hard disk.

markwgit:
Add Sleep to mitigate the error and log filling in case of old OMS Agent.
Add BaselineType and ActualResult attributes

### DIFF
--- a/installer/conf/omsagent.d/security_baseline.conf
+++ b/installer/conf/omsagent.d/security_baseline.conf
@@ -10,7 +10,7 @@
 <source> 
     type exec
     tag oms.security_baseline
-    command /opt/microsoft/omsagent/plugin/omsbaseline -d /etc/opt/microsoft/omsagent/conf/omsagent.d/ 
+    command sleep 60 && /opt/microsoft/omsagent/plugin/omsbaseline -d /etc/opt/microsoft/omsagent/conf/omsagent.d/ 
     format json
     run_interval 24h
 </source>
--- a/source/code/plugins/security_baseline_lib.rb
+++ b/source/code/plugins/security_baseline_lib.rb
@@ -99,7 +99,8 @@ module OMS
                         "InformationalFailedRules" => informational_failed_rules,
                         "PercentageOfPassedRules" => percentage_of_passed_rules,
                         "AssessmentId" => assessment_id,
-                        "OSName" => "Linux"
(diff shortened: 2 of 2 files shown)


markwgit:
@Microsoft/omsagent-devs
@lagalbra

NarineM:
Mark , can you explain how the workaround is solving the disk space issue? What is the error that is fixes?

lagalbra:
Related to [DSC PR 212](https://github.com/Microsoft/PowerShell-DSC-for-Linux/pull/212)
The command running every 24 hours failed when omsbaseline was not present (this is by design). Because of the way FluentD works (this will be investigated further), this failure creates an error message but a command like this: "<successful command like sleep> && <failing command like omsbaseline>" will not create an error message. This is why the "sleep" command will be used in both the run-once input configuration and the run-every-24h input configuration.

NarineM:
If the whole purpose was to update the command to add something with && to avoid logging the error, it could be some dummy echo or dummy command. Sleep 60 gives an impression that the expectation is to wait and start the config after 60 sec which is very confusing . Some customers can remove the wait accidentally from the conf. Also the description should be updated to tell that the actual issue with fluentd trying after failed retrywill be addressed in  a separate PR.

lagalbra:
I did bring up the fact that he could have used "echo" on a call with Mark. I think the placeholder command should remain as "sleep 60" since it will be consistent across the file. Customers should not be changing our config like this.
The "sleep 60" command in the top <source> config was a design detail suggested by Robbie, so it did not fail in the rush of commands to run as soon as the agent is onboarded.

brandubh:
Guys I fear you still have the issue, the new conf just adds a source, bnut doesn't remove the previous one:
<source>
    type exec
    tag oms.security_baseline
    command sleep 60 && /opt/microsoft/omsagent/plugin/omsbaseline -d  /etc/opt/microsoft/omsagent/conf/omsagent.d/
    format json
</source>

<source>
    type exec
    tag oms.security_baseline
    command /opt/microsoft/omsagent/plugin/omsbaseline -d /etc/opt/microsoft/omsagent/conf/omsagent.d/
    format json
    run_interval 24h
</source>

lagalbra:
@brandubh The conf has the same number of source blocks as it did before. This change will be tested locally before it is deployed to public customers.

PR #368, opened 2017-02-03
Remove collectd setup via bundle install and omsadmin.sh

sugr4:
Modify docs for collectd manual setup

### DIFF
--- a/docs/OMS-Agent-for-Linux.md
+++ b/docs/OMS-Agent-for-Linux.md
@@ -123,7 +123,6 @@ Options:
   --version              Version of this shell bundle.
   --version-check        Check versions already installed to see if upgradable.
   --debug                use shell debug mode.
-  --collectd             Enable collectd.
   
   -w id, --id id         Use workspace ID <id> for automatic onboarding.
   -s key, --shared key   Use <key> as the shared key for automatic onboarding.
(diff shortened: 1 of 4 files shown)


docs/OMS-Agent-for-Linux.md
```diff
@@ -547,16 +546,32 @@ The OMS Agent for Linux also listens on port 26000 for CollectD metrics and then
...
+    ```
+    If your CollectD config directory is located in `/etc/collectd/collectd.conf.d/`:
+    ```
+    sudo cp /etc/opt/microsoft/omsagent/sysconf/omsagent.d/oms.conf /etc/collectd/collectd.conf.d/oms.conf
+    ```
+    **Note:** For CollectD versions before 5.5 you will have to modify the tags in `oms.conf` as shown above.
+
+2. Copy `collectd.conf` to the desired workspace's omsagent configuration directory.
+    ```
+    sudo cp /etc/opt/microsoft/omsagent/sysconf/omsagent.d/collectd.conf /etc/opt/microsoft/omsagent/<workspace id>/conf/omsagent.d/
```
lagalbra: Include the command "sudo chown omsagent:omiusers /etc/opt/microsoft/omsagent/<workspace id>/conf/omsagent.d/collectd.conf"

installer/bundle/bundle_skel.sh
```diff
@@ -553,16 +552,6 @@ do
             shift 1
             ;;
 
-        --collectd)
```
lagalbra: Maybe we could keep this condition, but if this parameter is used, just print an informational message like "CollectD can no longer be configured with the --collectd option. Please see OMSAgent-for-Linux documentation here: https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/docs/OMS-Agent-for-Linux.md"

installer/scripts/omsadmin.sh
```diff
@@ -236,10 +232,6 @@ parse_args()
         a)
             AZURE_RESOURCE_ID=$OPTARG
             ;;
-        c) 
-            COLLECTD=1
```
lagalbra: Same printout idea here: print an informational message like "CollectD can no longer be configured with the --collectd option. Please see OMSAgent-for-Linux documentation here: https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/docs/OMS-Agent-for-Linux.md"

docs/OMS-Agent-for-Linux.md
sugr4: Done

installer/bundle/bundle_skel.sh
sugr4: I don't think this is necessary as I expect this message to be a part of the release notes.

PR #378, opened 2017-02-17
Adding shutdown -r command to omsagent

Atchub:
@omsagent-devs

Update management through automation worker restarts the system after installation of the updates.
this change will allow the patch runbook to initiate a restart.  

the change will be removed once the automation worker specific user is added

### DIFF
--- a/installer/conf/sudoers
+++ b/installer/conf/sudoers
@@ -34,6 +34,9 @@ omsagent ALL=(ALL) NOPASSWD: /opt/microsoft/omsagent/ruby/bin/ruby /opt/microsof
 # Enable nxOMSAuditdResource
 omsagent ALL=(ALL) NOPASSWD: /opt/microsoft/omsconfig/Scripts/OMSAuditdPlugin.sh
 
+#Enable shutdown for update management
+omsagent ALL=(ALL) NOPASSWD: /sbin/shutdown -r *
+
 #run update management commands
 omsagent ALL=(ALL) NOPASSWD: /usr/bin/apt-get dist-upgrade *
 omsagent ALL=(ALL) NOPASSWD: /usr/bin/apt-get update


lagalbra:
@Microsoft/omsagent-devs

lagalbra:
Where's the code that calls shutdown as the omsagent user?

Atchub:
its in a Runbook which is triggered by automation worker which runs as the omsagent user

NarineM:
@Atchub , 
hi, can you give more info on why system restart needs to be done from within the agent ? It is seems weird restarting services that are not owned by omsagent user.

Atchub:
Update Management solution for Linux has a concept of Maintenance window, With in the Maintenance window, we install the updates and restart the system at the end of the update.

KrisBash:
lgtm

installer/conf/sudoers
```diff
@@ -34,6 +34,9 @@ omsagent ALL=(ALL) NOPASSWD: /opt/microsoft/omsagent/ruby/bin/ruby /opt/microsof
 # Enable nxOMSAuditdResource
 omsagent ALL=(ALL) NOPASSWD: /opt/microsoft/omsconfig/Scripts/OMSAuditdPlugin.sh
 
+#Enable shutdown for update management
+omsagent ALL=(ALL) NOPASSWD: /sbin/shutdown -r *
```
niroyb: Avoid asterisks in the sudoers file if it is not needed. `shutdown` is usually called with `now` as a parameter.

Atchub: * is used to vary it between now and delay depending on customer feedback

lagalbra: Then you should have two lines: `/sbin/shutdown -r now` and `/sbin/shutdown -r delay`

Atchub: there is no delay option explicitly, I meant now or we can add a min number as delay.

so it can be "shutdown -r 6" for 6mins from current time

lagalbra: Oh, I understand. The * is probably the simplest solution here then.

PR #382, opened 2017-02-22
Run logrotate for omsagent logs every 5 minutes

sugr4:
@Microsoft/omsagent-devs


NarineM:
Is 5 min configurable? I wonder if 5 min is too fast. Have you run any perf tests with this change?

sugr4:
@NarineM Yes it is configurable to make it run every 5 minutes. Even if it checks every 5 minutes, it rotates the log only if the size exceeds 50kb. No issue with perfs.

NarineM:
Signoff.

Issue #313, opened 2016-12-29
Installer: Detection of architecture error prone


arvedarved:
The get_arch() function currently uses uname for detecting if it is running 64bit. 

This does not detect if a 32bit userland is used with a 64bit kernel. Better solution: use `getconf LONG_BIG` which returns either 64 or 32.

sugr4:
Closing this issue as this has been resolved in the latest agent release: omsagent-1.3.1-15

PR #384, opened 2017-02-27
Improve architecture detection

sugr4:
This commit uses userspace to determine
architecture instead of checking CPU or kernel

@Microsoft/omsagent-devs 
Resolves issue https://github.com/Microsoft/OMS-Agent-for-Linux/issues/313

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -295,7 +295,7 @@ pkg_upd_list() {
 
 get_arch()
 {
-    if [ $(uname -m) = 'x86_64' ]; then
+    if [ $(getconf LONG_BIT) = 64 ]; then
         echo "x64"
     else
         echo "x86"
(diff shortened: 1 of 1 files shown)


installer/bundle/bundle_skel.sh
```diff
@@ -293,26 +293,17 @@ pkg_upd_list() {
...
 compare_arch()
 {
     #check if the user is trying to install the correct bundle (x64 vs. x86) 
     echo "Checking host architecture ..."
-    AR=$(get_arch)
+    HOST_ARCH="$(getconf LONG_BIT)" 
     
     case $OMS_PKG in
-        *"$AR") 
+        *"$HOST_ARCH")
```
lagalbra: Suju is changing this to say "x$HOST_ARCH"

installer/bundle/bundle_skel.sh
```diff
@@ -293,26 +293,17 @@ pkg_upd_list() {
...
-        echo "x86"
-    fi
-}
-
 compare_arch()
 {
     #check if the user is trying to install the correct bundle (x64 vs. x86) 
     echo "Checking host architecture ..."
-    AR=$(get_arch)
+    HOST_ARCH="$(getconf LONG_BIT)"
```
lagalbra: This returns 32 for x86 architecture; my suggestion would be to keep the get_arch() method and echo this refactored result out like before.

sugr4:
Made mentioned changes, tested on both x64 and x86 systems

installer/bundle/bundle_skel.sh
```diff
@@ -304,15 +304,15 @@ get_arch()
...
-    AR=$(get_arch)
+    HOST_ARCH=$(get_arch)
     
     case $OMS_PKG in
-        *"$AR") 
+        *"$HOST_ARCH") 
             ;;
         *)         
-            echo "Cannot install $OMS_PKG on ${AR} platform"
+            echo "Cannot install $OMS_PKG on ${HOST_ARCH} bit platform"
```
lagalbra: Since this value should still come out as "x64" or "x86", I think you can remove the "bit" from the echo.

PR #385, opened 2017-02-28
Relaxing requirements for installing SCX and OMI packages under --force

lagalbra:
Due to other Azure VM Extensions using the packages we depend on (SCX, OMI), the installation of the omsagent has been failing when these packages cannot be upgraded.
When the OMSAgent is installed via Azure extension, `--upgrade --force`, and the goal is to do as much as possible to succeed installation.
This change still checks that any dependencies fulfill the minimum requirements for the omsagent package to be installed, but does not require any check more than that if `--force` is used.

@Microsoft/omsagent-devs

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -30,6 +30,9 @@ OMISERV_CONF="/etc/opt/omi/conf/omiserver.conf"
 OLD_OMISERV_CONF="/etc/opt/microsoft/scx/conf/omiserver.conf"
 OMS_RUBY_DIR="/opt/microsoft/omsagent/ruby/bin"
 OMS_CONSISTENCY_INVOKER="/etc/cron.d/OMSConsistencyInvoker"
+# Minimum versions of dependency packages
+OMI_MIN_VERSION="1.0.8.6"
+SCX_MIN_VERSION="1.6.2.129"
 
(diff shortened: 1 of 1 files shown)


lagalbra:
Unit and system tests have passed and pBuild is running now.

I have tested that upgrading, normal functions, and purging a bundle work correctly, both with and without `--force`.

installer/bundle/bundle_skel.sh
```diff
@@ -96,64 +99,87 @@ cleanup_and_exit()
...
+    elif [ $A_BUILD -gt $B_BUILD ]; then
         return 1
     fi
 
-    # Version available is idential to installed version, so don't install
+    # VersionA is idential to VersionB
     return 1
 }
 
+check_version_depending_on_force() {
```
sugr4: Looks good to me. But can we improve this function name? maybe use the old name check_version_installable, and add a comment each before lines 177 and 179? It's up to you @lagalbra

lagalbra:
Should note: this PR is mainly for code review purposes. I am going to test this change in the Azure Linux VM Extension before merging this change to master.

installer/bundle/bundle_skel.sh
lagalbra: Will do. I had created this name before I renamed the original `check_version_installable` to `first_version_lower_than_second`

PR #396, opened 2017-03-03
npmd: remove npm_version file post installation of omsagent

chparimi:
Removing npm_version file will allow DSC to update the binary
and provide the proper capabilities for proper functioning of
NPM.


chparimi:
@Microsoft/omsagent-devs

PR #399, opened 2017-03-06
Remove multi-homing feature from documentation

lagalbra:
@keikhara @NarineM 

Removed mentions of "primary" or "secondary" workspaces
Removed directions for onboarding with multi-homing marker
Removed mentions of workspace ID parameter in daemon commands

Did not change references to workspace-specific paths so that the existence of such paths will still be referenced in our documentation.


keikhara:
sign-off. checked it thanks.

PR #401, opened 2017-03-08
Improve warning message from service_control when onboarding fails

lagalbra:
@sugr4 @Microsoft/omsagent-devs 

This warning message is shown when the bundle is installed but onboarding fails; service_control is called and gives the message that the workspace "doesn't exist", when in fact it does, just not in the context of the machine.

### DIFF
--- a/installer/scripts/service_control
+++ b/installer/scripts/service_control
@@ -35,7 +35,7 @@ setup_variables()
         if [ -f $ETC_DIR/$1/conf/omsadmin.conf ]; then
             . "$ETC_DIR/$1/conf/omsadmin.conf"
         else
-            echo "Warning: Specified workspace $1 doesn't exist"
+            echo "Warning: Specified workspace $1 does not exist on the machine. Check the correctness of the workspace ID and that onboarding succeeded."
             exit 0
         fi
     fi


NarineM:
the new message does not make any sense to me either . Also it is not an error but a warning when your PR description says "improve the error message"

lagalbra:
There are two main cases when this message will be shown:
1. The bundle is installed with the workspace ID/key parameters, but onboarding fails for some reason. The service_control script will still get called with the workspace ID, but since the workspace was not onboarded, the configuration it checks for will not exist.
2. The service_control script is called directly for a restart or some other process operation with a workspace ID that is not connected to the machine: `sudo /opt/microsoft/omsagent/bin/service_control restart <incorrect-workspace-ID>`.

In both cases, we know that the workspace ID provided has not been onboarded, but the user may not have read the onboarding logs as carefully. I've added a sentence about checking the onboarding status.

NarineM:
the thing is that it could be also because of incorrect WID was provided. I think that is what the original intent was here. So I am not sure what the proposed "workspace not configured in the machine " would mean?

lagalbra:
If the meaning is not clear, I'll change it. How does this sound:

"Warning: Specified workspace $1 doesn't exist on the machine. Check the correctness of the workspace ID and that onboarding succeeded."

PR #402, opened 2017-03-08
Remove VERIFY_NONE from SSL verification process

lagalbra:
@Microsoft/omsagent-devs 

Prevent unknown endpoints from accepting data from the agent by checking the SSL certificate in HTTP requests.

### DIFF
--- a/source/code/plugins/oms_common.rb
+++ b/source/code/plugins/oms_common.rb
@@ -672,7 +672,7 @@ module OMS
                                 proxy[:addr], proxy[:port], proxy[:user], proxy[:pass])
         end
         http.use_ssl = true
-        http.verify_mode = OpenSSL::SSL::VERIFY_NONE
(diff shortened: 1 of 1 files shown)


NarineM:
Can you explain by "none" check is not replaced with a validation flag but instead verify mode is completely removed? If the purpose of this PR is to use default verification (which I am not 100% sure), there should a comment about what https verification will be used moving forward in each of these file or even better , set the default flag.

lagalbra:
According to the [Ruby HTTP doc](https://ruby-doc.org/stdlib-2.3.3/libdoc/net/http/rdoc/Net/HTTP.html#verify_mode) page: If you set :use_ssl as true, you can use https and default value of #verify_mode is set as OpenSSL::SSL::VERIFY_PEER.

I will make this explicit and test and commit again.

PR #403, opened 2017-03-08
npmd: delete binary and version files if agent misses capability

chparimi:
If NPMD agent binary does not have raw socket capability then
do not run the agent and delete the version files as well as the
agent binary. DSC in the next poll interval will update the
binary with the proper capabilities.

### DIFF
--- a/source/code/plugins/in_npmd_server.rb
+++ b/source/code/plugins/in_npmd_server.rb
@@ -52,6 +52,8 @@ module Fluent
         attr_accessor :num_config_sent
         attr_accessor :is_purged
         attr_accessor :omsagentUID
+        attr_accessor :do_capability_check
+        attr_accessor :npmd_version_dir_prefix
 
         CMD_START         = "StartNPM"
         CMD_STOP          = "StopNPM"
@@ -79,12 +81,19 @@ module Fluent
         EXIT_RESTART_BACKOFF_TIMES_SECS = [60, 120, 300, 600, 1200, 2400]
         EXIT_RESTART_BACKOFF_THRES_SECS = 900
 
+        NPMD_AGENT_CAPABILITY = "cap_net_raw+ep"
+        NPMD_VERSION_FILE_DIR = "state"
+        NPMD_VERSION_FILE_NAME = "npm_version"
+
         def start
             @binary_presence_test_string = "npmd_agent" if @binary_presence_test_string.nil?
             @binary_invocation_cmd = @location_agent_binary if @binary_invocation_cmd.nil?
+            @npmd_version_dir_prefix = "/var/opt/microsoft/omsagent" if @npmd_version_dir_prefix.nil?
             kill_all_agent_instances()
             upload_pending_stderrors()
             check_and_update_binaries()
+            @do_capability_check = true unless !@do_capability_check
+            check_agent_capability() unless !@do_capability_check
             setup_endpoint()
             @fqdn = get_fqdn()
             check_and_get_guid()
@@ -94,7 +103,7 @@ module Fluent
             @stop_sync = Mutex.new
             @agent_restart_count = 0
             @last_npmd_start = nil
-            start_npmd()
+            start_npmd() if File.exist?(@location_agent_binary)
             @watch_dog_thread = nil
             @watch_dog_sync = Mutex.new
             @watch_dog_last_pet = Time.new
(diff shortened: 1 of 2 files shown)


chparimi:
This is to ensure that even if upgrade step is utilized by a customer (because of which binary loses the prior assigned capability) the binary will not run and eventually DSC will update the binary with proper capabilities.

chparimi:
@Microsoft/omsagent-devs

source/code/plugins/in_npmd_server.rb
```diff
@@ -334,6 +343,28 @@ def check_and_update_binaries
             File.unlink(_x64BinPath) if File.exist?(_x64BinPath)
         end
 
+        def check_agent_capability
+            return if !File.exist?(@location_agent_binary) or @skip_capability_check
```
varastog: what are we using @skip_capability_check for?

source/code/plugins/in_npmd_server.rb
```diff
@@ -334,6 +343,28 @@ def check_and_update_binaries
...
+
+            # Checking if capabilties are proper else delete binary and version file
+            _getCapResult = `getcap #{@location_agent_binary}`
+            if !_getCapResult.include?(NPMD_AGENT_CAPABILITY)
+                # Delete agent binary
+                File.unlink(@location_agent_binary)
+
+                # Delete version file(s)
+                _fileLocPostFix = "#{NPMD_VERSION_FILE_DIR}/#{NPMD_VERSION_FILE_NAME}"
```
varastog: For path manipulations, wouldn't it be better to use any Path apis?

chparimi: Removing the same.

Also adding the case when NPMD binary is absent but version files are present

source/code/plugins/in_npmd_server.rb
```diff
@@ -79,12 +81,19 @@ def configure(conf)
...
+
         def start
             @binary_presence_test_string = "npmd_agent" if @binary_presence_test_string.nil?
             @binary_invocation_cmd = @location_agent_binary if @binary_invocation_cmd.nil?
+            @npmd_version_dir_prefix = "/var/opt/microsoft/omsagent" if @npmd_version_dir_prefix.nil?
             kill_all_agent_instances()
             upload_pending_stderrors()
             check_and_update_binaries()
+            @do_capability_check = true unless @do_capability_check == false
+            check_agent_capability() unless !@do_capability_check
```
lagalbra: For consistency, I would suggest using the same syntax to check `@do_capability_check`: either `@do_capability_check == false` or `!@do_capability_check`

source/code/plugins/in_npmd_server.rb
```diff
@@ -334,6 +343,34 @@ def check_and_update_binaries
             File.unlink(_x64BinPath) if File.exist?(_x64BinPath)
         end
 
+        def check_agent_capability
+            _deleteVersionFiles = false
+
+            if !File.exist?(@location_agent_binary)
+                _deleteVersionFiles = true
+            else
+                _getCapResult = `getcap #{@location_agent_binary}`
```
lagalbra: Have you verified that the `getcap` program exists on [all distros/versions we support](https://github.com/Microsoft/OMS-Agent-for-Linux/tree/master#supported-linux-operating-systems)?

source/code/plugins/in_npmd_server.rb
```diff
@@ -334,6 +343,34 @@ def check_and_update_binaries
...
+                end
+            end
+
+            if _deleteVersionFiles
+                # Single/First workspace scenario
+                _fileLocPostFix = "#{NPMD_VERSION_FILE_DIR}/#{NPMD_VERSION_FILE_NAME}"
+                _loc = "#{@npmd_version_dir_prefix}/#{_fileLocPostFix}"
+                File.unlink(_loc) if File.exist?(_loc)
+                # Multiple workspace scenario
+                _stateDirs = Dir.glob("#{@npmd_version_dir_prefix}/**/#{NPMD_VERSION_FILE_DIR}")
```
lagalbra: Have you tested this plugin on a machine with multiple workspaces onboarded, and that all version files get deleted?

chparimi: have added that in the unit test.

source/code/plugins/in_npmd_server.rb

source/code/plugins/in_npmd_server.rb
chparimi: The DSC module is not yet ready to add version files for all workspaces. However if files are present this is deleting properly.

chparimi:
Pbuild succeeded as well.

PR #413, opened 2017-03-16
Update omsadmin.sh for LAD onboarding

hosungsmsft:
Per our discussion at our 3/15/2017 meeting. This is all we need for now.

Thanks Laura for the SCOM pointer. It was very helpful.

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -352,6 +352,21 @@ onboard_scom()
     configure_logrotate
 }
 
+onboard_lad()
+{
+    echo "Onboarding LAD"
+    create_workspace_directories "LAD"
+
+    touch $CONF_DIR/omsagent.conf
+    echo "@include omsagent.d/*" > $CONF_DIR/omsagent.conf
+    chown_omsagent $CONF_DIR/*
+    make_dir $CONF_DIR/omsagent.d
+    #Always register LAD as secondary Workspace
+    echo "LAD Workspace" > $CONF_DIR/.multihoming_marker
+    save_config
+    configure_logrotate
+}
+
 onboard()
 {
     if [ $VERBOSE -eq 1 ]; then
@@ -362,6 +377,11 @@ onboard()
 
     if [ "$WORKSPACE_ID" = "scom" ]; then
         onboard_scom
+        clean_exit $?
+    fi
+
+    if [ "$WORKSPACE_ID" = "LAD" ]; then
+        onboard_lad
         clean_exit $?
     fi
 


lagalbra:
@Microsoft/omsagent-devs

installer/scripts/omsadmin.sh
```diff
@@ -349,6 +349,21 @@ onboard_scom()
     make_dir $CONF_DIR/omsagent.d
     #Always register SCOM as secondary Workspace
     echo "SCOM Workspace" > $CONF_DIR/.multihoming_marker
+    save_config
```
lagalbra: Why did you add a save_config call to both onboard_scom and onboard_lad? That method writes to `omsadmin.conf`, which is not an appropriately named file for you to store your variables.

hosungsmsft: Thanks. I'll remove this call from onboard_scom (I shouldn't have assumed anyway), but for LAD, I wasn't able to run the service_control script without creating omsadmin.conf, so that's why I added. I still would like to rely on the service_control script to start/stop omsagent, so I thought I need omsadmin.conf anyway. Please correct me if I'm wrong.

installer/scripts/omsadmin.sh
```diff
@@ -349,6 +349,21 @@ onboard_scom()
...
+    configure_logrotate
+}
+
+onboard_lad()
+{
+    echo "Onboarding LAD"
+    create_workspace_directories "LAD"
+
+    echo "@include omsagent.d/*" > $CONF_DIR/omsagent.conf
```
hosungsmsft: Will do. Thanks.

lagalbra:
If you rely on service_control to start/stop an omsagent daemon (which I think you may have to create a copy of, given that Kris said the daemon points to the FluentD root .conf file which you must populate), you have to add this line, like `onboard()` has:
`$SERVICE_CONTROL start "LAD"`
Will you be including this piece in a separate PR?

installer/scripts/omsadmin.sh
lagalbra: Yes, currently service_control relies on omsadmin.conf being present.

hosungsmsft:
No, I don't think I'll be including the line ('service_control start LAD') in omsadmin.sh in this or a separate PR. Currently omsagent.conf and omsagent.d/* are blank (except '@include omsagent.d/*' in omsagent.conf) after initial onboarding, so I don't think it makes sense to start the service from omsadmin.sh. I'm thinking of calling 'service_control start LAD' separately from LAD extension, after configuring omsagent.d/* according to the passed LAD settings.

PR #418, opened 2017-03-20
Update syslog config for LAD support

hosungsmsft:
LAD still needs to configure rsyslog (new/old) or syslog-ng for syslog logging even with omsagent, and I'd definitely prefer to utilize the existing configure_syslog.sh capability, so this change is for that.

The omsagent-side syslog config is NOT included here, and maybe not in the future either. That'll be done by LAD, and I wanted to rely on configure_syslog.sh for restarting whatever syslog daemon, so that's why I added RestartSyslog function in the script.

Built/tested without any issue and manually dev-tested with the resulting universal shell bundle (for LAD scenario only, though).


lagalbra:
@Microsoft/omsagent-devs

installer/conf/rsyslog-lad.conf
```diff
@@ -0,0 +1,2 @@
+# OMS Syslog collection for workspace LAD (SYSLOG_PORT=%SYSLOG_PORT%)
```
lagalbra: nit: remove "OMS" from the descriptive text, since OMS service really isn't involved.

installer/conf/syslog-ng-lad.conf
```diff
@@ -0,0 +1,31 @@
+# OMS Syslog collection for workspace LAD (SYSLOG_PORT=%SYSLOG_PORT%) (LAD_oms marker for unconfig)
```
lagalbra: nit: remove OMS here too

installer/conf/syslog-ng-lad.conf
```diff
@@ -0,0 +1,31 @@
...
+# To be added for each facility/min-level: log { source(src); filter(f_LAD_oms_f_<facility>); filter(f_LAD_oms_ml_<min_level>); destination(d_LAD_oms); };
+destination d_LAD_oms { udp("127.0.0.1" port(%SYSLOG_PORT%)); };
```
lagalbra: To follow [convention](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/conf/syslog-ng.conf), the destination line should be after the filter lines.

With this, why don't you have a log line?

installer/conf/rsyslog-lad.conf
lagalbra: To have these `%SYSLOG_PORT%` and other similar strings replaced, you will eventually need logic in `omsadmin.sh` or other script to replace them in these files.

lagalbra:
@hosungsmsft General note: when contributing to the OMSAgent and related repositories, tag our omsagent-devs team.

installer/conf/rsyslog-lad.conf
hosungsmsft: I'd prefer not to, because removing "OMS" will break the UnconfigureRsyslog() function at [here](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/scripts/configure_syslog.sh#L72). All the code changes here were deliberately chosen to make them work with the existing configure_syslog.sh and/or omsadmin.sh logic, without disturbing them too much.

hosungsmsft: No separate logic needed. It's already done in the existing configure_syslog.sh [here](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/scripts/configure_syslog.sh#L50), and that's why chose the name.

installer/conf/syslog-ng-lad.conf
hosungsmsft: No, because of the same reason.

installer/conf/syslog-ng-lad.conf
hosungsmsft: No problem. Moved.

I don't have a log line here, because I don't want rsyslog/syslog-ng to start sending logs to the configured ports when there's no omsagent started (remember I don't start it when onboarding). Only after LAD starts omsagent using service_control start LAD, LAD will configure rsyslog/syslog-ng (with log lines) and restart syslog.

hosungsmsft:
@lagalbra Thanks. Will make sure to tag omsagent-devs in the future.

Responded to your comments and pushed another commit to address them. Thanks.

PR #419, opened 2017-03-22
Fix docker provider install script to account for both 17.* and 1.* v…

samisms:
Fix docker provider install script to account for both 17.* (new docker-ee and docker-ce) and 1.1*  versionings 
@Microsoft/omsagent-devs @keikhara


installer/oss-kits/docker-oss-test.sh
```diff
@@ -21,4 +23,4 @@ if [ "$(uname -m)" == "x86_64" ]; then
 else
     echo "  Only 64-bit systems are supported. Docker agent will not be installed." 1>&2
     exit 1
-fi
\ No newline at end of file
+fi
```
lagalbra: nit: undo this change unless you got some error without redoing this line.

samisms: Artefact of vim. Just a new line. Should I still change it ?

lagalbra: I guess it doesn't really matter; this doesn't look like one of the scripts that's super picky.

samisms: I fixed it anyway. b5fae9e

sugr4:
Please squash your commits to one single commit, and rebase to master before pushing and merging with master

installer/oss-kits/docker-oss-test.sh
NarineM: It matters actually in Linux files. some tools still require empty line at the end of the file. Let us make sure we do not break that.

PR #420, opened 2017-03-24
Update path for container.conf if exists

samisms:
Update path variables in container.conf similar to other conf files
@Microsoft/omsagent-devs  @keikhara

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -813,6 +813,9 @@ copy_omsagent_d_conf()
 
     update_path $OMSAGENTD_DIR/monitor.conf
     update_path $OMSAGENTD_DIR/heartbeat.conf
+    if [ -f $OMSAGENTD_DIR/container.conf ] ; then
+        update_path $OMSAGENTD_DIR/container.conf
+    fi
 
     chown_omsagent $OMSAGENTD_DIR/*
 }


installer/scripts/omsadmin.sh
```diff
@@ -813,6 +813,9 @@ copy_omsagent_d_conf()
 
     update_path $OMSAGENTD_DIR/monitor.conf
     update_path $OMSAGENTD_DIR/heartbeat.conf
+    if [ -s $OMSAGENTD_DIR/container.conf ] ; then
+      update_path $OMSAGENTD_DIR/container.conf
```
lagalbra: For consistency, use 4 spaces for indentation.

installer/scripts/omsadmin.sh
```diff
@@ -813,6 +813,9 @@ copy_omsagent_d_conf()
 
     update_path $OMSAGENTD_DIR/monitor.conf
     update_path $OMSAGENTD_DIR/heartbeat.conf
+    if [ -s $OMSAGENTD_DIR/container.conf ] ; then
```
lagalbra: Any reason you're using -s instead of -f? The rest of the file uses -f to check.

samisms: I just did a file exists check, will change it to -f to be consistent with the rest of the file

PR #422, opened 2017-03-27
Make OMSAgent using SCX installer kit not branch

lagalbra:
Copied new SCX shell files for create_bundle
Modified bundle files to get and ball the SCX installer script

@Microsoft/omsagent-devs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ git:
 
 before_install:
   - sudo apt-get update -qq
-  # Pre-reqs for building omsagent
-  - sudo apt-get install -qq pkg-config make ruby bison g++ rpm librpm-dev libpam0g-dev libssl-dev libmysqlclient-dev
+  # Pre-reqs for building omsagent and omi
+  - sudo apt-get install -qq pkg-config make ruby bison g++ rpm librpm-dev libpam0g-dev libssl-dev libmysqlclient-dev libkrb5-dev
   # Pre-reqs for building auoms
   - sudo apt-get install cmake libboost-dev libboost-test-dev libaudit-dev libauparse-dev unzip
   - sudo groupadd omiusers
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -34,10 +34,9 @@ OMS_CONSISTENCY_INVOKER="/etc/cron.d/OMSConsistencyInvoker"
 # These symbols will get replaced during the bundle creation process.
 
 TAR_FILE=<TAR_FILE>
-OMI_PKG=<OMI_PKG>
 OMS_PKG=<OMS_PKG>
 DSC_PKG=<DSC_PKG>
-SCX_PKG=<SCX_PKG>
+SCX_INSTALLER=<SCX_INSTALLER>
 INSTALL_TYPE=<INSTALL_TYPE>
 SCRIPT_LEN=<SCRIPT_LEN>
 SCRIPT_LEN_PLUS_ONE=<SCRIPT_LEN+1>
@@ -366,24 +365,6 @@ getInstalledVersion()
     fi
 }
 
-shouldInstall_omi()
-{
-    local versionInstalled=`getInstalledVersion omi`
-    [ "$versionInstalled" = "None" ] && return 0
-    local versionAvailable=`getVersionNumber $OMI_PKG omi-`
-
-    check_version_installable $versionInstalled $versionAvailable
-}
-
-shouldInstall_scx()
-{
-    local versionInstalled=`getInstalledVersion scx`
-    [ "$versionInstalled" = "None" ] && return 0
-    local versionAvailable=`getVersionNumber $SCX_PKG scx-cimprov-`
-
-    check_version_installable $versionInstalled $versionAvailable
-}
-
 shouldInstall_omsagent()
 {
     local versionInstalled=`getInstalledVersion omsagent`
@@ -500,17 +481,8 @@ do
         --version-check)
             printf '%-15s%-15s%-15s%-15s\n\n' Package Installed Available Install?
 
-            # omi
-            versionInstalled=`getInstalledVersion omi`
-            versionAvailable=`getVersionNumber $OMI_PKG omi-`
-            if shouldInstall_omi; then shouldInstall="Yes"; else shouldInstall="No"; fi
-            printf '%-15s%-15s%-15s%-15s\n' omi $versionInstalled $versionAvailable $shouldInstall
-
-            # scx
-            versionInstalled=`getInstalledVersion scx`
-            versionAvailable=`getVersionNumber $SCX_PKG scx-cimprov-`
-            if shouldInstall_scx; then shouldInstall="Yes"; else shouldInstall="No"; fi
-            printf '%-15s%-15s%-15s%-15s\n' scx $versionInstalled $versionAvailable $shouldInstall
+            # scx (and omi)
+            ./bundles/$SCX_INSTALLER --version-check
 
             # OMS agent itself
             versionInstalled=`getInstalledVersion omsagent`
@@ -545,9 +517,10 @@ do
             echo "" >&2
             echo "SCRIPT_INDIRECT: $SCRIPT_INDIRECT" >&2
             echo "SCRIPT_DIR:      $SCRIPT_DIR" >&2
-            echo "EXTRACT DIR:     $EXTRACT_DIR" >&2
+            echo "EXTRACT_DIR:     $EXTRACT_DIR" >&2
             echo "SCRIPT:          $SCRIPT" >&2
             echo >&2
+            debugMode=true
             set -x
             shift 1
             ;;
@@ -649,7 +622,13 @@ if [ "$installMode" = "R" -o "$installMode" = "P" ]; then
 
         # If MDSD is installed and we're just removing (not purging), leave SCX
 
-        if [ ! -d /var/lib/waagent/Microsoft.OSTCExtensions.LinuxDiagnostic-*/mdsd -o "$installMode" = "P" ]; then
+        MDSD_INSTALLED=1
+        check_if_pkg_is_installed azsec-mdsd
+        if [ $? -eq 0 -o -d /var/lib/waagent/Microsoft.OSTCExtensions.LinuxDiagnostic-*/mdsd ]; then
+            MDSD_INSTALLED=0
+        fi
+
+        if [ $MDSD_INSTALLED -ne 0 -o "$installMode" = "P" ]; then
             if [ -f /opt/microsoft/scx/bin/uninstall ]; then
                 /opt/microsoft/scx/bin/uninstall $installMode
             else
@@ -784,13 +763,19 @@ case "$installMode" in
         if [ $scx_installed -ne 0 -a $omi_installed -ne 0 ]; then
             echo "Installing OMS agent ..."
 
-            pkg_add_list $OMI_PKG omi
-            pkg_add_list $SCX_PKG scx
             pkg_add_list $OMS_PKG omsagent
 
             python_ctypes_installed
             if [ $? -eq 0 ]; then
                 pkg_add_list $DSC_PKG omsconfig
+            fi
+
+            # Install SCX (and OMI)
+            [ -n "${forceFlag}" ] && FORCE="--force" || FORCE=""
+            [ -n "${debugMode}" ] && DEBUG="--debug" || DEBUG=""
+            ./bundles/${SCX_INSTALLER} --install $FORCE $DEBUG
+            if [ "$?" -ne 0 ]; then
+                cleanup_and_exit 1
             fi
 
             # Now actually install of the "queued" packages
@@ -848,12 +833,6 @@ case "$installMode" in
     U)
         echo "Updating OMS agent ..."
 
-        shouldInstall_omi
-        pkg_upd_list $OMI_PKG omi $?
-
-        shouldInstall_scx
-        pkg_upd_list $SCX_PKG scx $?
-
         shouldInstall_omsagent
         pkg_upd_list $OMS_PKG omsagent $?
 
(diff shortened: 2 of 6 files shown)


lagalbra:
@vimish For context

lagalbra:
This PR cannot be committed until the PR is completed in the superproject to add the SCX kits submodule: https://github.com/Microsoft/Build-OMS-Agent-for-Linux/pull/24

installer/bundle/bundle_skel.sh
```diff
@@ -500,17 +481,8 @@ do
...
-            if shouldInstall_omi; then shouldInstall="Yes"; else shouldInstall="No"; fi
-            printf '%-15s%-15s%-15s%-15s\n' omi $versionInstalled $versionAvailable $shouldInstall
-
-            # scx
-            versionInstalled=`getInstalledVersion scx`
-            versionAvailable=`getVersionNumber $SCX_PKG scx-cimprov-`
-            if shouldInstall_scx; then shouldInstall="Yes"; else shouldInstall="No"; fi
-            printf '%-15s%-15s%-15s%-15s\n' scx $versionInstalled $versionAvailable $shouldInstall
+            # scx (and omi)
+            ./bundles/$SCX_INSTALLER --version-check
```
jeffaco: Nice catch, I love that you caught details like this. Awesome.

installer/bundle/bundle_skel.sh
```diff
@@ -548,6 +520,7 @@ do
             echo "EXTRACT DIR:     $EXTRACT_DIR" >&2
             echo "SCRIPT:          $SCRIPT" >&2
             echo >&2
+            debugMode=true
```
jeffaco: Umm, I'm not sure why this is here. Is this a debugging mistake?

installer/bundle/bundle_skel.sh
```diff
@@ -759,11 +732,21 @@ then
     echo "Failed: could not extract the install bundle."
     cleanup_and_exit ${STATUS}
 fi
+[ -n "${forceFlag}" ] && FORCE="--force" || FORCE=""
+[ -n "${debugMode}" ] && DEBUG="--debug" || DEBUG=""
+./bundles/${SCX_INSTALLER} --extract $FORCE $DEBUG
```
jeffaco: I'm not sure why you're extracting the shell bundle like this. Why not leave the shell bundle in the list of extracted files as is (as a `.sh` file), and then just run it when you need to for installation or upgrade?

If they really did want SCX extracted, then they can run that shell bundle with --extract afterwards. But given that the bundle skeleton file should just execute the SCX shell bundle as a `.sh` file, why not just leave it in that form?

installer/bundle/bundle_skel.sh
```diff
@@ -784,15 +767,19 @@ case "$installMode" in
...
             python_ctypes_installed
             if [ $? -eq 0 ]; then
                 pkg_add_list $DSC_PKG omsconfig
             fi
 
+            # Install SCX (and OMI)
+            [ -n "${forceFlag}" ] && FORCE="--force" || FORCE=""
+            [ -n "${debugMode}" ] && DEBUG="--debug" || DEBUG=""
+            ./bundles/${SCX_INSTALLER} --install $FORCE $DEBUG
+            SCX_STATUS=$?
```
jeffaco: If SCX failed, is it worth continuing? Perhaps we should just abort at that point?

In what cases would it make sense to continue installation even if SCX couldn't be installed?

installer/bundle/bundle_skel.sh
```diff
@@ -784,15 +767,19 @@ case "$installMode" in
...
+
             # Now actually install of the "queued" packages
             if [ -n "${add_list}" ]; then
                 if [ "$INSTALLER" = "DPKG" ]; then
```
jeffaco: So historically, we thought that bundling packages was more reliable for installation. But then we came up with cases where we were better off without bundled installations and having a recovery procedure for SCX.

Which is better for OMS? Is there a reason we should keep bundled package installation, or should we go back to individual package installation, like it used to be? (We changed it to be consistent with SCX because at the time we thought that was best ...)

I don't know the answer, but wanted you to think about it, so I'm bringing it up ... I'm likely to accept any well thought out, reasoned answer. We do have a little evidence, though, that bundled installations don't always pan out as we hoped they would (thus the recovery procedure).

installer/bundle/bundle_skel.sh
```diff
@@ -863,6 +844,12 @@ case "$installMode" in
             pkg_upd_list $DSC_PKG omsconfig $?
         fi
 
+        # Install SCX (and OMI)
+        [ -n "${forceFlag}" ] && FORCE="--force" || FORCE=""
+        [ -n "${debugMode}" ] && DEBUG="--debug" || DEBUG=""
+        ./bundles/${SCX_INSTALLER} --upgrade $FORCE $DEBUG
+        SCX_STATUS=$?
```
jeffaco: Same here. In what cases does it make sense to continue installation if low-level dependencies couldn't be upgraded for some reason?

One might argue that it could improve the SLA, and I'd agree. But at what cost? Is it worth improving the SLA when you're running older versions of SCX that are no longer recommended?

installer/bundle/bundle_skel.sh
```diff
@@ -930,7 +917,7 @@ esac
 
 # Remove temporary files (now part of cleanup_and_exit) and exit
 
-if [ "$KIT_STATUS" -ne 0 -o "$BUNDLE_EXIT_STATUS" -ne 0 ]; then
+if [ "$SCX_STATUS" -ne 0 -o "$KIT_STATUS" -ne 0 -o "$BUNDLE_EXIT_STATUS" -ne 0 ]; then
     cleanup_and_exit 1
```
jeffaco: So if SCX failed, we do return an error. I guess the only question is: Is it worth trying to do the other things before returning the error? I'm not sure, but we should think about this.

installer/bundle/bundle_skel.sh
lagalbra: No, this was intended. Since I tried to pass the same arguments that the omsagent bundle was given, and pass to the SCX installer, I wanted to track whether --debug was used in the same way that forceFlag is used and tracked.

installer/bundle/bundle_skel.sh
lagalbra: It depends; what should we do, from a design perspective? I hadn't heard of the --extract use case until I started looking at this last week.

From your comments, I think the user can call --extract on the SCX installer separately. I will make this change.

installer/bundle/bundle_skel.sh
lagalbra: No cases that I know of. I will make this change.

installer/bundle/bundle_skel.sh
lagalbra: For packages that are not installed by the SCX installer (omsconfig, omsagent), I know of no defined recovery procedure. For SCX, you provided a defined approach that SCX/OMI should follow when installing in case of conflicts. However, I have no evidence that bundled installations of omsconfig/omsagent aren't the best way to proceed.

In my opinion, we should keep the status quo for the package installations that we haven't seen related errors for. Changing it to be non-bundled merely to be consistent with SCX doesn't seem like the best approach at this fork in the road.

installer/bundle/bundle_skel.sh
lagalbra: Will change this.

installer/bundle/bundle_skel.sh
lagalbra: Will change this.

installer/bundle/bundle_skel.sh
jeffaco: I'm fine with this. I just wanted you folks to give it consideration and decide what you wanted. I agree with your reasoning for now (without evidence to the contrary, why make the change). Once we have more solid data, we'll know (for sure) where to go.

In the case of SCX, we do have solid data (with at least one data point). Not so for OMS.

jeffaco:
I think everything is addressed (or in flight) except for my last comment in the review:

I didn't see anything for --purge handling or --remove handling. Don't we want to call the SCX shell bundle for --remove or --purge, if that was specified in the shell bundle for OMS?

If OMS installed SCX, OMS should remove it as well with --remove or --purge (assuming nobody else is dependent on it).

lagalbra:
@jeffaco For the --purge handling or --remove handling, will the SCX installer file exist in the `EXTRACT_DIR`?

I believe the original reason I kept the current pkg_rm scx approach was because the [SCX purge/remove code](https://github.com/Microsoft/SCXcore/blob/master/installer/bundle/bundle_skel_Linux.sh#L495) looks a little different from ours, since we operate under the assumption that something else may be using SCX, while the SCX installer knows that it is only called when no one wants it there anymore. This results in SCX not checking if it's still installed when it [deletes its directories](https://github.com/Microsoft/SCXcore/blob/master/installer/bundle/bundle_skel_Linux.sh#L517) and [omsagent should still make that check](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/scripts/uninstall.sh#L109).

Calling --remove or --purge from the omsagent may take some thinking, and it may only get called after pkg_rm scx is called anyway. Let me know if you have any ideas about this.

jeffaco:
That's a good point. I would argue that that's a bug in the SCX bundle code. If you call --remove, then we should first remove the package (AND ONLY DELETE DIRECTORIES if there are no dependencies). This, I think, should be the case for layered providers (MySQL, Apache, Docker, etc), as well as SCX itself.

It's interesting that the bundle handling code does manage OMI properly, but not SCX.

@vimish What are your thoughts on this? Laura properly pointed out that SCX --remove will remove it's directories even if some other project is dependent on it. Should the SCX shell bundle be smarter about that, or should OMS just remove the code itself as it does (pre-check if the package was removed prior to cleanup)?

@lagalbra I'm not sure if the shell bundle contents are extracted in order to remove/purge. It may not be, but perhaps we should change that if we end up having OMS call SCX to remove/purge itself.

vimish:
@jeffaco @lagalbra I think that SCX installer should handle the remove and purge of scx package and scx bundle should check for presence of scx package before removing the directories.
I have added the check in my current [PR](https://github.com/Microsoft/SCXcore/pull/73) on scxcore.

installer/bundle/bundle_skel.sh
vimish: When I was trying to repro the errors (upgrade using omsagent1.2.0-148 on ubuntu 14) I noticed that omsconfig was trying to start omsagent before any package were configured. It didn't caused any failures but I thought this is something we should avoid and do a package wise update. Not sure it will improve anything (also I haven't tried it with latest omsagent bundles) but just wanted to bring out my observation.

lagalbra: Adding Jeff's comment from the email thread:

This is a good point, and I didn’t want this to be lost. I’ve seen this too in the past, typically on Ubuntu, but I haven’t tried to replicate lately.

I think, perhaps due to bundling of installations, perhaps for some other reasons, Ubuntu can try and start services before the packages are fully configured, causing errors. This, if memory serves, appeared to be Ubuntu-specific (not on CentOS, for example).

We should definitely test these cases and make sure that sort of thing doesn’t happen in installation/upgrade scenarios.

lagalbra:
Small issue with calling --remove/purge on the SCX installer: that file does NOT [get extracted](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/bundle/bundle_skel.sh#L753) out of the OMS bundle before the [uninstall code](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/bundle/bundle_skel.sh#L629) runs. Thus, we either need to extract earlier and verify [cleanup](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/bundle/bundle_skel.sh#L719) in the --remove/purge method as well

lagalbra:
@jeffaco Can you explain why the build on Travis (and therefore Ubuntu 14 systems) is failing, but it passes on my pBuild machine? The OMI_LIBRARY_DIR is a little different on non-ULINUX systems: -L<build_dir>/omi/Unix/output/lib

jeffaco:
Your Travis definitions are a little whacky, by the way.  Looks like Suju set up Travis originally (a good thing), but was a bit confused on how to handle the subproject/superproject thing. You might want to look at [SCXcore](https://github.com/Microsoft/SCXcore/blob/master/.travis.yml) for a more simple, easy way to deal with that.

Your PBUILD system is likely set up for ULINUX, but Travis is not (to set up Travis for that, we'd need to build and install multiple versions of SSL, and that's not worth it). Given that Ubuntu isn't a distribution platform anyway (you don't build there for distribution), I'd argue that your Ubuntu system likely shouldn't have the multiple OpenSSL versions installed, and then Ubuntu should build similarly to Travis.

If this doesn't make sense, let me know, and I'll come visit.

lagalbra:
Remove/purge strategy: after discussion with Jeff, the omsagent remove/purge code will stay the same. The SCX installer method of removing/purging currently removes directories earlier than we want to, so a later sync between SCX/OMI/OMSAgent will be necessary to properly depend on each other's uninstall scripts rather than our own code for removing packages and directories.

Travis should succeed now. @jeffaco please approve the current commit if you don't see anything I haven't covered.

.travis.yml
```diff
@@ -7,8 +7,8 @@ git:
 
 before_install:
   - sudo apt-get update -qq
-  # Pre-reqs for building omsagent
-  - sudo apt-get install -qq pkg-config make ruby bison g++ rpm librpm-dev libpam0g-dev libssl-dev libmysqlclient-dev
+  # Pre-reqs for building omsagent and omi
+  - sudo apt-get install -qq pkg-config make ruby bison g++ rpm librpm-dev libpam0g-dev libssl-dev libmysqlclient-dev libkrb5-dev
```
jeffaco: I'd probably break omi and omsagent into separate lines to keep them independent. But no biggie either way.

PR #432, opened 2017-04-05
Update proxy documentation with multi-homing workaround

lagalbra:
@Microsoft/omsagent-devs 

A regression was discovered; proxy will not work with a multi-homing bundle because DSC updates some plugins to look for proxy config in the old location. The code will be updated soon, but the workaround is documented in this review's changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,8 +44,8 @@ We love feedback!  Whether it be good, bad or indifferent, it really helps us bu
 ### [Zabbix alert collection](docs/OMS-Agent-for-Linux.md#viewing-zabbix-alerts)
 ### [CollectD Metrics Collection](docs/OMS-Agent-for-Linux.md#collectd-metrics)
 ### [Custom JSON Data](docs/OMS-Agent-for-Linux.md#custom-json-data-sources)
-### [VMware Monitoring] (docs/VMware-Instruction.md)
-## [Full documentation](docs/OMS-Agent-for-Linux.md)
+### [VMware Monitoring](docs/VMware-Instruction.md)
+### [Full documentation](docs/OMS-Agent-for-Linux.md)
 
 ## [Troubleshooting Guide](docs/Troubleshooting.md)
 
--- a/docs/Docker-Instructions.md
+++ b/docs/Docker-Instructions.md
@@ -27,14 +27,14 @@ And one of following OS (x64):
 - ACS Kubernetes 1.4.5
 
 ## Release Note
-Update Information is [here.](https://github.com/Microsoft/OMS-docker/blob/keikoacs/ReleaseNote.md)
+Update Information is [here.](https://github.com/Microsoft/OMS-docker/blob/master/ReleaseNote.md)
 
 ## Setting up
 As a pre-requisite, docker must be running prior to this OMS Linux Agent installation. If you have installed before running docker, please re-install OMS Agent. For more information about docker, please go to https://www.docker.com/.
 
 This set up is not for ACS Mesosphere DC/OS or ACS Kubernetes. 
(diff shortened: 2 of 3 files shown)


keikhara:
@lagalbra per our conversation, we should properly clarify the instruction. As far as I understood, you will need to have a "mandatory" user/password for the proxy. In case the customer does not have user/password, they will need to add a fake one. We should add those details in. We had a customer call today and this was presented back to us saying all was optional which was failing.

lagalbra:
@keikhara Removed the use of "optional". If you want to add more details to the doc, could you create a separate PR? I would rather commit this one ASAP so customers can see the workaround on our main page.

NarineM:
@lagalbra 
Please put the workaround in all release branches that contain multi -homing (starting 1.3..) instead of master. We are going to fix this regression in the master branch so for the documentation on master branch only optional user/password part should stay.

lagalbra:
@NarineM Do you mean that the "optional" wording should stay in master, but the workaround should still be merged into master? Or did you mean that this PR's changes should only be merged into 1.3.* release branches and not master?

If you meant the latter, I think the workaround changes at least must be merged into master because the doc on the master branch is referenced in several Microsoft docs ([example](https://docs.microsoft.com/en-us/azure/log-analytics/log-analytics-linux-agents)) when instructing customers to onboard OMSAgent with proxy.

If you meant the former, then I might be inclined to agree, but Keiko may disagree.

lagalbra:
@NarineM Regardless, I will merge the workaround commit into the 1.3.* release branches now.

NarineM:
@lagalbra,
If user/password optional parameter is also a regression, then no need to change doc in the working branch at all.

lagalbra:
Commit hash 32f93e4 fixed a typo.

@keikhara Can you give details about the scenario where you found username/password to be required? Robbie's initial multi-homing feature ([PR here](https://github.com/Microsoft/OMS-Agent-for-Linux/pull/329)) did not change any requirements for user/password with proxy. My guess is that it is a server-side configuration that needs a username/password to be provided when communicating with the agent.

lagalbra:
@KrisBash Can you comment on the proxy support and when user/password must be specified? I ask because you are the one who originally documented that support in [this commit](https://github.com/Microsoft/OMS-Agent-for-Linux/commit/e44752dd4061e48386ddce6e71d13ff4176c5054#diff-cd6e5517745e5df6839a6c33459996eaR122).

keikhara:
+Sambit

We were with T-Systems ( Deustch Telekcom ) call and they had this very case. We had to add make a mandatory add of user/password to be added. The customer did read github and followed it properly but they lost data coming in and that caused the escalation.

We need to be more careful with our documentation and what we are providing. Things will need to be precise so we do not get these calls.

lagalbra:
Adding @samisms to this thread

Commit hash efdf11a gives details and examples for mandatory username/password strings, and specifies that they can be placeholder (non-meaningful) values. @keikhara Please review.

NarineM:
Let us hold on on the documentation changes till it is certain what parts are regressions what is by design. It seems really weird requiring users put fake username and password.

keikhara:
Agree.

lagalbra:
@keikhara @NarineM Can I change this PR to only include the changes that give the workaround for the proxy.conf path, and merge with master then? Since the regression will take a while to fix, I want to have the workaround accessible via documentation soon rather than later.

lagalbra:
Above changed made. Other changes are saved in branch [lagalbra-proxy-optional-fields](https://github.com/Microsoft/OMS-Agent-for-Linux/tree/lagalbra-proxy-optional-fields)

PR #435, opened 2017-04-06
Improve detection of MDSD

lagalbra:
@Microsoft/omsagent-devs @hosungsmsft 

This change is to improve the detection of mdsd on the machine when removing OMSAgent.
If OMSAgent is removed (and not purged), then SCX/OMI packages will not be removed if mdsd is detected on the machine. This will be true when LinuxAsm or LinuxDiagnostic extensions are installed on the machine, which are dependent on SCX/OMI packages.
Those teams have been notified and their explicit package dependencies will also be improved from their side.

I have verified that the omsagent built with this change does detect and does not remove SCX/OMI packages in the --remove scenario.
pBuild is currently running.

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -627,9 +627,10 @@ if [ "$installMode" = "R" -o "$installMode" = "P" ]; then
 
         MDSD_INSTALLED=1
         check_if_pkg_is_installed azsec-mdsd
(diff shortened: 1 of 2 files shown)


lagalbra:
@hosungsmsft That's fine. I think this is a design that we should consider more as your LAD workspace design is discussed. I will close this PR until these decisions are finalized.

One thing I would like to include in these discussions is why the LinuxDiagnostic extension will ever run --remove on our bundle - what if our OmsAgentForLinux extension is installed, or the customer still expects data to be collected for OMS workspaces?

PR #436, opened 2017-04-07
add temporary workaround

keikhara:
@NarineM @lagalbra 

removed "Optional" on http proxy until we have fixed the user/password. Also added pseudo user/password usage for customers who does not have user/password set on proxy. This is based on a customer ICM call. 

Please go thru and see whether it reads better.

### DIFF
--- a/docs/OMS-Agent-for-Linux.md
+++ b/docs/OMS-Agent-for-Linux.md
@@ -137,13 +137,16 @@ The proxy configuration value has the following syntax:
 Property|Description
 -|-
 Protocol|http or https
-user|Optional username for proxy authentication
-password|Optional password for proxy authentication
+user|username for proxy authentication
+password|password for proxy authentication
 proxyhost|Address or FQDN of the proxy server
 port|Optional port number for the proxy server
 
 For example:
 http://user01:password@proxy01.contoso.com:8080
+
+*Note: Although you do not have any user/password set for the proxy, you will still need to add a psuedo user/password. This can be any username or password.    
+(This will be enhanced in future so that these psuedo user/password will not be necessary)*
 
 The proxy server can be specified during installation or directly in a file (at any point). 
 


lagalbra:
My branch with this doc change is here: [lagalbra-proxy-optional-fields](https://github.com/Microsoft/OMS-Agent-for-Linux/compare/lagalbra-proxy-optional-fields). Feel free to take any of those changes if you want.

docs/OMS-Agent-for-Linux.md
```diff
@@ -137,14 +137,17 @@ The proxy configuration value has the following syntax:
 Property|Description
 -|-
 Protocol|http or https
-user|Optional username for proxy authentication
-password|Optional password for proxy authentication
+user|username for proxy authentication
+password|password for proxy authentication
```
lagalbra: Please capitalize Username and Password for readability

docs/OMS-Agent-for-Linux.md
```diff
@@ -137,14 +137,17 @@ The proxy configuration value has the following syntax:
...
 proxyhost|Address or FQDN of the proxy server
 port|Optional port number for the proxy server
 
 For example:
 http://user01:password@proxy01.contoso.com:8080
```
lagalbra: I think we should add another example here, like you suggested previously.

docs/OMS-Agent-for-Linux.md
```diff
@@ -137,14 +137,17 @@ The proxy configuration value has the following syntax:
...
 
+*Note: Although you do not have any user/password set for the proxy, you will still need to add a psuedo user/password. This can be any username or password.    
+(This will be enhanced in future so that these psuedo user/password will not be necessary)*
```
lagalbra: nit: phrase it like so: "This will be enhanced in the future so that a psuedo username/password will not be necessary."

docs/OMS-Agent-for-Linux.md
keikhara: sure I agree. Let me add that.

PR #438, opened 2017-04-07
Remove unused SCXCore code after SCX refactoring

lagalbra:
@Microsoft/omsagent-devs 

This change is to stop running configure on SCXCore
As of this commit, we no longer build SCX, so this is unneeded code.
After this is checked in, I will submit a PR (from [this branch](https://github.com/Microsoft/Build-OMS-Agent-for-Linux/compare/lagalbra-remove-unused-opsmgr)) to remove the SCXCore (opsmgr) submodule from the superproject.

Testing completed:
Unit and system tests pass
Installing/onboarding verified
pBuild is running now


NarineM:
good catch! I saw this too in the brownbag session when talking about submodules. Let us remove the dead code!

PR #449, opened 2017-04-20
npmd: make sure npmdclient sock is reset in shutdown

chparimi:
Certain times the npmdClientSock might not be set back to nil
in unit tests as the npmd_reader thread might not set it in
time. This is more of a fail safe for unit tests.


chparimi:
@Microsoft/omsagent-devs

PR #450, opened 2017-04-20
out_oms_diag: adding out_oms_diag plugin for diagnostic logging

chparimi:
Adding new output plugin that would send data to DiagnosticEvent
channel.

Utilize the helper functions provided in oms_diag for logging to
diagnostic channel.


chparimi:
Need to check if oms_diag.rb should be changed to oms_diag_lib.rb so that some unit tests could run on it.
Also need to write down few unit tests for out_oms_diag.

installer/conf/omsagent.conf
```diff
@@ -97,6 +97,29 @@
...
 
+<match diag.oms diag.oms.**>
+  type out_oms_diag
+  log_level info
+  num_threads 5
+
+  omsadmin_conf_path %CONF_DIR_WS%/omsadmin.conf
+  cert_path %CERT_DIR_WS%/oms.crt
+  key_path %CERT_DIR_WS%/oms.key
+  # TODO: Remove the below line when ods supports decompression
```
chparimi: This will be removed once ODS supports decompression

source/code/plugins/out_oms_diag.rb
```diff
@@ -0,0 +1,112 @@
...
+      require_relative 'oms_configuration'
+      require_relative 'oms_common'
+      require_relative 'oms_diag'
+    end
+
+    config_param :omsadmin_conf_path, :string, :default => '/etc/opt/microsoft/omsagent/conf/omsadmin.conf'
+    config_param :cert_path, :string, :default => '/etc/opt/microsoft/omsagent/certs/oms.crt'
+    config_param :key_path, :string, :default => '/etc/opt/microsoft/omsagent/certs/oms.key'
+    config_param :proxy_conf_path, :string, :default => '/etc/opt/microsoft/omsagent/proxy.conf'
+    # TODO: To make compress default to true once supported
```
chparimi: This is kept here rather than in omsagent.conf file because from DSC we can update this file once compression is supported. However I did not see omsagent.conf getting updated using that mechanism.

test/code/plugins/oms_diag_test.rb
```diff
@@ -0,0 +1,243 @@
...
+            dataitem = Hash.new
+            dataitem[OMS::Diag::DI_KEY_LOGMESSAGE] = 'Just a test string'
+            dataitem[OMS::Diag::DI_KEY_IPNAME] = OMS::Diag::DEFAULT_IPNAME
+            dataitem[OMS::Diag::DI_KEY_TIME] = Time.now.utc.strftime("%Y-%m-%dT%H:%M:%S.%6NZ")
+            # Create few spurious cases
+            if y == 1
+                dataitem.delete(OMS::Diag::DI_KEY_LOGMESSAGE)
+            elsif y == 3
+                # note this is valid spurious here but without ipname
+                # write call of out_oms_diag will set it to default
```
chparimi: To elaborate on this comment.

- A dataitem of diagnostic log first goes from emission into the write call of output plugin as part of buffer chunk.
- There if there is no ipname then we assign that dataitem the default ipname (Diagnostics).
- The check is made later after aggregation by ipname in ProcessDataItemsPostAggregation for absence of ipname, where it will discard the dataitem (if missing ipname key)

- However while testing out_oms_diag write function, a spurious dataitem to it cannot be one that is missing ipname as it overrides with default ipname if it is missing

test/code/plugins/out_oms_diag_plugintest.rb
```diff
@@ -0,0 +1,116 @@
...
+        OMS::Configuration.configurationLoaded = true
+    end
+
+    def teardown
+    end
+
+    def create_spurious_dataitem(log, ipname)
+        dataitem = create_valid_dataitem(log, ipname)
+        # only delete logmessage or time as ipname is set to default if absent
+        # thereby having no ipname before sending to write is non spurious
```
chparimi: Check the comment above on line 159 of oms_diag_test.rb to understand this comment in source code.

source/code/plugins/out_oms_diag.rb
lagalbra: It would be possible to make a DSC module to update omsagent.conf; see [nxOMSPerfCounter](https://github.com/Microsoft/PowerShell-DSC-for-Linux/blob/master/Providers/Scripts/2.4x-2.5x/Scripts/nxOMSPerfCounter.py#L175) and how it updates perfcounter sources there.

source/code/plugins/out_oms_diag.rb
```diff
@@ -0,0 +1,112 @@
...
+      unless req.nil?
+        http = OMS::Common.create_ods_http(OMS::Configuration.diagnostic_endpoint, @proxy_config)
+        start = Time.now
+
+        # This method will raise on failure alerting the engine to retry sending this data
+        OMS::Common.start_request(req, http)
+
+        ends = Time.now
+        time = ends - start
+        count = record[OMS::Diag::RECORD_DATAITEMS].size
```
lagalbra: What if this key does not exist in record? [out_oms](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/source/code/plugins/out_oms.rb#L54) makes this check, so I would recommend making it here as well.

source/code/plugins/out_oms_diag.rb
```diff
@@ -0,0 +1,112 @@
...
+      super
+      @proxy_config = OMS::Configuration.get_proxy_config(@proxy_conf_path)
+    end
+
+    def shutdown
+      super
+    end
+
+    def handle_record(ipname, record)
+      @log.trace "Handling diag records for ipname : #{ipname}"
```
lagalbra: For logged messages, I would recommend spelling out diagnostic, for added clarity. Same for lines 51, 55, 56, 63, 70, etc.

source/code/plugins/out_oms_diag.rb
```diff
@@ -0,0 +1,112 @@
...
+      # if bad data caused the exception, the engine will continuously
+      # try and fail to resend it. (Infinite failure loop)
+      OMS::Log.error_once("Unexpecting exception, dropping diag data. Error:'#{e}'")
+    end
+
+    # This method is called when an event reaches to Fluentd.
+    # Convert the event to a raw string.
+    def format(tag, time, record)
+      if record != {}
+        @log.trace "Buffering diag #{tag}"
```
lagalbra: For clarity, I would change this message to "Buffering diagnostic record with tag #{tag}"

source/code/plugins/out_oms_diag.rb
```diff
@@ -0,0 +1,112 @@
...
+        retval = record.to_msgpack
+        return retval
+      else
+        return ""
+      end
+    end
+
+    def write(chunk)
```
lagalbra: Can you explain how you've tested this? The aggregation logic is a little mysterious to me.

source/code/plugins/oms_diag.rb
```diff
@@ -0,0 +1,135 @@
...
+      # @ipname[optional]: IPName can be optionally provided to depict a
+      # customized one other than the DEFAULT_IPNAME in diagnostic event
+      # @properties[optional]: Hash corresponding to key value pairs that
+      # would be added as part of this data item.
+      #
+      # NOTE: Certain mandatory properties to the dataitem are added by default
+      def LogDiag(logMessage, tag=DEFAULT_TAG, ipname=DEFAULT_IPNAME, properties=nil)
+        # Process default values for tag and ipname if they are passed as nil
+        tag ||= DEFAULT_TAG
+        ipname ||= DEFAULT_IPNAME
```
chparimi: Well the intention is the user can do the following:

1. LogDiag('Hi') where tag is DEFAULT_TAG, ipname is DEFAULT_IPNAME, properties are nil
2. LogDiag('Hi', nil, SomeIpName) the tag is DEFAULT_TAG, ipname is SomeIpName, properties are nil
3. LogDiag('Hi', nil, nil, SomeProperties) the tag is DEFAULT_TAG, ipname is DEFAULT_IPNAME and properties are SomeProperties

The intention is not to raise errors while reporting diagnostic logs. The intention is to allow the user to define any single parameter of the three that is intended and the remaining should utilize the default values.

source/code/plugins/oms_diag.rb
```diff
@@ -0,0 +1,135 @@
...
+
+        dataitem = Hash.new
+
+        # Adding parameterized properties
+        dataitem.merge!(properties) if properties.is_a?(Hash)
+
+        # Adding mandatory properties
+        dataitem[DI_KEY_LOGMESSAGE] = logMessage
+        dataitem[DI_KEY_IPNAME]     = ipname
+        dataitem[DI_KEY_TIME]       = Time.now.utc.strftime("%Y-%m-%dT%H:%M:%S.%6NZ")
```
chparimi: Ok will check that

source/code/plugins/oms_diag.rb
```diff
@@ -0,0 +1,135 @@
...
+        # 3. Addition of DI_KEY_TYPE key value pair to dataitem
+
+        # Emitting the record
+        Fluent::Engine.emit(tag, Fluent::Engine.now, dataitem)
+      end
+
+      # Methods for OUTPUT Plugin (out_oms_diag)
+
+      # Method:
+      # ProcessDataItemsPostAggregation(dataitems, agentId)
```
chparimi: Done. removing function names

source/code/plugins/oms_diag.rb
```diff
@@ -0,0 +1,135 @@
...
+      # @dataitems[mandatory]: Array of dataitems sent via LogDiag from
+      # Input and Filter plugins
+      # @agentId[mandatory]: The omsagent guid parsed from omsadmin.conf
+      def ProcessDataItemsPostAggregation(dataitems, agentId)
+        # Remove all invalid dataitems
+        dataitems.delete_if{|x| !IsValidDataItem?(x)}
+        for x in dataitems
+          x.delete(DI_KEY_IPNAME)
+          x[DI_KEY_TYPE] = DI_TYPE_JSON
+          x[DI_KEY_AGENTGUID] = agentId
```
chparimi: Done.

source/code/plugins/oms_diag.rb
```diff
@@ -0,0 +1,135 @@
...
+        end
+      end
+
+      # Method:
+      # CreateDiagRecord(dataitems, ipname, optionalAttributes)
```
chparimi: Done.

source/code/plugins/out_oms_diag.rb
chparimi: OK

source/code/plugins/out_oms_diag.rb
chparimi: OK

source/code/plugins/out_oms_diag.rb
chparimi: Well the aggregation is based on IPName alone as the diagnostics channel has that alone is mandatory (as per the postdaitaitems endpoint implementation). So all dataitems pertaining to a ipname are aggregating and handled together.
If there are spurious dataitems (dataitems constitute the chunk) which have missing logmessage or time in the hash then they are dropped and remaining are aggregated. the unit test file out_oms_diag_plugintest.rb will give you detailed clarity on this.

source/code/plugins/out_oms_diag.rb
chparimi: The link implies that we would require changes to the DSC module to update it. I was going for not changing something in the code for DSC for this plugin. It will be downloaded on an update to module version of Common plugins and can have the change without altering the omsagent.conf.

source/code/plugins/oms_diag.rb
chparimi: The Time format present in diagnostic logs is different from the one given out in oms_common.

- Format expected in the diagnostic logs (as per existing) is "%Y-%m-%dT%H:%M:%S.%6NZ"

- Format provided by format_time_str of oms_common is %FT%H:%M:%S.%3NZ

chparimi:
@lagalbra  Here is the testing overview with the unit-tests
image

source/code/plugins/oms_diag.rb
lagalbra: Ok, thanks for looking into that!

PR #452, opened 2017-04-20
Ensure that omsadmin.conf is removed in remove scenario

lagalbra:
@Microsoft/omsagent-devs @NarineM 

The existence of omsadmin.conf is assumed to mean that a workspace is onboarded. However, with PR https://github.com/Microsoft/OMS-Agent-for-Linux/pull/444 some conf files may still exist on the machine, but omsadmin.conf should be removed when --remove is called.

This change enforces that omsadmin.conf is deleted for each remaining workspace with configuration on the machine.

--remove has been tested
--remove then --upgrade has been tested, ensuring that omsadmin.sh -l shows "No Workspace"
Unit and system tests passed
pBuild running now


NarineM:
I am pretty sure you have not validated that the agent ID does not change in upgrade which will be broken.

lagalbra:
Closing until further discussion has taken place

PR #453, opened 2017-04-21
Improve list_workspace messaging

lagalbra:
@Microsoft/omsagent-devs @NarineM 

There are certain states where the omsagent service is registered on the machine, but it is not running. In these cases, list_workspaces should print out something different from when the service is running.

Unit and system tests pass.
Tested with the VM extension, all APIs.
pBuild running now.

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -617,12 +617,16 @@ show_workspace_status()
     local is_primary=$3
     local status='Unknown'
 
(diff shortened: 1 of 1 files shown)


lagalbra:
Once Travis and pBuild finish, I will include this commit in the 1.3.4-15 release branch and start Jenkins to include this for the extension release.

NarineM:
Sounds good. thanks.

lagalbra:
Jenkins building now.

PR #454, opened 2017-04-21
Moved changetracking diagnostic log to be informational

NarineM:
Moved changetracking diagnostic log to be informational and fixed the message. Since this message is logged in error stream, as a result it is shown, in any environment where change tracking is not on, in the console when omsagent is restarted.
@gauravga @vrdmr


gauravga:
approved

PR #455, opened 2017-04-23
Fix Baseline bugs (by Tad Glines)

markwgit:
- fix bug in the distro driver code that resulted in false positives for the CheckServiceEnabled code.
- fix gaps in the rpcbind rules for systems running systemd.

@lagalbra 
@NarineM 
@mgladi

### DIFF
--- a/installer/conf/omsagent.d/oms_audits.xml
+++ b/installer/conf/omsagent.d/oms_audits.xml
@@ -272,7 +272,7 @@
         command="CheckNoPromiscInterfaces" />
     </audit>
     <audit
-      description="All wireless interfaces should be disabled. (disabled)"
+      description="All wireless interfaces should be disabled."
       msid="49"
       cceid="CCE-4276-2"
       severity="Warning"
@@ -386,7 +386,7 @@
         service="mdsd" />
     </audit>
     <audit
-      description="Rsyslog should not accept remote messages. (reject)"
+      description="Rsyslog should not accept remote messages."
       msid="67"
       cceid="CCE-17639-6"
       severity="Important"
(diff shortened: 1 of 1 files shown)


installer/conf/omsagent.d/oms_audits.xml
```diff
@@ -767,9 +767,13 @@
       remediation="Run the command '/usr/local/bin/azsecd remediate -r disable-rpcbind'. This will disable the rpcbind service."
       ruleId="f4a80328-1d67-45ed-b915-274d2e6c699e">
       <check
-        distro="*"
+        distro="Debian|Ubuntu|Oracle|CentOS&lt;7|RedHat&lt;7|SLES=11"
```
lagalbra: How was this list made? Is it supposed to check both distro type and version? If so, we do not only support versions 7 and 11 of CentOS/RedHat/SLES; see full list [here](https://github.com/Microsoft/OMS-Agent-for-Linux/tree/master#supported-linux-operating-systems).

installer/conf/omsagent.d/oms_audits.xml
```diff
@@ -767,9 +767,13 @@
...
         command="CheckServiceDisabled"
         service="rpcbind" />
+      <check
+        distro="CentOS&gt;=7|RedHat&gt;=7|SLES&gt;11"
```
lagalbra: Same distro question here

installer/conf/omsagent.d/oms_audits.xml
markwgit: @taglines 
Tad, could you please answer on questions regarding distro and versions

markwgit:
I only tested that new Baseline binaries  and xml are working on different Linux distributions.
I did not changes source ruby files and not FluentD configuration files and event did not changes files names. I only change the content of Baseline scanner files.
These files just copied by build and installation and also does not in use in UTs.

Also, I am planning to test these changes as part of my testing for new agent version.

lagalbra:
@markwgit All changes should be tested before they are merged into master, not only with a new agent version. Related to this, I believe April 21st was the last check-in date for the agent release, so this change will not be included, unless exceptional circumstances are run by @keikhara .

lagalbra:
@markwgit @taglines Are these changes still relevant?

Issue #434, opened 2017-04-06
Cisco ASA Logs Not Parsing Correctly


mw711:
I'm having an issue using rsyslog and forwarding to OMS agent with the security_event.conf format. For single digit days in the time stamp there is an extra (double) space before the single digit day (5) and the pattern does not match. Example from omsagent.log

> `[warn]: pattern not match: "Apr  5 14:46:54 ASADEVICE : %ASA-6-302021: Teardown ICMP connection for faddr 1.1.1.1/0 gaddr 1.1.1.1/1046 laddr 1.1.1.1/1046"`

If I change the date on the cisco asa to something with a double digit date (10, 11, 12, etc) there is only a single space in between Apr and the day and log parses correctly and is sent to OMS portal. Its tough to see the double space here so i am attaching part of the oms log file where you can see the double space more clearly. 
[example.txt](https://github.com/Microsoft/OMS-Agent-for-Linux/files/904128/example.txt)

lagalbra:
I believe the regex in [security_events.conf](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/conf/omsagent.d/security_events.conf) will have to be improved. This configuration is not currently included in a DSC resource, so a new one may have to be created so as to update this file.

@mgladi Please take a look. Perhaps a more tried-and-true regex for this time format can be used?

mgladi:
The format in the [security_events.conf](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/conf/omsagent.d/security_events.conf) file was updated to support single digit days. 
Unfortunately, as we do not yet have a DSC resource for this file, existing configurations of Cisco/CEF collection will need to be manually updated

PR #458, opened 2017-04-26
Fix for security syslog format to support single digit dates

mgladi:
Fix for issue #434


lagalbra:
Did you explicitly test this against the previously-failing date strings from the customer?
LGTM

mgladi:
Yes I did. In addition all previously known formats were verified

PR #461, opened 2017-04-29
Install some basic dependencies using OS vendor

lagalbra:
@Microsoft/omsagent-devs 

Tested on an Oracle 7 machine (yum), Ubuntu 14 machine (apt-get), SLES 12 machine (zypper)
Unit, system, pBuild tests running now

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -117,6 +117,18 @@ cleanup_and_exit()
     fi
 }
 
+check_program_in_path() {
+    # Parameter: name of program to check
+    # Returns: 0 if program is in path, non-zero if not
+    if [ $# -ne 1 ]; then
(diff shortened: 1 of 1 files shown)


lagalbra:

sugr4:
Sign off, assuming that install and upgrade scenarios were tested with this bundle on suse, debian based and rhel based machines.

installer/bundle/bundle_skel.sh
```diff
@@ -370,6 +383,11 @@ EOF
     $tempFile 1> /dev/null 2> /dev/null
     [ $? -eq 0 ] && hasCtypes=0
     rm $tempFile
+
+    # Check #2: Attempt to run python with the single import command
+    python -c "import ctypes" 1> /dev/null 2> /dev/null
```
sugr4: Maybe use > /dev/null 2>&1

lagalbra: I noticed in the code that both forms are used; I may standardize this in a later review.

PR #465, opened 2017-05-02
Add --enable-opsmgr flag in omsagent bundle to open port 1270

vimish:
@Microsoft/omsagent-devs 
@jeffaco 
@uashok 

In order for omsagent to be able to connect to SCOM, port 1270
nees to be opened. A new flag (--enable-opsmgr) has been added
in omsagent bundle for opening port 1270. If someone intends to
use omsagent with SCOM, he will now have an option to pass the
flag during install/upgrade of the agent. Also the port 1270 will
be opened(if not already open) when scom workspace is created.

Following changes are also done:
1. Added scom workpsace in aggregate commands (start all, stop all etc.)
2. Removed code duplicity in timer based plugins.
3. Removed a warning in scom plugin's unittest.

Following tests are done:
1. PBuild was successful.
2. Successful onboarding on both OMS and SCOM
3. Verified functionality of changed plugins

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -74,6 +74,7 @@ usage()
     echo "  --restart-deps             Reconfigure and restart dependent service(s)."
     echo "  --source-references        Show source code reference hashes."
     echo "  --upgrade                  Upgrade the package in the system."
+    echo "  --enable-opsmgr            Enable port 1270 for usage with opsmgr."
     echo "  --version                  Version of this shell bundle."
     echo "  --version-check            Check versions already installed to see if upgradable."
     echo "  --debug                    use shell debug mode."
@@ -613,6 +614,11 @@ do
             shift 2
             ;;
 
+        --enable-opsmgr)
+            enableOMFlag=true
+            shift 1
+            ;;
+
         -\? | -h | --help)
             usage `basename $0` >&2
             cleanup_and_exit 0
@@ -883,7 +889,8 @@ case "$installMode" in
             # Install SCX (and OMI)
             [ -n "${forceFlag}" ] && FORCE="--force" || FORCE=""
             [ -n "${debugMode}" ] && DEBUG="--debug" || DEBUG=""
-            ./bundles/${SCX_INSTALLER} --install $FORCE $DEBUG
+            [ -n "${enableOMFlag}" ] && ENABLE_OM="--enable-opsmgr" || ENABLE_OM=""
+            ./bundles/${SCX_INSTALLER} --install $FORCE $DEBUG $ENABLE_OM
             TEMP_STATUS=$?
             if [ $TEMP_STATUS -ne 0 ]; then
                 echo "$SCX_INSTALLER package failed to install and exited with status $TEMP_STATUS"
@@ -963,7 +970,8 @@ case "$installMode" in
         # Install SCX (and OMI)
         [ -n "${forceFlag}" ] && FORCE="--force" || FORCE=""
         [ -n "${debugMode}" ] && DEBUG="--debug" || DEBUG=""
-        ./bundles/${SCX_INSTALLER} --upgrade $FORCE $DEBUG
+        [ -n "${enableOMFlag}" ] && ENABLE_OM="--enable-opsmgr" || ENABLE_OM=""
+        ./bundles/${SCX_INSTALLER} --upgrade $FORCE $DEBUG $ENABLE_OM
         TEMP_STATUS=$?
         if [ $TEMP_STATUS -ne 0 ]; then
             echo "$SCX_INSTALLER package failed to upgrade and exited with status $TEMP_STATUS"
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -85,6 +85,7 @@ MONITOR_AGENT_PORT=$DEFAULT_MONITOR_AGENT_PORT
 
 # SCOM variables
 SCX_SSL_CONFIG=/opt/microsoft/scx/bin/tools/scxsslconfig
+OMI_CONF_FILE=/etc/opt/omi/conf/omiserver.conf
 
 usage()
 {
@@ -325,6 +326,34 @@ set_proxy_setting()
     fi
 }
 
+is_scom_port_open()
+{
+    OIFS=$IFS
+# Ignore formatting. It was required for IFS to take new line on both bash and sh
+IFS='
+'
+    for line in `cat $OMI_CONF_FILE | grep httpsport`
+    do
+        echo $line | grep -E '^[ ]*httpsport[ ]*=.*$' > /dev/null 2>&1
+        if [ $? -ne 0 ]; then
+        continue
+        fi
+        IFS=$OIFS
+        # 1st line starting with <optional-spaces>httpsport<optional-spaces>=
+        # Replace = and , with space.
+        # e.g httpsport=0,1270 should become httpsport 0 1270
+        m_line=`echo $line | sed "s/[=,]/ /g"`
+        # Do a exact word match for port 1270.
+        # This ensures cases like 11270 does not match 
+        echo $m_line | grep -w 1270 > /dev/null 2>&1
+        if [ $? -eq 0 ]; then
+            echo "Port 1270 already open"
+            return 0
+        fi
+        return 1
+    done
+}
+
 onboard_scom()
 {
     echo "Onboarding SCOM"
@@ -353,6 +382,16 @@ onboard_scom()
     #Always register SCOM as secondary Workspace
     echo "SCOM Workspace" > $CONF_DIR/.multihoming_marker
     configure_logrotate
+
+    #Open port 1270 if not already open
+    is_scom_port_open
+    if [ $? -eq 1 ]; then
+        echo "Opening port 1270"
+        /opt/omi/bin/omiconfigeditor httpsport -a 1270 < /etc/opt/omi/conf/omiserver.conf > /etc/opt/omi/conf/omiserver.conf_temp
+        mv /etc/opt/omi/conf/omiserver.conf_temp /etc/opt/omi/conf/omiserver.conf
+        # Restart OMI
+        /opt/omi/bin/service_control restart
+    fi
 }
 
 onboard_lad()
@@ -608,6 +647,13 @@ remove_all()
     # Remove LAD workspace
     WORKSPACE_ID="LAD"
     remove_workspace
+
+    # remove scom workspace
+    ls -l $ETC_DIR | grep -w scom > /dev/null 2>&1
+    if [ $? -eq 0 ]; then
+        WORKSPACE_ID="scom"
+        remove_workspace
+    fi
 }
 
 show_workspace_status()
@@ -639,6 +685,16 @@ show_workspace_status()
     else
         echo "Workspace${mh_marker}: ${ws_id}    Status: ${status}"
     fi
+}
+
+list_scom_workspace()
+{
+    ls -1 $ETC_DIR | grep -w scom > /dev/null 2>&1
+    if [ $? -eq 0 ]; then
+        show_workspace_status $ETC_DIR/scom/conf scom 0
+        return 1
+    fi
+    return 0
 }
 
 list_workspaces()
@@ -684,6 +740,11 @@ list_workspaces()
             found_ws=1
             show_workspace_status $ETC_DIR/${ws_id}/conf ${ws_id} 0
         done
+    fi
+    # check scom workspace
+    list_scom_workspace
+    if [ $found_ws -eq 0 ]; then
+        found_ws=$?
     fi
 
     if [ $found_ws -eq 0 ]; then
--- a/installer/scripts/service_control
+++ b/installer/scripts/service_control
@@ -22,6 +22,9 @@ BIN_DIR=/opt/microsoft/omsagent/bin
 
 WORKSPACE_REGEX='^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$'
 
+# Space seperated list of non oms workspaces
+NON_OMS_WS="scom"
+
 setup_variables()
 {
     if [ -z "$1" ]; then
@@ -321,6 +324,18 @@ disable_omsagent_service()
     rm -f $CONF_DIR/.service_registered
 }
 
+check_ws_and_invoke()
+{
+    for ws_id in $NON_OMS_WS
+    do
+        ls -1 $ETC_DIR | grep -w ${ws_id} > /dev/null 2>&1
+        if [ $? -eq 0 ]; then
+            setup_variables ${ws_id}
+            $1
+        fi
+    done
+}
+
 start_all_omsagent()
 {
     for ws_id in `ls -1 $ETC_DIR | grep -E $WORKSPACE_REGEX`
@@ -328,6 +343,8 @@ start_all_omsagent()
        setup_variables ${ws_id}
        start_omsagent
     done
+
+    check_ws_and_invoke start_omsagent
 }
 
 stop_all_omsagent()
@@ -337,6 +354,8 @@ stop_all_omsagent()
        setup_variables ${ws_id}
        stop_omsagent
     done
+
+    check_ws_and_invoke stop_omsagent
 }
 
 restart_all_omsagent()
@@ -346,6 +365,8 @@ restart_all_omsagent()
        setup_variables ${ws_id}
        restart_omsagent
     done
+
+    check_ws_and_invoke restart_omsagent
 }
 
 enable_all_omsagent_services()
@@ -355,6 +376,8 @@ enable_all_omsagent_services()
        setup_variables ${ws_id}
        enable_omsagent_service
     done
+
+    check_ws_and_invoke enable_omsagent_service
 }
 
 disable_all_omsagent_services()
@@ -364,6 +387,8 @@ disable_all_omsagent_services()
        setup_variables ${ws_id}
        disable_omsagent_service
     done
+
+    check_ws_and_invoke disable_omsagent_service
 }
 
 case "$1" in
--- a/source/code/plugins/filter_scom_cor_match.rb
+++ b/source/code/plugins/filter_scom_cor_match.rb
@@ -1,5 +1,7 @@
+require_relative 'scom_common'
+
 module Fluent
-  class SCOMCorrelatedMatchFilter < Filter
+  class SCOMCorrelatedMatchFilter < SCOMTimerFilterPlugin
     #Filter plugin that generates an event when first regex matches and 
     #sceond regex matches before given time interval
     Fluent::Plugin.register_filter('filter_scom_cor_match', self)
@@ -8,25 +10,15 @@ module Fluent
     config_param :regexp1, :string, :default => nil
     desc 'stores regex which needs to match after match for regexp1'
     config_param :regexp2, :string, :default => nil
-    desc 'time interval before which regexp2 needs to match'
-    config_param :time_interval, :integer, :default => 0
-    desc 'event number to be sent to SCOM'
-    config_param :event_id, :string, :default => nil
-    desc 'event description to be sent to SCOM'
-    config_param :event_desc, :string, :default => nil
         
     attr_reader :expression1
     attr_reader :key1
     attr_reader :expression2
     attr_reader :key2
     attr_reader :time_interval
-
+    
     def initialize()
       super
-      require_relative 'scom_common'
-      @exp1_found = false
-      @timer = nil
-      @lock = Mutex.new
     end
     
     def start
@@ -41,8 +33,6 @@ module Fluent
       super
             
       raise ConfigError, "Configuration does not contain 2 expressions" unless @regexp1 and @regexp2
-      raise ConfigError, "Configuration does not have corresponding event ID" unless @event_id
-      raise ConfigError, "Configuration does not have a time interval" unless @time_interval
       @key1, exp1 = @regexp1.split(/ /,2)
       raise ConfigError, "regexp1 does not contain 2 parameters" unless exp1
       @expression1 = Regexp.compile(exp1)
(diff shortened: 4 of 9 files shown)


installer/bundle/bundle_skel.sh
```diff
@@ -881,7 +887,8 @@ case "$installMode" in
             # Install SCX (and OMI)
             [ -n "${forceFlag}" ] && FORCE="--force" || FORCE=""
             [ -n "${debugMode}" ] && DEBUG="--debug" || DEBUG=""
-            ./bundles/${SCX_INSTALLER} --install $FORCE $DEBUG
+            [ -n "${enableOMFlag}" ] && ENABLE_OM="--enable-opsmgr" || ENABLE_OM=""
+            ./bundles/${SCX_INSTALLER} --install $FORCE $DEBUG $ENABLE_OM
```
lagalbra: Can you provide a link to the changes that are in the SCX installer bundle to support this flag?

installer/scripts/omsadmin.sh
```diff
@@ -353,6 +353,15 @@ onboard_scom()
     #Always register SCOM as secondary Workspace
     echo "SCOM Workspace" > $CONF_DIR/.multihoming_marker
     configure_logrotate
+
+    #Open port 1270 if not already open
+    IS_PORT_OPEN=`cat $OMI_CONF_FILE | grep httpsport | grep 1270 | wc -l`
```
lagalbra: Can this search be more exact? According to the [Microsoft docs](https://docs.microsoft.com/en-us/azure/log-analytics/log-analytics-linux-agents), the exact line should be `httpsport=1270`

lagalbra: Also, wc is not a Linux command we've run during installation before. I would rather we simply evaluate whether `cat $OMI_CONF_FILE | grep httpsport | grep 1270` returns 0 or 1.

It may seem like something the customer should have, but I've seen errors recently with our install because the customer didn't have tar installed :)

installer/scripts/omsadmin.sh
```diff
@@ -608,6 +617,12 @@ remove_all()
     # Remove LAD workspace
     WORKSPACE_ID="LAD"
     remove_workspace
+
+    # remove scom workspace
+    if [ `ls -1 $ETC_DIR | grep scom | wc -l` -eq 1 ]; then
```
lagalbra: Same wc concern here. Can we evaluate `ls -1 $ETC_DIR | grep scom` only?

installer/scripts/omsadmin.sh
```diff
@@ -665,6 +680,11 @@ list_workspaces()
                 show_workspace_status $ETC_DIR/${ws_id}/conf ${ws_id} 0
             fi
         done
+        # check scom workspace
+        if [ `ls -1 $ETC_DIR | grep scom | wc -l` -eq 1 ]; then
```
lagalbra: Same wc concern here. Can we evaluate ls -1 $ETC_DIR | grep scom only?

installer/scripts/omsadmin.sh
```diff
@@ -684,6 +704,11 @@ list_workspaces()
             found_ws=1
             show_workspace_status $ETC_DIR/${ws_id}/conf ${ws_id} 0
         done
+        # check scom workspace
+        if [ `ls -1 $ETC_DIR | grep scom | wc -l` -eq 1 ]; then
```
lagalbra: Same wc concern here. Can we evaluate ls -1 $ETC_DIR | grep scom only?

installer/scripts/service_control
```diff
@@ -328,6 +331,14 @@ start_all_omsagent()
        setup_variables ${ws_id}
        start_omsagent
     done
+
+    for ws_id in $NON_OMS_WS
+    do
+        if [ `ls -1 $ETC_DIR | grep ${ws_id} | wc -l` -eq 1 ]; then
```
lagalbra: Same wc concern here. Can we evaluate ls -1 $ETC_DIR | grep scom only?

installer/scripts/service_control
```diff
@@ -337,6 +348,14 @@ stop_all_omsagent()
        setup_variables ${ws_id}
        stop_omsagent
     done
+
+    for ws_id in $NON_OMS_WS
+    do
+        if [ `ls -1 $ETC_DIR | grep ${ws_id} | wc -l` -eq 1 ]; then
```
lagalbra: Same wc concern here. Can we evaluate ls -1 $ETC_DIR | grep scom only?

installer/scripts/service_control
```diff
@@ -346,6 +365,14 @@ restart_all_omsagent()
        setup_variables ${ws_id}
        restart_omsagent
     done
+
+    for ws_id in $NON_OMS_WS
+    do
+        if [ `ls -1 $ETC_DIR | grep ${ws_id} | wc -l` -eq 1 ]; then
```
lagalbra: Same wc concern here. Can we evaluate ls -1 $ETC_DIR | grep scom only?

installer/scripts/service_control
```diff
@@ -355,6 +382,14 @@ enable_all_omsagent_services()
        setup_variables ${ws_id}
        enable_omsagent_service
     done
+
+    for ws_id in $NON_OMS_WS
+    do
+        if [ `ls -1 $ETC_DIR | grep ${ws_id} | wc -l` -eq 1 ]; then
```
lagalbra: Same wc concern here. Can we evaluate ls -1 $ETC_DIR | grep scom only?

installer/scripts/service_control
```diff
@@ -364,6 +399,14 @@ disable_all_omsagent_services()
        setup_variables ${ws_id}
        disable_omsagent_service
     done
+
+    for ws_id in $NON_OMS_WS
+    do
+        if [ `ls -1 $ETC_DIR | grep ${ws_id} | wc -l` -eq 1 ]; then
```
lagalbra: Same wc concern here. Can we evaluate ls -1 $ETC_DIR | grep scom only?

source/code/plugins/filter_scom_cor_match.rb
```diff
@@ -1,5 +1,7 @@
+require_relative 'scom_common'
+
 module Fluent
-  class SCOMCorrelatedMatchFilter < Filter
+  class SCOMCorrelatedMatchFilter < SCOMTimerFilterPlugin
```
lagalbra: It might be better if these plugin changes were separated out into a different PR from the scom service changes, but that's my opinion.

source/code/plugins/filter_scom_cor_match.rb
```diff
@@ -8,25 +10,15 @@ class SCOMCorrelatedMatchFilter < Filter
...
         
     attr_reader :expression1
     attr_reader :key1
     attr_reader :expression2
     attr_reader :key2
     attr_reader :time_interval
-
-    def initialize()
+    
+    def initialize
```
lagalbra: nit: Can you add back the parentheses?

installer/bundle/bundle_skel.sh
jeffaco:
```
commit 056ebea
Author: Vishal Mishra <vimish@microsoft.com>
Date:   Fri Mar 24 00:20:32 2017
```

This was super easy to find with `git log -p installer/bundle/bundle_skel_Linux.sh`

jeffaco: That said, we're going to have to have a stable supported version of SCX published in [scxcore-kits][] for OMS to pick this up.

SCX v1.6.3-212 is pushed to [scxcore-kits][], and that supports it. So we're good!

[scxcore-kits]: https://github.com/Microsoft/scxcore-kits

installer/scripts/omsadmin.sh
```diff
@@ -85,7 +85,7 @@ MONITOR_AGENT_PORT=$DEFAULT_MONITOR_AGENT_PORT
 
 # SCOM variables
 SCX_SSL_CONFIG=/opt/microsoft/scx/bin/tools/scxsslconfig
-
+OMI_CONF_FILE=/etc/opt/omi/conf/omiserver.conf
```
jeffaco: Insert blank line after this for readability please.

installer/scripts/omsadmin.sh
jeffaco: No, not exactly. The typical line here would be `httpsport=0,1270` due to an artifact on how `omiconfigeditor` works. The 0 would be ignored.

This is actually harder than you might think. What if port 11270 is exposed. Looking for 1270 would work, but would actually be wrong. And `omiconfigeditor` isn't going to help you because it only has mechanisms to change a config file, not to help read it and parse it. This is trivial in python, but we don't have that readily available here.

Our existing scripting is wrong in this regard. And it's hard to do in bash, not to mention cross platform (UNIX, which we should get right).  Things to be concerned about:

1. Is httpsport commented out? If so, ignore it.
2. Port 1270 is different from 11270, and we should know that.
3. The line can contain `=1270`, `=0,1270`, or `=0, 1270`, and we should recognize that.

In retrospect, if we wanted to fix this dependably, we should likely modify `omiconfigeditor` to have a query command (like `-q 1270 httpsport`) that would return `true` or `false` based on what you are querying for. It already knows how to deal with commented out entries, and comma-separated entries. It should be easy there.

Either that, or do what Laura said (actually `grep httpsport $OMI_CONF_FILE | grep 1270 > /dev/null 2> /dev/null`), but know that that is broken in certain cases.

But I think `omiconfigeditor` is the "right" way to do this, and can then be used in lots of other places (including UNIX) trivially.

installer/scripts/omsadmin.sh
jeffaco: This is susceptible to the same issues as `httpsport`, above. `egrep` would make this trivial, but I wouldn't care suggest that for OMS, we already have too many problems with programs not being installed. In fact, it wouldn't be a bad idea to see what programs our installation scripting for `omi` and `SCXcore` uses, and at least make sure that Laura is aware of that.

Is there some parser in OMS that we can call that can tell us if the scom workspace exists? Or I suppose you could have a little bash function that enumerated files in `$ETC_DIR`, and returned success if it found the file you're looking for. Bash can do that, and this need not be portable because OMS only runs on Linux.

I guess it all depends on how bullet-proof you want this to be ...

source/code/plugins/filter_scom_cor_match.rb
jeffaco: I don't really speak Ruby, so I'll leave this (and other Ruby changes) to Laura. I don't care much about commits, but if Laura wants it in a separate commit, you should do that (it is her project). I work a lot in mesos these days, and I get to live by their rules! 😃

vimish:
@jeffaco 
@lagalbra 

Please review next iteration.
Following changes are done:
1. Updated the logic to detect if port 1270 is open or not. 
2. Added exact word match to detect SCOM workspace.
3. Addressed other review comments

source/code/plugins/filter_scom_cor_match.rb
vimish: @lagalbra 
If it is fine by you I would like to merge both parts together this time :) . Moving forward I will keep the ruby and non-ruby commits separate.

installer/scripts/omsadmin.sh
```diff
@@ -325,6 +326,35 @@ set_proxy_setting()
...
 
+is_scom_port_open()
+{
+    OIFS=$IFS
+# Ignore formatting. It was required for IFS to take new line on both bash and sh
+IFS='
+'
+    for line in `cat $OMI_CONF_FILE | grep httpsport`
+    do
+        echo $line | grep -E '^#.*$' > /dev/null 2> /dev/null
```
lagalbra: For all the spots where you forward to `/dev/null`, please use this syntax: `> /dev/null 2>&1`. We use a couple different formats for this throughout the code right now, but I'm trying to standardize it.

installer/scripts/omsadmin.sh
```diff
@@ -325,6 +326,35 @@ set_proxy_setting()
...
+        is_comment=$?
+        if [ $is_comment -eq 0 ]; then
+            continue
+        fi
+        # 1st uncommented line is checked for port
+        # Replace = and , with space.
+        # e.g httpsport=0,1270 should become httpsport 0 1270
+        m_line=`echo $line | sed "s/[=,]/ /g"`
+        echo $m_line | grep -w 1270 > /dev/null 2> /dev/null
```
lagalbra: Have you tested that this doesn't match 11270 like Jeff suggested?

installer/scripts/omsadmin.sh
```diff
@@ -353,6 +383,17 @@ onboard_scom()
     #Always register SCOM as secondary Workspace
     echo "SCOM Workspace" > $CONF_DIR/.multihoming_marker
     configure_logrotate
+
+    #Open port 1270 if not already open
+    is_scom_port_open
+    IS_PORT_OPEN=$?
```
lagalbra: Not sure you need this line; why not simply do `if [ $? -eq 1 ]; then`

installer/scripts/omsadmin.sh
```diff
@@ -677,13 +739,20 @@ list_workspaces()
...
     else
         # no default conf folder, check all the potential workspace folders
         for ws_id in `ls -1 $ETC_DIR | grep -E '^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$'`
         do
             found_ws=1
             show_workspace_status $ETC_DIR/${ws_id}/conf ${ws_id} 0
         done
+        # check scom workspace
+        list_scom_workspace
+        found_ws=$?
```
lagalbra: Seem to me like you use this block in all cases: if there is a symlink at conf/, if conf/ exists but isn't a symlink, and the else case. Why not just move this whole block to right before `if [ $found_ws -eq 0 ]; then` ?

source/code/plugins/filter_scom_cor_match.rb
lagalbra: That's fine.

installer/scripts/omsadmin.sh
```diff
@@ -325,6 +326,35 @@ set_proxy_setting()
...
+    do
+        echo $line | grep -E '^#.*$' > /dev/null 2> /dev/null
+        is_comment=$?
+        if [ $is_comment -eq 0 ]; then
+            continue
+        fi
+        # 1st uncommented line is checked for port
+        # Replace = and , with space.
+        # e.g httpsport=0,1270 should become httpsport 0 1270
+        m_line=`echo $line | sed "s/[=,]/ /g"`
```
jeffaco: Well, this is one way to do it. It might have been just as easy to extend `omiconfigeditor`, but that's neither here nor there. But this would work. I was afraid to use `egrep` for this (but did think about that). But then I forgot that `grep -E` does the same thing.

This code is definitely flawed. You'll match on 1270 on a line, even if it's on some other tag, and that's a problem. I would probably change the line at line 339 to continue if we don't begin with `<optional-spaces>httpsport<optional-spaces>=`. If we don't begin with that, then we definitely don't care about this line. Or maybe just see if we begin with `httpsport=`; it would be best to look at [OMI](https://github.com/Microsoft/omi/blob/master/Unix/server/server.c#L774) and see how picky it is about parsing. I'd need to dig deeper into `Conf_Read()` to fully understand that.

Also, did we have a prior dependence on the `sed` program? OMS is super-picky about not picking up dependencies on programs, as some minimal installations of Linux may not include tools that I thought were everywhere (tar, etc). I wouldn't want to inadvertently pick up a dependency on `sed` if we didn't already have it.

That was part of my reason for suggesting changes to `omiconfigeditor`; that would allow us to avoid new dependencies. But I just want something that works and satisfies OMS's needs.

installer/scripts/omsadmin.sh
```diff
@@ -353,6 +383,17 @@ onboard_scom()
     #Always register SCOM as secondary Workspace
     echo "SCOM Workspace" > $CONF_DIR/.multihoming_marker
     configure_logrotate
+
+    #Open port 1270 if not already open
+    is_scom_port_open
+    IS_PORT_OPEN=$?
+    if [ $IS_PORT_OPEN -eq 1 ]; then
+        echo "Opening port 1270"
+        /opt/omi/bin/omiconfigeditor httpsport -a 1270 < /etc/opt/omi/conf/omiserver.conf > /etc/opt/omi/conf/omiserver.conf_temp
```
jeffaco: I'm not 100% sure how `omiconfigeditor` works. Yeah, I know I wrote it, but it's been a while.

Is `omiconfigeditor` smart enough to uncomment a line if it's commented out when you add a parameter to a line? Worth checking, just to be 100% certain this will work.

My concern is something like this in `omiserver.conf`:

```
#httpsport=0
```

If you invoke `omiconfigeditor` with the `-a` qualifier, will it be smart enough to uncomment the line and then add the option?

installer/scripts/omsadmin.sh
```diff
@@ -608,6 +649,13 @@ remove_all()
     # Remove LAD workspace
     WORKSPACE_ID="LAD"
     remove_workspace
+
+    # remove scom workspace
+    ls -l $ETC_DIR | grep -w scom > /dev/null 2> /dev/null
```
jeffaco: Do you really want `-l` on the `ls` command? It feels like that gives you a lot more things to hit on that you may not want to match. Could you get by with just `ls $ETC_DIR` and still have `grep` match it?

installer/scripts/omsadmin.sh
```diff
@@ -641,6 +689,16 @@ show_workspace_status()
     fi
 }
 
+list_scom_workspace()
+{
+    ls -1 $ETC_DIR | grep -w scom > /dev/null 2> /dev/null
```
jeffaco: Ditto to above.

installer/scripts/service_control
```diff
@@ -328,6 +331,15 @@ start_all_omsagent()
        setup_variables ${ws_id}
        start_omsagent
     done
+
+    for ws_id in $NON_OMS_WS
+    do
+        ls -1 $ETC_DIR | grep -w scom > /dev/null 2> /dev/null
```
jeffaco: Ditto to above (for `ls -l`). Also, do you want to match on `scom` here? This looks like a generalized routine, wouldn't you want to match on `${ws_id}` instead of `scom`?

installer/scripts/service_control
```diff
@@ -337,6 +349,15 @@ stop_all_omsagent()
        setup_variables ${ws_id}
        stop_omsagent
     done
+
+    for ws_id in $NON_OMS_WS
+    do
+        ls -1 $ETC_DIR | grep -w scom > /dev/null 2> /dev/null
```
jeffaco: Ditto * 2 (`ls -l`, matching on `scom` vs `${ws_id}`.

installer/scripts/service_control
```diff
@@ -346,6 +367,15 @@ restart_all_omsagent()
        setup_variables ${ws_id}
        restart_omsagent
     done
+
+    for ws_id in $NON_OMS_WS
+    do
+        ls -1 $ETC_DIR | grep -w scom > /dev/null 2> /dev/null
```
jeffaco: Ditto * 2 again, see above. There's a lot of code duplication here, should this be a common subroutine?

installer/scripts/service_control
```diff
@@ -355,6 +385,15 @@ enable_all_omsagent_services()
        setup_variables ${ws_id}
        enable_omsagent_service
     done
+
+    for ws_id in $NON_OMS_WS
+    do
+        ls -1 $ETC_DIR | grep -w scom > /dev/null 2> /dev/null
```
jeffaco: Ditto * 2, I'm thinking this should be a common subroutine, but I'll leave that to you.

installer/scripts/service_control
```diff
@@ -364,6 +403,15 @@ disable_all_omsagent_services()
        setup_variables ${ws_id}
        disable_omsagent_service
     done
+
+    for ws_id in $NON_OMS_WS
+    do
+        ls -1 $ETC_DIR | grep -w scom > /dev/null 2> /dev/null
```
jeffaco: Ditto * 3 (add common subroutine).

installer/scripts/omsadmin.sh
lagalbra: As far as I know, we did not ever specify a dependence on `sed`, but we use it in lots of places. I recently made a change to install sed if it's not already installed; you can see the commit here: https://github.com/Microsoft/OMS-Agent-for-Linux/commit/e93aa8926e2f93912dfe0cc09a8ae3e05e42b8e0

vimish:
@jeffaco 
@lagalbra 

Did following changes:
1. Added check to look for port 1270 on line starting with format <optional-spaces>httpsport<optional-spaces>=
2. Removed suggested code duplication
3. Addressed other comments

@jeffaco I checked if omiconfigeditor can uncomment a line but it seems it cannot. I tried adding port 1270 when httpsport was commented but the port didn't get open.
I agree with you that right place to make all these changes should be in omiconfigeditor. We can look into getting the query feature added into it and when that is ready I will changes the logic here to use omiconfigeditor. For now I will have to keep these changes here for the upcoming SCOM release.

installer/scripts/omsadmin.sh
vimish: I am using ls -1 (digit 1) here that only lists prints the name of files present. Then I grep for word scom to find if som workspace is present or not.

jeffaco:
@vimish Morning, Vishal. I took a look at the config editor, and it sort of supports it. The intent was to allow chaining of `omiconfigeditor` calls to make multiple changes. So, for example, you could run `omiconfigeditor` to uncomment (if it's already uncommented, nothing happens), pipe that to `omiconfigeditor` to add port 1270, and then be along your way.

If we added a query function, it should just ignore the line if it's commented out (we have code to recognize this via a boolean). Then, if it's not commented out, use existing parsing code to look at individual values and set the return code. All in all, I think adding a query function would be under 30 minutes of work, almost certainly. But I suppose after code review, PBUILD, and related overhead, it would be a few hours.

With all that in mind: Do you want to keep this as is, or just add the query function? Note: If we added a query function, this would require a newer version of OMI, and I'm not sure of the OMI release cycle.

How about this:

1. I'll sign off, you can commit this now,
2. Go ahead and submit a change to add a query function to `omiconfigeditor`, that's generally useful anyway,
3. Add a bug or issue to change this code to use query function after OMI ships a new version (that will include the query code). Presumably v1.3 or later.

Sound okay?

installer/scripts/service_control
```diff
@@ -321,13 +324,27 @@ disable_omsagent_service()
...
+
 start_all_omsagent()
 {
     for ws_id in `ls -1 $ETC_DIR | grep -E $WORKSPACE_REGEX`
     do
        setup_variables ${ws_id}
        start_omsagent
     done
+
+    check_ws_and_invoke start_omsagent
```
lagalbra: Have you run a quick test with these cases just to make sure this works as intended:

1. only SCOM workspace
2. only OMS workspace
3. SCOM and OMS workspace

It looks good to me, but verification would be ideal.

vimish: Sure. Will run all tests (start, stop, list etc) before merging.

vimish:
@jeffaco 
I am fine with the proposed approach.
I will merge this PR (after running the sanity test asked by Laura).
I will make the changes in omiconfigeditor to support query function and when it is available with omsagent(after next omi release), I will update the logic here to use omiconfigeditor.

PR #468, opened 2017-05-08
Tests for File Content Change Tracking Feature

gauravga:
@vrdmr 
@Microsoft/omsagent-devs

### DIFF
--- a/source/code/plugins/change_tracking_runner.rb
+++ b/source/code/plugins/change_tracking_runner.rb
@@ -22,7 +22,6 @@ class ChangeTrackingRunner
 	def self.transform_and_wrap()
 		if File.exist?(CHANGE_TRACKING_FILE)
 			@@log.debug ("Found the change tracking inventory file.")
-
 			# Get the parameters ready.
 			time = Time.now
 			force_send_run_interval_hours = 24
@@ -33,67 +32,47 @@ class ChangeTrackingRunner
 			file = File.open(CHANGE_TRACKING_FILE, "rb")
 			xml_string = file.read; nil # To top the output to show up on STDOUT.
 
+			previousSnapshot = ChangeTracking.getHash(CHANGE_TRACKING_STATE_FILE)
+                        previous_inventory_checksum = {}
+                        begin
+                          if !previousSnapshot.nil?
+                            previous_inventory_checksum = JSON.parse(previousSnapshot["PREV_HASH"])
+                          end
+                        rescue 
+			@@log.warn ("Error parsing previous hash file")
+                             previousSnapshot = nil
+                        end
 			#Transform the XML to HashMap
 			transformed_hash_map = ChangeTracking.transform(xml_string, @@log)
-			output = ChangeTracking.wrap(transformed_hash_map, @hostname, time)
-			hash = Digest::SHA256.hexdigest(output.to_json)
+                        current_inventory_checksum = ChangeTracking.computechecksum(transformed_hash_map)
+                        changed_checksum = ChangeTracking.comparechecksum(previous_inventory_checksum, current_inventory_checksum)
+                        transformed_hash_map_with_changes_marked = ChangeTracking.markchangedinventory(changed_checksum, transformed_hash_map)
 
-			previousSnapshot = getHash()
+			output = ChangeTracking.wrap(transformed_hash_map_with_changes_marked, @hostname, time)
+			hash = current_inventory_checksum.to_json
 
 			# If there is a previous hash
 			if !previousSnapshot.nil?
 				# If you need to force send
 				if force_send_run_interval > 0 and 
 					Time.now.to_i - previousSnapshot[LAST_UPLOAD_TIME].to_i > force_send_run_interval
-					setHash(hash, Time.now)
-				# If the content changed.
-				elsif hash != previousSnapshot[PREV_HASH]
-					setHash(hash, Time.now)
+					ChangeTracking.setHash(hash, Time.now,CHANGE_TRACKING_STATE_FILE)
+				elsif !changed_checksum.nil? and !changed_checksum.empty?
+					ChangeTracking.setHash(hash, Time.now, CHANGE_TRACKING_STATE_FILE)
 				else
 					return {}
 				end
 			else # Previous Hash did not exist. Write it
 				# and the return the output.
-				setHash(hash, Time.now)
+				ChangeTracking.setHash(hash, Time.now, CHANGE_TRACKING_STATE_FILE)
 			end
 			return output
 		else
-			@@log.info ("The change tracking inventory file does not exist and runner will be disabled.")
+			@@log.warn ("The ChangeTracking File does not exists. Make sure it is present at the correct")
 			return {}
 		end 
 	end
 
-	def self.getHash()
-		ret = {}
-		if File.exist?(CHANGE_TRACKING_STATE_FILE) # If file exists
-			@@log.debug "Found the file {CHANGE_TRACKING_STATE_FILE}. Fetching the Hash"
-	        File.open(CHANGE_TRACKING_STATE_FILE, "r") do |f| # Open file
-	        	f.each_line do |line|
-	            	line.split(/\r?\n/).reject{ |l|  
-	                	!l.include? "=" }.map { |s|  
-	                    	s.split("=")}.map { |key, value| 
-	                    		ret[key] = value 
-	                    	}
-	        	end
-			end
-	        return ret
-	    else
-	    	@@log.debug "Could not find the file #{CHANGE_TRACKING_STATE_FILE}"
-	        return nil
-	    end
-	end
-
-	def self.setHash(prev_hash, last_upload_time)
-		# File.write('/path/to/file', 'Some glorious content')
-		if File.exist?(CHANGE_TRACKING_STATE_FILE) # If file exists
-			File.open(CHANGE_TRACKING_STATE_FILE, "w") do |f| # Open file
-				f.puts "#{PREV_HASH}=#{prev_hash}"
-				f.puts "#{LAST_UPLOAD_TIME}=#{last_upload_time}"
-			end
-		else
-			File.write(CHANGE_TRACKING_STATE_FILE, "#{PREV_HASH}=#{prev_hash}\n#{LAST_UPLOAD_TIME}=#{last_upload_time}")
-		end
-	end
 end
 
 ret = ChangeTrackingRunner.transform_and_wrap()
--- a/source/code/plugins/changetracking_lib.rb
+++ b/source/code/plugins/changetracking_lib.rb
@@ -8,6 +8,9 @@ require_relative 'oms_common'
 
 class ChangeTracking
 
+    PREV_HASH = "PREV_HASH"
+    LAST_UPLOAD_TIME = "LAST_UPLOAD_TIME"
+
     @@log =  Logger.new(STDERR) #nil
     @@log.formatter = proc do |severity, time, progname, msg|
         "#{severity} #{msg}\n"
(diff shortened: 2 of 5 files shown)


source/code/plugins/changetracking_lib.rb
```diff
@@ -39,6 +42,7 @@ def self.instanceXMLtoHash(instanceXML)
     def self.serviceXMLtoHash(serviceXML)
         serviceHash = instanceXMLtoHash(serviceXML)
         serviceHash["CollectionName"] = serviceHash["Name"]
+        serviceHash["Inventorychecksum"] = Digest::SHA256.hexdigest(serviceHash.to_json)
```
vrdmr: nit: Camel casing (in other places as well.)

lagalbra:
@gauravga Ready for review? If so, @Microsoft/omsagent-devs

source/code/plugins/change_tracking_runner.rb
```diff
@@ -33,67 +32,45 @@ def self.transform_and_wrap()
...
-					setHash(hash, Time.now)
+					ChangeTracking.setHash(hash, Time.now,CHANGE_TRACKING_STATE_FILE)
 				else
-					return {}
+					ChangeTracking.setHash(hash, Time.now, CHANGE_TRACKING_STATE_FILE)
 				end
 			else # Previous Hash did not exist. Write it
 				# and the return the output.
-				setHash(hash, Time.now)
+				ChangeTracking.setHash(hash, Time.now, CHANGE_TRACKING_STATE_FILE)
```
lagalbra: Looks like you can move this `ChangeTracking.setHash(hash, Time.now, CHANGE_TRACKING_STATE_FILE)` call outside the `if !previousSnapshot.nil?` check.

source/code/plugins/change_tracking_runner.rb
```diff
@@ -33,67 +32,45 @@ def self.transform_and_wrap()
...
 			end
 			return output
 		else
-			@@log.info ("The change tracking inventory file does not exist and runner will be disabled.")
+			@@log.warn ("The ChangeTracking File does not exists. Make sure it is present at the correct")
```
lagalbra: Looks like this hasn't been rebased to include Narine's message change: https://github.com/Microsoft/OMS-Agent-for-Linux/commit/f282c340fe8950165103f701cc24835811ea3bd7

source/code/plugins/changetracking_lib.rb
```diff
@@ -136,6 +142,67 @@ def self.transform(inventoryXMLstr, log = nil)
...
+
+        if inventory_hash.has_key?("packages")
+           inventory = inventory_hash["packages"]
+        end
+
+        if inventory_hash.has_key?("services")
+           inventory = inventory_hash["services"]
+        end
+
+        if inventory_hash.has_key?("fileInventories")
```
lagalbra: Seems like these checks could be better expressed as an if/elsif/else block, and reordering to reflect the ones you want used as default.

source/code/plugins/out_oms_changetracking_file.rb
```diff
@@ -0,0 +1,591 @@
+module Fluent
+
+  class OutChangeTrackingFile < BufferedOutput
+
+    Plugin.register_output('out_oms_oms.changetracking.file', self)
```
lagalbra: Please follow plugin naming convention: `out_oms_changetracking_file`

test/code/plugins/out_oms_changetracking_file_systest.rb
```diff
@@ -0,0 +1,179 @@
+require 'fluent/test'
+require_relative ENV['BASE_DIR'] + '/source/ext/fluentd/test/helper'
+require_relative ENV['BASE_DIR'] + '/source/code/plugins/out_oms_changetracking_file'
+require_relative 'omstestlib'
+require_relative 'out_oms_systestbase'
+
+class TestableOutChangeTrackingFile < Fluent::OutChangeTrackingFile
```
vrdmr: Just a thought - Is there a way to add a test (a more functional - E2E test) which uploads the blob content as well?

source/code/plugins/out_oms_changetracking_file.rb
```diff
@@ -0,0 +1,591 @@
...
+      require 'uri'
+      require_relative 'omslog'
+      require_relative 'oms_configuration'
+      require_relative 'oms_common'
+    end
+
+    config_param :omsadmin_conf_path, :string, :default => '/etc/opt/microsoft/omsagent/conf/omsadmin.conf'
+    config_param :cert_path, :string, :default => '/etc/opt/microsoft/omsagent/certs/oms.crt'
+    config_param :key_path, :string, :default => '/etc/opt/microsoft/omsagent/certs/oms.key'
+    config_param :proxy_conf_path, :string, :default => '/etc/opt/microsoft/omsagent/proxy.conf'
```
lagalbra: As I noted before, please check that these paths are all correct in the plugin files for default/non-multi-homing scenarios. You should also consider that these will have to be changed to be multi-homing compatible fairly soon.

source/code/plugins/out_oms_changetracking_file.rb
```diff
@@ -0,0 +1,591 @@
...
+    ####################################################################################################
+
+    # create a HTTP request to PUT blob
+    # parameters:
+    #   uri: URI. blob URI
+    #   msg: string. body of the request
+    #   file_path: string. file path
+    # returns:
+    #   HTTPRequest. blob PUT request
+    def create_blob_put_request(uri, msg, request_id, file_path = nil)
```
lagalbra: Please explain how this method is different from [out_oms_blob create_blob_get_request](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/source/code/plugins/out_oms_blob.rb#L57) and why it is necessary to make this a separate plugin from out_oms_blob. It would be much better if you utilized out_oms_blob entirely or merely its methods.

test/code/plugins/out_oms_changetracking_file_systest.rb
gauravga: I thought of that, it will essentially mean storing the keys, which I did not want to do.
other way would be to integrate secret store, which is a lot more work again. so no, not this time.

source/code/plugins/change_tracking_runner.rb
```diff
@@ -21,8 +21,7 @@ class ChangeTrackingRunner
 
 	def self.transform_and_wrap()
 		if File.exist?(CHANGE_TRACKING_FILE)
-			@@log.debug ("Found the change tracking inventory file.")
-
+                        @@log.debug ("Found the change tracking inventory file.")
```
gauravga: ok, i will fix it.

gauravga: will fix it

source/code/plugins/change_tracking_runner.rb
gauravga: yup, will do

source/code/plugins/change_tracking_runner.rb
gauravga: i did, before this CR. Will rebase before checking in

source/code/plugins/out_oms_changetracking_file.rb
gauravga: there is no plugin naming convention, as i understand. we have plugins with all sorts of name for pluigns.
you have ones with _ and .

the reason this is "out_oms_oms.changetracking.file", becuase it is a generated from Tag "oms.changetracking.file"
to change this will mean changing the tag for entire change tracking system (service, daemons as well)

source/code/plugins/out_oms_changetracking_file.rb
gauravga: i copied the same content as you have in out_oms_blob and out_oms. you will need to tell me what exact change is needed, i dont know what needs to be done for multihoming.
also change tracking does not support multihoming yet, as I understand.

source/code/plugins/out_oms_changetracking_file.rb
gauravga: This feature required complex functionality in the out plugin.
Basically, this output plugin needs to these functions

Create two blob url based on config parameters
extract changed files from records and Upload files to blob
Upload to blob should happen without gets (security reasons)
Do failsafe if one url fails.
Send the record to ODS
Send a notification to ODS on failures.
there is no match rule that can do this. we have scattered functionality in out_oms_blob and out_oms, but I cant call methods in those, due to (3).

its better to create a new match that works only for file change tracking and can be maintained independent of others

gauravga: we discussed this in the other PR as well.

test/code/plugins/out_oms_changetracking_file_systest.rb
gauravga: Instead, I created a testable class, which has mock of append_blob method. so I test the parameters of the method and make sure they as expected. This is the best I can do at this point

source/code/plugins/out_oms_changetracking_file.rb
lagalbra: Can you provide more explanation here? Can you share a link to a conf file which uses this tag? Also, tag and plugin name do not have to be related; data with a certain tag can go through any number of plugins with any number of names.

We do not have a strongly enforced convention, but if you look at the core output plugins, they use the convention I mentioned (see [out_oms_blob.rb](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/source/code/plugins/out_oms_blob.rb)).

source/code/plugins/out_oms_changetracking_file.rb
lagalbra: Code copying is what I what I would like to avoid. I think what would be best is to create a filter plugin to perform any additional processing you want on the data, and then send it to the out_oms_blob output plugin so that you don't have to copy its methods.

Besides code quality, output plugin duplication should be avoided because each output plugin needs to be altered in configuration to be multi-homing compatible. In omsagent.conf, you can see that string replacement is currently used for the `<match oms.blob.**>` block, which is performed at onboarding time. If your plugin is not present at onboarding time (i.e. if it's delivered by DSC), then this gets trickier.

source/code/plugins/out_oms_changetracking_file.rb
lagalbra: In this case, I would advise creating a blob library for these functionalities and abstracting your calls and the current out_oms_blob calls into the library.

I'm currently changing out_oms_blob to include the ODS SupportWriteOnly flag, and I don't want to keep track of two places where the blob code is utilized.

source/code/plugins/out_oms_changetracking_file.rb
gauravga: the conf file using this is generated dynamically. I want to make tags and plugin name related, so i can generate it dynamically? What is your exact opposition here? I can change the name to "out.oms.changetracking.file",  but I cannot do "out_oms_changetracking_file" since this is a bigger change for me, which I dont think is worthwhile.

source/code/plugins/out_oms_changetracking_file.rb
gauravga: you took my statement "i copied the same content as you have in out_oms_blob and out_oms" out of context and generalized it. 

the code in question, is the lines for "config_param", which are a necessary artifact of fluend plugin. every plugin that expect some parameters have to have them and cannot be avoided. so yes there will be "duplication" of these few lines across all plugins.

Multihoming is all togather a different problem, that has noting to do with fluentd plugins. 
you still havent answered my question? 
"What is the change you require, so this is multihoming compatible if it not currently so"?

source/code/plugins/out_oms_changetracking_file.rb
gauravga: In general, i agree with what you are saying, but the functionality we are trying to imprelement is very very specific/peculiar to file content tracking feature, so this is of no use for anyone else therefore it is not in a library. In future, if anyone wants the functionality, the code can be refactored.

Also. this blob put request differs from other, as it eliminated "get" call

PR #470, opened 2017-05-10
Support write-only ODS for BLOB interactions

lagalbra:
@Microsoft/omsagent-devs 

Unit and system tests running now.
pBuild running now.
Tested locally by turning of DSC updates and ensuring success messages were printed in omsagent.log and data was sent to portal.

A PR will be created in DSC repo with new out_oms_blob.rb and bump to nxOMSPlugin version when this is approved and 2.19 has been completed.

### DIFF
--- a/source/code/plugins/out_oms_blob.rb
+++ b/source/code/plugins/out_oms_blob.rb
@@ -49,18 +49,6 @@ module Fluent
     # Methods
     ####################################################################################################
 
-    # create a HTTP request to GET blob
-    # parameters:
-    #   uri: URI. blob URI
-    # returns:
-    #   HTTPRequest. blob GET request
-    def create_blob_get_request(uri)
-      headers = {}
-
-      req = Net::HTTP::Get.new(uri.request_uri, headers)
-      return req
-    end # create_blob_get_request
-
     # create a HTTP request to PUT blob
     # parameters:
     #   uri: URI. blob URI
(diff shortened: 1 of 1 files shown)


lagalbra:

source/code/plugins/out_oms_blob.rb
```diff
@@ -136,16 +125,42 @@ def get_blob_uri(container_type, data_type, custom_data_type, suffix)
...
+    # get the blob SAS URI and committed blocks from ODS
+    # parameters
+    #   container_type: string. ContainerType of the data
+    #   data_type: string. DataTypeId of the data
+    #   custom_data_type: string. CustomDataType of the CustomLog
+    #   suffix: string. Suffix of the blob
+    # returns:
+    #   URI. blob SAS URI
+    #   string[]. a list of committed blocks
+    def get_blob_uri_and_committed_blocks(container_type, data_type, custom_data_type, suffix)
```
haitch: **Size** is also available, it tells you the current size of the blob.

source/code/plugins/out_oms_blob.rb
```diff
@@ -136,16 +125,42 @@ def get_blob_uri(container_type, data_type, custom_data_type, suffix)
...
+      return blob_uri, blocks_committed
+    end # get_blob_uri_and_committed_blocks
 
     # append data to the blob
     # parameters:
     #   uri: URI. blob URI
     #   msgs: string[]. messages
     #   file_path: string. file path
-    def append_blob(uri, msgs, file_path)
+    def append_blob(uri, msgs, file_path, blocks_committed)
```
haitch: nit: I would call this blob_path.

source/code/plugins/out_oms_blob.rb
lagalbra: Is there any reason we'd need to use it though? I can read it from the JSON, but I don't think any of our code relies on that property.

lagalbra:
pBuild passed.

source/code/plugins/out_oms_blob.rb
lagalbra: Which variable are you suggesting a change to? `uri` or `file_path`? `uri` could be considered a BLOB path, but `file_path` refers to a path on the Linux machine.

haitch: nevermind, I don't see we read from this file_path, and thought it's the blob_path.  Then I see it's been used as a blob Metadata.

PR #471, opened 2017-05-11
Fix System tests build break

gauravga:


lagalbra:
Please test this on your machine's system tests running as well as a different machine. If you want, I can run it on mine in the morning.

lagalbra:
Also, it looks like running these tests produces files at `test/code/plugins/InventoryFileHash.hash` and `test/code/plugins/InventoryFileHash1.hash.` Could you clean all of these up at the end of the test?

PR #472, opened 2017-05-11
Update download script and README.md for release 1.3.4-127

lagalbra:
@Microsoft/omsagent-devs


NarineM:
@lagalbra updating readme is fine, but onboard script needs to be updated only if VM extension is published.

lagalbra:
Updated; download script will be updated when VM Extension is updated with this new agent release.

PR #473, opened 2017-05-11
Fix system tests

gauravga:
@lagalbra


lagalbra:
@Microsoft/omsagent-devs 
I am running these tests on my machine to check the hostname correctness, then I will sign-off.

PR #478, opened 2017-05-18
Corrected --version-check behavior with SCX installer

lagalbra:
+ Fix typo in bundle_skel for --version-check (pytyon -> python)
Verified correctness of --version-check and functionality of other APIs.

@Microsoft/omsagent-devs

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -558,25 +558,9 @@ do
             ;;
 
         --version-check)
-            printf '%-15s%-15s%-15s%-15s\n\n' Package Installed Available Install?
-
-            # scx (and omi)
-            ./bundles/$SCX_INSTALLER --version-check
-
-            # OMS agent itself
-            versionInstalled=`getInstalledVersion omsagent`
-            versionAvailable=`getVersionNumber $OMS_PKG omsagent-`
-            if shouldInstall_omsagent; then shouldInstall="Yes"; else shouldInstall="No"; fi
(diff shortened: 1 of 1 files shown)


lagalbra:
Added a small change to correctly set the exit code in case of installMode error

lagalbra:
Here's the output compared to our latest release:

# With these changes
$ sh omsagent-1.3.4-151.universal.x64.sh --version-check
Extracting...
Package        Installed      Available      Install?

omi            1.2.0.42       1.2.0.35.ulinuxNo
scx            1.6.3.212      scx.1.6.3-212  No
omsagent       1.3.4.151      1.3.4.151      No
omsconfig      1.1.1.553      1.1.1.553      No
Shell bundle exiting with code 0


# Latest release without these changes
$ sh omsagent-1.3.4-127.universal.x64.sh --version-check
Package        Installed      Available      Install?

omsagent-1.3.4-127.universal.x64.sh: 569: omsagent-1.3.4-127.universal.x64.sh: ./bundles/scx-1.6.3-212.universal.x64.sh: not found

lagalbra:
Printing out the exit code is part of `cleanup_and_exit`, which is necessary to call to remove the extracted folder. I'd like to keep the printout of the exit code because we now have more precise exit codes indicating failures; unless a user knows to run `echo $?`, they will not see the exact exit code and be able to reference our troubleshooting docs.

If they're reading the printout, they'll probably see a more precise error message anyway, but I've found it very useful.

NarineM:
@lagalbra I agree with Jeff here that printing error code is not usefull for su

PR #479, opened 2017-05-18
Add option to enable SCOM server authentication

vimish:
@Microsoft/omsagent-devs 

SCOM, being an OnPrem server, cannot enable server authentication
by default. To make it work the should should generate a
certificate with their enterprise CA and and replac default
cert on server. Also the CA should be provisioned on the Linux
machines.
If a customer does these, he can enable server authentication
by setting enable_server_auth parameter to true in SCOM's
output plugin.

### DIFF
--- a/source/code/plugins/scom_common.rb
+++ b/source/code/plugins/scom_common.rb
@@ -66,7 +66,13 @@ module SCOM
     def self.create_secure_http(uri)
       http = Net::HTTP.new( uri.host, uri.port )
       http.use_ssl = true
-      http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+      if SCOM::Configuration.enable_server_auth
+        OMS::Log.info_once("Enabling server certificate validation")
+        http.verify_mode = OpenSSL::SSL::VERIFY_PEER
+      else
+        OMS::Log.info_once("Disabling server certificate validation")
+        http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+      end # if
       http.open_timeout = 30
       return http
     end
--- a/source/code/plugins/out_scom.rb
+++ b/source/code/plugins/out_scom.rb
@@ -14,6 +14,9 @@ module Fluent
       require_relative 'scom_common'
     end
 
+    desc 'Set this parameter to true to validate server certificate'
+    config_param :enable_server_auth, :bool, :default => false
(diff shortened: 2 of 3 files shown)


source/code/plugins/scom_common.rb
```diff
@@ -66,7 +66,13 @@ def self.parse_json_record_encoding(record)
     def self.create_secure_http(uri)
       http = Net::HTTP.new( uri.host, uri.port )
       http.use_ssl = true
-      http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+      if SCOM::Configuration.enable_server_auth
+        OMS::Log.info_once("Enabling server certificate validation")
+        http.verify_mode = OpenSSL::SSL::VERIFY_PEER
+      else
+        OMS::Log.info_once("Disabling server certificate validation")
+        http.verify_mode = OpenSSL::SSL::VERIFY_NONE
```
lagalbra: I would recommend adding documentation that it is strongly recommended that customers do setup and enable server cert authentication. Using VERIFY_NONE is not secure at all.

source/code/plugins/scom_common.rb
```diff
@@ -66,7 +66,13 @@ def self.parse_json_record_encoding(record)
     def self.create_secure_http(uri)
       http = Net::HTTP.new( uri.host, uri.port )
       http.use_ssl = true
-      http.verify_mode = OpenSSL::SSL::VERIFY_NONE
+      if SCOM::Configuration.enable_server_auth
+        OMS::Log.info_once("Enabling server certificate validation")
+        http.verify_mode = OpenSSL::SSL::VERIFY_PEER
+      else
+        OMS::Log.info_once("Disabling server certificate validation")
```
lagalbra: Out of curiosity, have you watched what happens if the using switches their enable_server_auth parameter more than once?

For example, if they have the default, and run for a while;
then they set it to true and restart OMSAgent to take effect;
then they set it back to false and restart OMSAgent to take effect....

Does it log three times, which includes logging "Disabling server certificate validation" twice?

vimish: @lagalbra 
Yes it does log three times. Since an in memory hash is maintained to check whether the message is already logged or not, it gets reset on every OMSAgent restart.

PR #480, opened 2017-05-18
diag: do not emit diagnostic logging items if omsagent version is old

chparimi:
Do not support diagnostic logging if omsagent version is older than
the version 1.3.4-127.

### DIFF
--- a/source/code/plugins/oms_diag_lib.rb
+++ b/source/code/plugins/oms_diag_lib.rb
@@ -24,7 +24,90 @@ module OMS
     # Record constant values
     RECORD_MGID_VALUE = '{00000000-0000-0000-0000-000000000002}'
 
+    # Diagnostic logging support minimum version
+    DIAG_MIN_VERSION = "1.3.4-127"
+
     class << self
+
+      @@DiagSupported = nil
+      @@InstallInfoPath = "/etc/opt/microsoft/omsagent/sysconf/installinfo.txt"
+
+      # Method to check if diagnostic logging is supported
+      #
+      # Description:
+      # This would tell if current omsagent version supports
+      # diagnostic logging
+      #
+      # NOTE: If this returns false then logs will be rejected silently
+      def IsDiagSupported()
+        return @@DiagSupported unless @@DiagSupported.nil?
+
+        begin
+          # Read installinfo.txt
+          versionline = IO.readlines(@@InstallInfoPath)[0]
+
+          # Extract version number
+          versionnum = versionline.split()[0]
+
+          # Extract major and minor parts of version number
+          cur_major, cur_minor, cur_patch, cur_build = GetVersionParts(versionnum)
+
+          # Check validity of major and minor parts
+          if cur_major.nil? or
+              cur_minor.nil? or
+              cur_patch.nil? or
+              cur_build.nil?
+            @@DiagSupported = false
+            return @@DiagSupported
+          end
+
+          # Compare version number
+          tar_major, tar_minor, tar_patch, tar_build = GetVersionParts(DIAG_MIN_VERSION)
+
+          # Compare major parts
+          if @@DiagSupported.nil?
+            if cur_major.to_i > tar_major.to_i
+              @@DiagSupported = true
+            elsif cur_major.to_i < tar_major.to_i
+              @@DiagSupported = false
+            end
+          end
+
+          # Compare minor parts
+          if @@DiagSupported.nil?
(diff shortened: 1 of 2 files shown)


chparimi:
@Microsoft/omsagent-devs

source/code/plugins/oms_diag_lib.rb
```diff
@@ -126,6 +198,15 @@ def GetCurrentFormattedTimeForDiagLogs()
           Time.now.utc.strftime("%Y-%m-%dT%H:%M:%S.%6NZ")
       end
 
+      # Method used to get major minor version parts as array of omsagent version
+      def GetVersionAsArray(versionStr)
```
lagalbra: I think this should be a little more explicit as major.minor.patch-build; see our [bundle code](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/bundle/bundle_skel.sh#L133) for how we compare version numbers there.

source/code/plugins/oms_diag_lib.rb
```diff
@@ -24,8 +24,78 @@ class Diag
...
+
+        begin
+          # Read installinfo.txt
+          versionline = IO.readlines(@@InstallInfoPath)[0]
+
+          # Extract version number
+          versionnum = versionline.split()[0]
+
+          # Extract major and minor parts of version number
+          major, minor = GetVersionAsArray(versionnum)
```
lagalbra: See my other comment below; I think this should be return as major, minor, patch, build.

source/code/plugins/oms_diag_lib.rb
```diff
@@ -24,8 +24,78 @@ class Diag
...
+             !minor.is_a?(String)
+            @@IsDiagSupported = false
+            return @@IsDiagSupported
+          end
+
+          # Compare version number
+          min_major, min_minor = GetVersionAsArray(DIAG_MIN_VERSION)
+
+          # Compare major parts
+          for x in 0..2
```
lagalbra: Once you make the change to consider all 4 parts separately, make sure the logic procedure matches the logic in bundle_skel [here](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/bundle/bundle_skel.sh#L159).

source/code/plugins/oms_diag_lib.rb
```diff
@@ -24,8 +24,78 @@ class Diag
...
+      @@InstallInfoPath = "/etc/opt/microsoft/omsagent/sysconf/installinfo.txt"
+
+      # Method to check if diagnostic logging is supported
+      #
+      # Description:
+      # This would tell if current omsagent version supports
+      # diagnostic logging
+      #
+      # NOTE: If this returns false then logs will be rejected silently
+      def IsDiagEnabled()
```
lagalbra: The method name is a little unclear; why didn't you use IsDiagSupported()?

test/code/plugins/oms_diag_lib_test.rb
```diff
@@ -240,4 +254,95 @@ def test_case_record_proc_01
...
+
+        # Checking that diag is enabled
+        assert_equal(true,
+                     OMS::Diag.IsDiagEnabled(),
+                     "Diagnostic logging should be supported for when version is min version")
+    end
+
+    # Checking diagnostic support with higher omsagent version
+    def test_case_version_check_03
+        versions = ["1.3.4-128", "1.3.5-127", "1.4.4-127", "2.3.4-127"]
```
lagalbra: I would also suggest adding cases like 1.4.0-1, which should evaluate as a higher version than 1.3.4-127, even though the patch and build numbers are lower.

source/code/plugins/oms_diag_lib.rb
chparimi: Ok will do that.

source/code/plugins/oms_diag_lib.rb
chparimi: Checked the logic. It is the same. Just that I would have to make it major, minor patch and build. Will update right away.

source/code/plugins/oms_diag_lib.rb
chparimi: Doing that.

test/code/plugins/oms_diag_lib_test.rb
chparimi: Sure will add that.

chparimi:
@lagalbra Can you please approve this PR?

PR #481, opened 2017-05-18
Adjust to new auoms input format

taglines:


lagalbra:
@Microsoft/omsagent-devs

source/code/plugins/in_auoms.rb
```diff
@@ -0,0 +1,151 @@
...
+require 'cool.io'
+require 'yajl'
+
+require 'fluent/input'
+require 'fluent/event'
+
+module Fluent
+  # obsolete
+  class AuOMSInput < Input
+    Plugin.register_input('auoms', self)
```
lagalbra: Please follow our naming convention: in_auoms

source/code/plugins/in_auoms.rb
```diff
@@ -0,0 +1,151 @@
+#
```
lagalbra: Where possible, unit tests should be added for this plugin.

source/code/plugins/in_auoms.rb
```diff
@@ -0,0 +1,151 @@
...
+require 'socket'
+
+require 'cool.io'
+require 'yajl'
+
+require 'fluent/input'
+require 'fluent/event'
+
+module Fluent
+  # obsolete
```
lagalbra: What's obsolete?

source/code/plugins/in_auoms.rb
```diff
@@ -0,0 +1,151 @@
...
+    desc 'Tag of the output events.'
+    config_param :tag, :string, default: nil
+    desc 'The path to your Unix Domain Socket.'
+    config_param :path, :string, default: nil
+    desc 'The backlog of Unix Domain Socket.'
+    config_param :backlog, :integer, default: nil
+
+    def initialize
+      require 'socket'
+      require 'yajl'
```
lagalbra: You already require yajl and socket at the top of the file; why do you need them here?

source/code/plugins/in_auoms.rb
```diff
@@ -0,0 +1,151 @@
...
+      @lsock.close
+      @thread.join
+    end
+
+    def listen
+      if File.exist?(@path)
+        File.unlink(@path)
+      end
+      FileUtils.mkdir_p File.dirname(@path)
+      log.info "listening fluent socket on #{@path}"
```
lagalbra: Use $log.info "message"

lagalbra: nit: change to "Listening to FluentD socket on #{@path}"

source/code/plugins/in_auoms.rb
```diff
@@ -0,0 +1,151 @@
...
+      s = Coolio::UNIXServer.new(@path, Handler, log, method(:on_message))
+      s.listen(@backlog) unless @backlog.nil?
+      s
+    end
+
+    def run
+      @loop.run(@blocking_timeout)
```
lagalbra: You're missing a `begin` keyword here.

source/code/plugins/in_auoms.rb
```diff
@@ -0,0 +1,151 @@
...
+    rescue
+      log.error "unexpected error", error: $!.to_s
```
lagalbra: For all logging, use $log.info and $log.error

source/code/plugins/in_auoms.rb
```diff
@@ -0,0 +1,151 @@
...
+        end
+        @on_message = on_message
+        @log = log
+        @log.trace {
+          remote_port, remote_addr = *Socket.unpack_sockaddr_in(@_io.getpeername) rescue nil
+          "accepted fluent socket from '#{remote_addr}:#{remote_port}': object_id=#{self.object_id}"
+        }
+      end
+
+      def on_connect
```
lagalbra: Why is this empty method here?

source/code/plugins/in_auoms.rb
taglines: Done

source/code/plugins/in_auoms.rb
taglines: Removed. Was cruft from in_stream.rb

source/code/plugins/in_auoms.rb
taglines: This is the correct naming convention. ALL in_<name>.rb input plugins use: Plugin.register_input('<name>', self)

source/code/plugins/in_auoms.rb
taglines: Removed.

source/code/plugins/in_auoms.rb
taglines: Changed to @log

source/code/plugins/in_auoms.rb
taglines: This is valid. If the "rescue" is at the same level as "def' then begin isn't required.

source/code/plugins/in_auoms.rb
taglines: Changed to @log

source/code/plugins/in_auoms.rb
taglines: Not empty anymore.

source/code/plugins/in_auoms.rb
lagalbra: Huh, interesting!

lagalbra:
Please confirm that you have run pBuild with these changes, as well as onboarding the OMSAgent with these changes and verifying the impacted scenarios.

I'm unfamiliar with the purpose of these plugins/what they log, so it might be good from that perspective to get a reviewer from your team.

taglines:
Yes, I've run pBuild. Yes, I've onboarded (many times), and tested upgrade scenario.

PR #483, opened 2017-05-19
Ensure that plugin directory is owned by omsagent

lagalbra:
@Microsoft/omsagent-devs @taglines 

Verified that directory permissions are applied in upgrade scenario and data flows correctly, plugin files are copied correctly.

pBuild running now.


lagalbra:
pBuild passed. Rebased against master.

PR #488, opened 2017-05-31
cleanup duplicate python prechecks and add error codes for curl and gpg

NarineM:
@KrisBash , @Microsoft/omsagent-devs

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -61,6 +61,8 @@ UNSUPPORTED_OPENSSL=60
 INSTALL_PYTHON_CTYPES=61
 INSTALL_TAR=62
 INSTALL_SED=63
+INSTALL_CURL=64
+INSTALL_GPG=65
 
 usage()
 {
@@ -369,14 +371,31 @@ compare_install_type()
 
 python_ctypes_installed() {
     # Check for Python ctypes library (required for omsconfig)
-    hasCtypes=1
-    echo "Checking for ctypes python module ..."
+    hasCtypes=1    
 	
     # Attempt to run python with the single import command
     python -c "import ctypes" 1> /dev/null 2> /dev/null
     [ $? -eq 0 ] && hasCtypes=0
 
     return $hasCtypes
+}
+
+curl_installed() {
+    # Check for curl (required for omsconfig)
+    preReqCurl=1
+    which curl 1> /dev/null 2> /dev/null
+    [ $? -eq 0 ] && preReqCurl=0
+
+    return $preReqCurl
+}
+
+gpg_installed() {
+    # Check for gpg (required for omsconfig)
+    preReqGpg=1
+    which gpg 1> /dev/null 2> /dev/null
+    [ $? -eq 0 ] && preReqGpg=0
+
+    return $preReqGpg
 }
 
 getInstalledVersion()
@@ -407,13 +426,17 @@ shouldInstall_omsagent()
 
 shouldInstall_omsconfig()
 {
-    # Package omsconfig will never install without Python ctypes ...
+    # Package omsconfig will never install without Python ctypes and curl ...
     if python_ctypes_installed 1> /dev/null 2> /dev/null; then
-        local versionInstalled=`getInstalledVersion omsconfig`
-        [ "$versionInstalled" = "None" ] && return 0
-        local versionAvailable=`getVersionNumber $DSC_PKG omsconfig-`
+        if curl_installed 1> /dev/null 2> /dev/null; then
+            local versionInstalled=`getInstalledVersion omsconfig`
+            [ "$versionInstalled" = "None" ] && return 0
+            local versionAvailable=`getVersionNumber $DSC_PKG omsconfig-`
 
-        check_version_installable $versionInstalled $versionAvailable
+            check_version_installable $versionInstalled $versionAvailable
+        else
+            return 1
+        fi
     else
         return 1
     fi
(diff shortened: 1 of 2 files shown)


installer/bundle/bundle_skel.sh
```diff
@@ -805,6 +829,23 @@ if [ "$installMode" = "I" -o "$installMode" = "U" ]; then
...
+        if [ -z "${forceFlag}" ]; then
+            echo "Error: curl is not installed, installation cannot continue."
+            echo "You can run this shell bundle with --force; in this case, we will install omsagent," >&2
+            echo "but omsconfig (DSC configuration) will not be available and will need to be re-installed." >&2
+            cleanup_and_exit $INSTALL_CURL
+        else
+            echo "curl is not installed,, please install curl and re-install omsconfig (DSC configuration) later."
+            echo "Installation will continue without installing omsconfig."
+        fi
+	fi
```
lagalbra: Indentation seems inconsistent here; please check that all indents are 4 spaces.

installer/bundle/bundle_skel.sh
```diff
@@ -838,9 +879,8 @@ if [ -n "${checkVersionAndCleanUp}" ]; then
 
     # omsconfig
     versionInstalled=`getInstalledVersion omsconfig`
-    versionAvailable=`getVersionNumber $DSC_PKG omsconfig-`
-    if ! python_ctypes_installed > /dev/null 2>&1; then ctypes_text=" (No ctypes)"; fi
-    if shouldInstall_omsconfig; then shouldInstall="Yes"; else shouldInstall="No${ctypes_text}"; fi
+    versionAvailable=`getVersionNumber $DSC_PKG omsconfig-`    
+    if shouldInstall_omsconfig; then shouldInstall="Yes"; else shouldInstall="No (No ctypes)"; fi
```
lagalbra: If `shouldInstall_omsconfig` fails, it's not necessarily because of ctypes. I think we should remove the reason text and just print "Yes" or "No"; otherwise we could make a helper function to check ctypes, curl, and gpg and return a reason text from that.

installer/bundle/bundle_skel.sh
```diff
@@ -871,10 +911,8 @@ case "$installMode" in
 
             pkg_add_list $OMS_PKG omsagent
 
-            python_ctypes_installed
-            if [ $? -eq 0 ]; then
-                pkg_add_list $DSC_PKG omsconfig
-            fi
+            shouldInstall_omsconfig
```
lagalbra: The update path uses the result of this line in `pkg_upd_list $DSC_PKG omsconfig $?`, but the `pkg_add_list` subroutine doesn't have a parameter like that. I believe we have to keep the check if the return type is 0 before we call `pkg_add_list`:

```
shouldInstall_omsconfig
if [ $? -eq 0 ]; then
    pkg_add_list $DSC_PKG omsconfig
fi
```

installer/bundle/bundle_skel.sh
```diff
@@ -805,6 +829,23 @@ if [ "$installMode" = "I" -o "$installMode" = "U" ]; then
         echo "sed was not installed, installation cannot continue. Please install sed."
         cleanup_and_exit $INSTALL_SED
     fi
+    curl_installed
+    if [ $? -ne 0 ]; then
+        if [ -z "${forceFlag}" ]; then
+            echo "Error: curl is not installed, installation cannot continue."
```
NarineM: It is actually not needed at all, echo command would not throw an error. 
It was probably copy paste for the python, but we never redirect the output from echo command

installer/bundle/bundle_skel.sh
```diff
@@ -379,6 +381,24 @@ python_ctypes_installed() {
...
+curl_installed() {
+    # Check for curl (required for omsconfig)
+    preReqCurl=1
+    which curl 1> /dev/null 2> /dev/null
+    [ $? -eq 0 ] && preReqCurl=0
+
+    return $preReqCurl
+}
+
+gpg_installed() {
```
NarineM: documented dsc requirements are out of scope for omsagent and unnecessary would require keeping these docs in sync. there is a separate repo for each of the dependencies.  So I would think that it should be doc'ed in corresponding repo.

installer/bundle/bundle_skel.sh
```diff
@@ -61,6 +61,8 @@ UNSUPPORTED_OPENSSL=60
 INSTALL_PYTHON_CTYPES=61
 INSTALL_TAR=62
 INSTALL_SED=63
+INSTALL_CURL=64
+INSTALL_GPG=65
```
NarineM: I am not sure why it is required to duplicate these docs in two different GitHub repos instead of having them linked? I will see if I can have that also fixed today. Most likely a separate PR since it belongs to extension repro.

installer/bundle/bundle_skel.sh
NarineM: I left the original verbose message as it was, but I agree we can just print "Yes" or "No" here.

installer/bundle/bundle_skel.sh
NarineM: Added new bundle error codes in omsagent repo in this PR. I am going to leave it to you to remove duplicated codes from extension repo since you originally created the table. Thanks.

NarineM:
Thanks for the review and brining the azure doc to attention :)

installer/bundle/bundle_skel.sh
```diff
@@ -369,8 +371,7 @@ compare_install_type()
 
 python_ctypes_installed() {
     # Check for Python ctypes library (required for omsconfig)
-    hasCtypes=1
-    echo "Checking for ctypes python module ..."
```
lagalbra: Why did you remove this message? It's not a  critical thing to print; I'm just curious.

installer/bundle/bundle_skel.sh
lagalbra: The reason the python message has it is because `>&2` redirects to stderr. It's a small thing considering we print lots of warning/error messages to stdout instead. You've made all the lines consistent now, so it's not a big deal.

installer/bundle/bundle_skel.sh
```diff
@@ -871,10 +910,7 @@ case "$installMode" in
 
             pkg_add_list $OMS_PKG omsagent
 
-            python_ctypes_installed
-            if [ $? -eq 0 ]; then
-                pkg_add_list $DSC_PKG omsconfig
-            fi
+            if shouldInstall_omsconfig; then pkg_add_list $DSC_PKG omsconfig; fi
```
lagalbra: We should standardize the single-line if statement vs. mutli-line if block throughout the code; as it is, this looks fine.

lagalbra:
Remember to rebase against master and squash commits before merging!

PR #489, opened 2017-06-06
Send additional headers to ODS for BLOB interactions

lagalbra:
Remove SAS token from Blob URL before notifying ODS

@Microsoft/omsagent-devs @haitch

### DIFF
--- a/source/code/plugins/out_oms_blob.rb
+++ b/source/code/plugins/out_oms_blob.rb
@@ -85,6 +85,9 @@ module Fluent
       headers["Content-Type"] = "application/octet-stream"
       headers["Content-Length"] = msg.bytesize.to_s
 
+      # If the request version is 2011-08-18 or newer, the ETag value will be returned
+      headers[OMS::CaseSensitiveString.new("x-ms-version")] = "2011-08-18"
+
       req = Net::HTTP::Put.new(uri.request_uri, headers)
       req.body = msg
       return req
@@ -151,8 +154,13 @@ module Fluent
       else
         blocks_committed = []
       end
+      if blob_json.has_key?("Size")
+        blob_size = blob_json["Size"]
+      else
+        blob_size = 0
+      end
 
-      return blob_uri, blocks_committed
+      return blob_uri, blocks_committed, blob_size
     end # get_blob_uri_and_committed_blocks
 
     # append data to the blob
@@ -184,8 +192,8 @@ module Fluent
       end
 
       # commit blocks
-      commit_blocks(uri, blocks_committed, blocks_uncommitted, file_path)
-      return dataSize
+      etag = commit_blocks(uri, blocks_committed, blocks_uncommitted, file_path)
+      return dataSize, etag
     end # append_blob
 
     # upload one block to the blob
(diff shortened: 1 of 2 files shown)


lagalbra:
Unit and system tests pass. pBuild running now.

source/code/plugins/out_oms_blob.rb
```diff
@@ -222,24 +233,50 @@ def commit_blocks(uri, blocks_committed, blocks_uncommitted, file_path)
...
         "DataType" => "BLOB_UPLOAD_NOTIFICATION",
         "IPName" => "",
         "DataItems" => [
           {
-            "BlobUrl" => uri.to_s,
-            "OriginalDataTypeId" => data_type_id
+            "BlobUrl" => blob_url,
+            "OriginalDataTypeId" => data_type_id,
+            "StartOffset" => offset_blob_size,
+            "FileSize" => sent_size,
```
haitch: this FileSize should be full blobSize, in this case, it should be offset_blob_size + sent_size.

source/code/plugins/out_oms_blob.rb
```diff
@@ -222,24 +233,50 @@ def commit_blocks(uri, blocks_committed, blocks_uncommitted, file_path)
...
+    #   sent_size: int. Amount of data we appended to the BLOB
+    #   etag: string. ETag from the BLOB for this data
+    def notify_blob_upload_complete(uri, data_type, custom_data_type, offset_blob_size, sent_size, etag)
       data_type_id = data_type
       if !custom_data_type.nil?
         data_type_id = "#{data_type}.#{custom_data_type}"
       end
 
+      # Remove SAS token from the URL
+      blob_url = uri.to_s.split("?")[0]
```
haitch: If uri is an strong type object, not sure if you have something similar to this Uri.GetLeftPartial [https://msdn.microsoft.com/en-us/library/system.uri.getleftpart(v=vs.110).aspx].

source/code/plugins/out_oms_blob.rb
```diff
@@ -54,9 +54,10 @@ def shutdown
     #   uri: URI. blob URI
     #   msg: string. body of the request
     #   file_path: string. file path
+    #   get_etag: bool. True if we should request the ETag
     # returns:
     #   HTTPRequest. blob PUT request
-    def create_blob_put_request(uri, msg, request_id, file_path = nil)
+    def create_blob_put_request(uri, msg, request_id, file_path = nil, get_etag = false)
```
lagalbra: Will change.

source/code/plugins/out_oms_blob.rb
lagalbra: Have found a solution here: https://stackoverflow.com/questions/10410523/removing-a-part-of-a-url-with-ruby

Will change to use `uri.fragment = uri.query = nil`

source/code/plugins/out_oms_blob.rb
lagalbra: Will change.

lagalbra:
@haitch Please review again.

lagalbra:
@Microsoft/omsagent-devs Please review. This is has been verified locally.

PR #492, opened 2017-06-09
Explictly create new directories as 755

lagalbra:
Before multi-homing, these directories were created as 755
See code at https://github.com/Microsoft/OMS-Agent-for-Linux/blob/467e845719215e2d4e0e24b9bec95eac4e999e29/installer/datafiles/base_omsagent.data

Restrict file permissions for in_dsc_monitor cache file

@Microsoft/omsagent-devs @FayzaSultan

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -858,7 +858,10 @@ create_workspace_directories()
 make_dir()
 {
     if [ ! -d $1 ]; then
-        mkdir $1
+        mkdir -m 755 $1
+    else
+        echo "Directory $1 already exists."
+        chmod 755 $1
     fi
 
     chown_omsagent $1
--- a/source/code/plugins/in_dsc_monitor.rb
+++ b/source/code/plugins/in_dsc_monitor.rb
@@ -1,6 +1,7 @@
 require 'fluent/input'
 require 'fluent/config/error'
 require 'yaml/store'
+require 'fileutils'
 
 module Fluent
   class DscMonitoringInput < Input
@@ -24,7 +25,12 @@ module Fluent
       @finished_check_status = false
       @thread_check_install = Thread.new(&method(:run_check_install))
       @thread_check_status = Thread.new(&method(:run_check_status))
-      @dsc_cache = YAML::Store.new(dsc_cache_file) #Creates the file if it doesn't exist; otherwise, the existing file will be read.
+      if !File.exists?(dsc_cache_file)
+        File.new(dsc_cache_file, "w", 0644)
+      else
+        File.chmod(0644, dsc_cache_file)
+      end
+      @dsc_cache = YAML::Store.new(dsc_cache_file) # The existing file will be read.
     end
 
     def check_install


lagalbra:
Unit and system tests passed.
Verified on a client machine.
pBuild running now.

lagalbra:
pBuild passed.

source/code/plugins/in_dsc_monitor.rb
```diff
@@ -85,6 +86,7 @@ def run_check_status
           @dsc_cache[:status] = dsc_status
           @dsc_cache.commit
         }
+        File.chmod(0644, dsc_cache_file)
```
FayzaSultan: Why don't you do chmod right after creating the file in line 28.  why here in check_status ? By doing it here there might be a duration when we create a file with ww and actually update the permissions later on.

lagalbra: I tested it that way, but the file was not actually created there. I've added it here because the file is guaranteed to exist there, and it is after the YAML Store class does whatever they do to edit that file.

installer/scripts/omsadmin.sh
```diff
@@ -862,6 +862,7 @@ make_dir()
     fi
 
     chown_omsagent $1
+    chmod 755 $1
```
mthompson83: Is this chmod expected to open the permissions, granting access to a wider audience, or closing access (such as removing world write access)? If it is the later, this means there **is** a window of time where unauthorized actors have access they should not have.

Recommendations:

- Look at setting the process umask to something narrow, such as 0027.
- Check that permission granted at file creation time are no more open then they must be, such as 0660.

source/code/plugins/in_dsc_monitor.rb
mthompson83: I'm surprised the YMAL::Store and its superclass don't seem to support providing the initial file permissions for when it needs to create the file.

Recommendation: Don't let the YAML::Store.new create the file. Always make sure it exists and is created with the minimal correct permissions **before** calling new. This ensure there is never a window time between creating the file and setting the permissions where it is too open.

installer/scripts/omsadmin.sh
lagalbra: This chmod is expected to restrict permissions, to ensure that these folders do not have world-write access.

This change is meant to correct the permissions to how they were set prior to multi-homing changes: see this line in the old code: https://github.com/Microsoft/OMS-Agent-for-Linux/blob/f06b438b51665d4a4426f970af12c677f305a8ea/installer/datafiles/base_omsagent.data#L155

source/code/plugins/in_dsc_monitor.rb
lagalbra: My worry is with that approach is YAML::Store changing the permissions when it runs over the file.

NarineM:
What happens to the files that sit under folder $1 that is being updated with permission 755? Can that path fail and how to handle that?

NarineM:
Is this code review meant to address also files under $1 folder or just the folder permission? I wonder if there is any security risk or can this possibly break components who do not expect 755?

lagalbra:
No, this PR is not expected to control the file permissions inside directories. The creators of those files should set the correct permissions.
Given that pre-multi-homing code set these directories to 755 at creation time, no components should expect anything other than 755, which omsagent will always have rights to.

NarineM:
I see, sounds good. Basically this is to fix the regression after multi homing was introduced. I sign off.

PR #496, opened 2017-06-13
install curl underhood if not present

NarineM:
Attempt to install curl if it is not present. This is mainly targeting customers in azure who use debian images which currently are not prep'ed with curl.
@Microsoft/omsagent-devs

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -830,13 +830,18 @@ if [ "$installMode" = "I" -o "$installMode" = "U" ]; then
     fi
     curl_installed
     if [ $? -ne 0 ]; then
+        echo "curl was not found, attempting to install curl..."
+        install_extra_package curl
+    fi
+    check_program_in_path curl
+    if [ $? -ne 0 ]; then
         if [ -z "${forceFlag}" ]; then
-            echo "Error: curl is not installed, installation cannot continue."
+            echo "Error: curl was not installed, installation cannot continue."
             echo "You can run this shell bundle with --force; in this case, we will install omsagent,"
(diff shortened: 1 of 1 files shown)


installer/bundle/bundle_skel.sh
```diff
@@ -830,13 +830,18 @@ if [ "$installMode" = "I" -o "$installMode" = "U" ]; then
     fi
     curl_installed
     if [ $? -ne 0 ]; then
+        echo "curl was not found, attempting to install curl..."
+        install_extra_package curl
+    fi
+    check_program_in_path curl
```
lagalbra: For the other packages we do this for (tar, sed), we also check the return value of the installed status like `sed_package_installed=$?`. For consistency and overall checking a `curl_package_installed=$?` like should be added here as well.

NarineM: That step is unnecessary. I noticed a duplicate code in tar and sed.  We would need to clean that up instead. These are programs agent installer depends on, that is all. whether they show up in package managers or not , that would be an issue with curl installer. In the agent code base we directly call these programs from command line.

lagalbra: I do not recall an exact case where the "which" check passed but the package was not installed, or vice versa. Therefore, if you think those checks should be removed from the other package installs, you can clean those lines up.

NarineM: That is the point. You were just proposing to check them both in curl case  like it was done for sed.
I did not make sed and tar changes, and they are not related to this PR. I am going to leave it to you to make the changes and test that whenever you have time since you owned it. thanks.

PR #497, opened 2017-06-14
Update omsbaseline, add omsremediate

taglines:
This updates the omsbaseline binaries with bug fixes, adds the omsremediate binaries and updates the oms_audits.xml file with changes to the remediation section so that it refers to omsremediate instead of azsecd, and adds the remediation instructions needed for omsremediate.


lagalbra:
@Microsoft/omsagent-devs 
@taglines Do you want this change to be released with the OMSAgent June release?

PR #499, opened 2017-06-15
Remove default perf counters

NarineM:
This PR resolves the issue with omsagent default perf counters being out of sync with OMs workspace defaults. Before the fix omsagent after clean installation will send data for 3 perf counters: Memory, Processor and LogicalDisk, when by default OMS workspace in the portal is setup without any defaults.

@Microsoft/omsagent-devs


NarineM:
@gauravga 
Can you review this PR? Thanks.

PR #500, opened 2017-06-22
Cleanup markdown in Security-Events-Preview-Configuration.md

lagalbra:
@Microsoft/omsagent-devs @mgladi 

Ensure multi-line code-formatted lines showed up in a block rather than a wrapped line
Other small fixes


lagalbra:
These markdown mistakes caused the customer incident 40727773, and one other customer complaint, so this should be resolved soon.

PR #501, opened 2017-06-23
npmd: utilizing FQDN retrived via oms_common

chparimi:
Getnameinfo for a particular IP was used to get the FQDN. That API
was returning "Temporary failure in name resolution". Now utilizing
the API provided by oms_common for getting FQDN.


lagalbra:
@Microsoft/omsagent-devs

PR #505, opened 2017-06-28
npmd: checking if subnet id with 0.0.0.0 and big prefix is ipv6

chparimi:
Trying to see if 0.0.0.0 with prefix like 128 is ipv6.


lagalbra:
@Microsoft/omsagent-devs

PR #507, opened 2017-06-29
Assign error codes to onboarding process

lagalbra:
Detect and log possible internet connectivity issues

@Microsoft/omsagent-devs 

pBuild running now.
Error codes verified on a test machine.

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -48,6 +48,10 @@ RUBY=/opt/microsoft/omsagent/ruby/bin/ruby
 AUTH_KEY_SCRIPT=/opt/microsoft/omsagent/bin/auth_key.rb
 TOPOLOGY_REQ_SCRIPT=/opt/microsoft/omsagent/plugin/agent_topology_request_script.rb
 MAINTENANCE_TASKS_SCRIPT=/opt/microsoft/omsagent/plugin/agent_maintenance_script.rb
+# Ruby error codes
+AGENT_MAINTENANCE_MISSING_CONFIG_FILE=4
+AGENT_MAINTENANCE_MISSING_CONFIG=5
+AGENT_MAINTENANCE_ERROR_WRITING_TO_FILE=12
 
 METACONFIG_PY=/opt/microsoft/omsconfig/Scripts/OMS_MetaConfigHelper.py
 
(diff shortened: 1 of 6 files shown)


lagalbra:
pBuild passed.

lagalbra:
Thanks @jeffaco ! I'll leave this up for another few days if anyone from Ruby side to review ( @Microsoft/omsagent-devs ); otherwise I will check in.

installer/scripts/omsadmin.sh
```diff
@@ -573,16 +614,19 @@ onboard()
             # This is a temp solution since the DSC doesn't support multi-homing now
             # Only the primary workspace receives the configuration from the DSC service
             if [ "$USER_ID" -eq "0" ]; then
-                su - $AGENT_USER -c $METACONFIG_PY > /dev/null
+                su - $AGENT_USER -c $METACONFIG_PY > /dev/null || error=$?
```
lagalbra: @jeffaco Do you have any idea why this error check wasn't present before? As far as I can tell, that's been the case for a while.

jeffaco: It looks like this change has been there for a while:

```
commit e40eddf
Author: Robbie Zhang <junjiez@microsoft.com>
Date:   Wed Jan 11 09:51:38 2017

    Remove the collectd configuration from the bundle
    
    Call omsadmin.sh -c to configure the collectd
    Configure the omsconfig only when primary workspace is onboarded
```

At least that's how it appears at first glance. Note that this differs based on user ID, in case that's relevant for you to understand this.

lagalbra:
Added documentation for the error codes

installer/scripts/omsadmin.sh
```diff
@@ -514,17 +538,34 @@ onboard()
...
-        return 1
+        cleanup_certs
+        return $ERROR_ONBOARDING
     fi
 
     if [ "$RET_CODE" = "200" ]; then
         $RUBY $MAINTENANCE_TASKS_SCRIPT --endpoints "$RESP_ONBOARD","$ENDPOINT_FILE" "$CONF_OMSADMIN" "$FILE_CRT" "$FILE_KEY" "$RUN_DIR/omsagent.pid" "$CONF_PROXY" "$OS_INFO" "$INSTALL_INFO" $CURL_VERBOSE
-        if [ $? -ne 0 ]; then
-            log_warning "During onboarding request, certificate update and DSC endpoints may not have been extracted."
+        endpoints_ret=$?
```
taglines: The agent maintenance script returns 12 different error codes, but only three are exposed here. If the rest are not exposed, why was the main script modified to return them?

source/code/plugins/agent_maintenance_script.rb
```diff
@@ -317,18 +335,19 @@ def form_post_request_and_http(headers, uri_string, body, cert, key)
     def heartbeat
       # Reload config in case of updates since last heartbeat
       @load_config_success = load_config
```
taglines: Since this is no longer semantically a Boolean, it might make sense to rename it to load_config_result_code, or something similar.

installer/scripts/omsadmin.sh
lagalbra: If you're asking why agent_maintenance_script.rb returns more than 3, it's for generally improving it. Only the 3 error codes exposed in omsadmin.sh are the main ones that could be encountered during onboarding.

source/code/plugins/agent_maintenance_script.rb
lagalbra: Agreed. Changed.

PR #508, opened 2017-06-29
npmd: check if agentid file exists to get omsagent guid

chparimi:
If agentid file exists then utilize that for computer id.
To identify an instance of omsagent we utilize computer id and
workspace id.

### DIFF
--- a/test/code/plugins/in_npmd_server_plugintest.rb
+++ b/test/code/plugins/in_npmd_server_plugintest.rb
@@ -1,6 +1,7 @@
 require 'fluent/test'
 require_relative '../../../source/code/plugins/in_npmd_server'
 require 'socket'
+require 'securerandom'
 
 class Logger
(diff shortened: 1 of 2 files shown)


test/code/plugins/in_npmd_server_plugintest.rb
```diff
@@ -15,6 +15,12 @@ class << self
 
 class NPMDServerTest < Test::Unit::TestCase
 
+    class Fluent::NPM
+        def get_agent_id_from_guid_file()
```
lagalbra: I would recommend creating a couple unit tests for the behavior of this "agent_id" string you're expecting.

chparimi: Done.

lagalbra:
@chparimi You should close this PR since I have opened PR https://github.com/Microsoft/OMS-Agent-for-Linux/pull/527

PR #517, opened 2017-07-12
Extended syslog documentation based on ICMs

NarineM:
This PR contains improvements to syslog documentation based on the ICMs that we received due to the lack of certain syslog topics.
@Microsoft/omsagent-devs

### DIFF
--- a/docs/OMS-Agent-for-Linux.md
+++ b/docs/OMS-Agent-for-Linux.md
@@ -296,7 +296,7 @@ sudo /opt/microsoft/omsconfig/Scripts/OMS_MetaConfigHelper.py -enable
 ```
 
 ### Syslog events
-Syslog events are sent from the syslog daemon (e.g. rsyslog or syslog-ng) to a local port the agent is listening on (by default port 25224). When the agent is installed, a default syslog configuration is applied. This can be found at:
+Syslog events are sent from the syslog daemon (e.g. rsyslog or syslog-ng) to a local port the agent is listening on (by default port 25224). When the agent is installed, a default syslog configuration is applied. See [OMS syslog overview](https://blogs.technet.microsoft.com/msoms/2016/05/12/syslog-collection-in-operations-management-suite/). This can be found at:
 •	Rsyslog: `/etc/rsyslog.d/rsyslog-oms.conf`
 •	Syslog-ng: `/etc/syslog-ng/syslog-ng.conf`
 The default OMS Agent syslog configuration uploads syslog events from all facilities with a severity of warning or higher. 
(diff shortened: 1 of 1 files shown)


docs/OMS-Agent-for-Linux.md
```diff
@@ -380,7 +380,31 @@ If you are using selinux, the semanage tool can be used to allow TCP traffic for
 	  tag oms.syslog
 </source>
 ```
-**Note:** ports 0-1024 are privileged, and require special permissions. As the OMS Agent for Linux runs as the `omsagent` user trying to allocate a source port of 0 to 1024 results in agent startup failure
+### Syslog troubleshooting
+**Note:** ports 0-1024 are privileged, and require special permissions. As the OMS Agent for Linux runs as the `omsagent` user trying to allocate a source port of 0 to 1024 results in agent startup failure.
+
+**Checking for existance of a syslog collector**
+
+Make sure there is a syslog collector available by running the following command in shell:
```
lagalbra: nit: commands, plural

docs/OMS-Agent-for-Linux.md
```diff
@@ -380,7 +380,31 @@ If you are using selinux, the semanage tool can be used to allow TCP traffic for
...
+```
+	PORT=<port you use>
+	netstat -an | grep $PORT
+```
+**Checking port availability using netstat**
+```
+	PORT=<port you want to use>	
+	[ -z "`netstat -an | grep $PORT`" ] && echo "$PORT port is available" || echo "$PORT port is NOT available"
+```
+**Customizing syslog log format using fluentd' format regexp parameter if syslog data shows up trancated in OMS portal**
```
lagalbra: typo: "using **FluentD's** format"... with s and capitalization
mispell: truncated

docs/OMS-Agent-for-Linux.md
```diff
@@ -380,7 +380,31 @@ If you are using selinux, the semanage tool can be used to allow TCP traffic for
...
+
+OMS agent syslog configuration(/etc/opt/microsoft/omsagent/<your_wid>/conf/omsagent.d/syslog.conf) is using fluentd's default syslog's log format. If your device is generating non-standard syslog formats you can add custom filters to syslog.conf through fluentd 'format' parameter as regexp as documented [here](http://docs.fluentd.org/v0.10/articles/in_syslog).
```
lagalbra: nit: path should be in `markdown code format`
remove unnecessary apostrophe: "default **syslog** log"
nit: fully spell out "regular expressions"

lagalbra:
@keikhara Is there a newer OMSAgent syslog blog or document we could reference?

NarineM:
Updated to fix the typos.

NarineM:
@lagalbra 
There is only one blob that has the diagrams that is currently linked. There is no design change in syslog, so I would think the link is fine.

docs/OMS-Agent-for-Linux.md
```diff
@@ -380,7 +380,35 @@ If you are using selinux, the semanage tool can be used to allow TCP traffic for
...
+```
+**Checking port availability using netstat**
+```
+	PORT=<port you want to use>	
+	[ -z "`netstat -an | grep $PORT`" ] && echo "$PORT port is available" || echo "$PORT port is NOT available"
+```
+**Customizing syslog log format using FluentD's format regular expression parameter if syslog data shows up truncated in OMS portal**
+
+OMS agent syslog configuration(see below for the syslog.conf location) is using FluentD's default syslog log format. If your device is generating non-standard syslog formats you can add custom filters to syslog.conf through fluentd 'format' parameter as regexp as documented [here](http://docs.fluentd.org/v0.10/articles/in_syslog).
+**OMS agent syslog configuration file location**
```
lagalbra: This line isn't getting indented as I expected; the "OMS agent syslog configuration file location" text shows up right after the "here" link. I think an inline code block would be fine, like so:

OMS agent syslog configuration (in the file `/etc/opt/microsoft/omsagent/<workspace_id>/conf/omsagent.d/syslog.conf`) is using...

keikhara:
No that is the latest. Do you want me to write a new blog?

lagalbra:
Syslog designed hasn't really changed since the blog went out; once mutli-homing is released, we should update the "customizing" steps, but for now it's still relevant.

Issue #451, opened 2017-04-20
missing netstat command in centos 7 core install


cjmather:
I don't see centos 7 core explicitly listed as supported or unsupported.  If it's not supported, then the installer should fail immediately (it doesn't).  If it is supported, then the installer should work around the missing netstat command:

...
Updating / installing...
   1:omsagent-1.3.3-15                ################################# [ 50%]
info	Reading onboarding params from: /etc/omsagent-onboard.conf
info	Reusing previous agent GUID
info	Onboarding success
**/opt/microsoft/omsagent/bin/omsadmin.sh: line 798: netstat: command not found**
Configuring rsyslog for OMS logging
Restarting service: rsyslog
**/opt/microsoft/omsagent/bin/omsadmin.sh: line 798: netstat: command not found**
Configuring OMS agent service ...

lagalbra:
CentOS 7 is supported: https://github.com/Microsoft/OMS-Agent-for-Linux/tree/master#supported-linux-operating-systems

Why is the netstat command not available on your machine? Is it a custom build?

lagalbra:
Spoke offline with Cliff; this machine is a CentOS7 core machine, which we do not support. I'm closing this issue with that in mind.

NarineM:
The installer should be fixed to handle Core failure better. It unnecessary creates the files and directories that come with the installer. Then to clean up one needs to run the shell with --remove.

NarineM:
This issue has been addressed in https://github.com/Microsoft/OMS-Agent-for-Linux/pull/520

NarineM:
The fix will be available in next omsagent release.

PR #520, opened 2017-07-18
Better handle onboarding if netstat is not present

NarineM:
@Microsoft/omsagent-devs


NarineM:
This PR addresses the following GitHub issue: https://github.com/Microsoft/OMS-Agent-for-Linux/issues/451

PR #522, opened 2017-07-24
npm: change unix socket ACL to restrict access to file

chparimi:
Altering unix socket ACL to prevent access to file.

### DIFF
--- a/source/code/plugins/in_npmd_server.rb
+++ b/source/code/plugins/in_npmd_server.rb
@@ -171,6 +171,10 @@ module Fluent
                     File.unlink(@location_unix_endpoint)
                 end
                 @server_obj = UNIXServer.new(@location_unix_endpoint)
+
+                if File.exist?(@location_unix_endpoint)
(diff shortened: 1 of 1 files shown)


lagalbra:
@Microsoft/omsagent-devs

source/code/plugins/in_npmd_server.rb
```diff
@@ -171,6 +171,10 @@ def setup_endpoint
                     File.unlink(@location_unix_endpoint)
                 end
                 @server_obj = UNIXServer.new(@location_unix_endpoint)
+
+                if File.exist?(@location_unix_endpoint)
+                    FileUtils.chmod 0700, @location_unix_endpoint
```
lagalbra: Can you add parentheses around your parameters? It looks a little inconsistent to me with the style in the lines around it, especially since there is more than one parameter for this line.
Also, is 700 necessary? Would you be able to get away with only 600?

lagalbra: Thanks!

PR #523, opened 2017-07-24
Make OMS::Log usage consistent across plugins

lagalbra:
@Microsoft/omsagent-devs 

Currently, Log.error_once (and warn_once, etc.) statements assume the Fluent::Log is being referred to, and this error shows in omsagent.log:
`.... error_class="NoMethodError" error="undefined method `error_once' for Fluent::Log:Class" ....`

This change explicitly scopes all uses of the OMS::Log class.

### DIFF
--- a/source/code/plugins/flattenjson_lib.rb
+++ b/source/code/plugins/flattenjson_lib.rb
@@ -9,7 +9,7 @@ module OMS
       begin
         data = eval(select)
       rescue => e
-        Log.error_once("Invalid select: #{select} #{e}");
+        OMS::Log.error_once("Invalid select: #{select} #{e}");
       end
 
       if data.instance_of?(Array)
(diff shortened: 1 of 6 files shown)


NarineM:
@lagalbra,
Is OMS::Log ruby standard log? Why fluentd log is failing?

lagalbra:
Looks like my pasted error message didn't get copied correctly above. It seems to be failing because the FluentD log (code [here](https://github.com/fluent/fluentd/blob/5e491f8f9dd93a8ed359f10d612272e7fb835ca2/lib/fluent/log.rb)) does not have error_once and warn_once methods, which the OMS Log specially defined.
Given that this error message occurred, it's clear that the Log Ruby assumes is the FluentD log.

NarineM:
Basically, these are typos, someone assumed fluentd and ruby interfaces are similar when they have different api names. Signoff.

PR #524, opened 2017-07-24
Heartbeat: let response be analyzed before returning the error code

lagalbra:
@Microsoft/omsagent-devs 

Commit 7096a0cc08e5f372477b16c525e8b9b40a48b161  added "return ERROR_SENDING_HTTP" which returns prematurely; the response from the failed HTTP transaction is checked and a return code is appropriately handled there.


lagalbra:
(rebased against master)

PR #526, opened 2017-07-24
Stop is-running check from printing out "does not exist" messages

lagalbra:
@Microsoft/omsagent-devs 

Without this change, every time a new workspace is onboarded, "Warning: Specified workspace xxxxxx does not exist on the machine. Check the correctness of the workspace ID and that onboarding succeeded." is printed to the console. This may be confusing to customers.

I have verified with this change that newly added workspaces still get onboarded, and already-added workspaces print "Workspace xxxxx already onboarded and agent is running." and exit cleanly.


lagalbra:
This commit was included in the release branch https://github.com/Microsoft/OMS-Agent-for-Linux/tree/v1.4.0-45 and is building in Jenkins now.

PR #528, opened 2017-07-25
Skip time skew test if ntpdate command is not found

lagalbra:
@Microsoft/omsagent-devs 

If the ntpdate command is not found on the machine, then this test will return immediately.

On pBuild-based CentOS 5 machines, the ntpdate command is not installed (and I have not yet discovered how to install it). Because of this, system tests always fail and stop before running [systemtestsh](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/build/Makefile#L575) tests.


lagalbra:
Ping! @gauravga Can you please review?

PR #529, opened 2017-07-26
Fix baseline commandline

taglines:


taglines:
I did a pbuild and then tested it.

PR #530, opened 2017-07-28
VMware syslog: resource fields

arjun-urs:
Cleaned up resource fields from the VMware Syslog plugin.

@Microsoft/omsagent-devs @keikhara


arjun-urs:
pBuild and tests are checked. Expected data flowing into OMS portal.

PR #532, opened 2017-07-28
Send custom logs less aggressively

lagalbra:
Increase flush_interval for the output plugin sending custom logs

@Microsoft/omsagent-devs


lagalbra:

PR #533, opened 2017-08-01
Machines with apt-get should update package list before attempting installation

lagalbra:
@Microsoft/omsagent-devs 

Without this package list update, the curl package will not be found on new Azure Debian VMs, and the OMSAgent will fail installation with error code 64.


lagalbra:
Closing since changes will be implemented in extension code.

PR #538, opened 2017-08-08
Enable multi-homing for in_dsc_monitor.rb

lagalbra:
@Microsoft/omsagent-devs 

Have tested this by onboarding multiple workspaces; they get separate DSC cache files placed and proper string replacement in each copy of operation.conf.


lagalbra:
pBuild has passed.

lagalbra:

PR #542, opened 2017-08-08
Set process limit for the omsagent user

lagalbra:
Maximum number of processes found during customer and in-house testing was 28; setting default to 75 to be safe
Give option in omsadmin.sh to set and reset the limit
Remove duplicate onboarding system tests

pBuild and system tests passed
Verified on a test machine by setting the limit to 10; no new processes were started. Main FluentD process continued to run.

@Microsoft/omsagent-devs


test/installer/scripts/test_omsadmin.sh
```diff
@@ -1,42 +0,0 @@
-BASE_DIR=$1
-RUBY_TESTVERS=$2
```
NarineM: This test was for testing omsadmin.sh shell script with onboarding parameters. why you think it has to be deleted?

test/installer/scripts/omsadmin_systest.rb
```diff
@@ -130,4 +131,45 @@ def test_onboard_two_workspaces
     assert_match(/No Workspace/, output_list, "No workspace should be listed: #{output_list}")
   end
 
+  def test_set_proc_limit_default
```
NarineM: Please add a test for proc minimum.

installer/scripts/omsadmin.sh
```diff
@@ -352,6 +367,25 @@ set_proxy_setting()
     fi
 }
 
+set_omsagent_proc_limit()
```
NarineM: I would expect this API to have a minimum accepted limit.

test/installer/scripts/test_omsadmin.sh
lagalbra: [omsadmin_systest.rb](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/test/installer/scripts/omsadmin_systest.rb#L60) already covers this test case and more.

installer/scripts/omsadmin.sh
lagalbra: Why? And if I add it, what should the minimum be? If a customer is setting this limit, they already are familiar with how processes work, and know that setting the limit too low may hinder functionality.

test/installer/scripts/test_omsadmin.sh
NarineM: If that is the case, please update the PR description to state "duplicate" instead of "redundant".

installer/scripts/omsadmin.sh
NarineM: Customer would not know what is the minimum accepted limit where the agent can still function.

NarineM: I am going to leave this PR for someone else to review who was initially involved with ulimit design discussions. I am still not clear what would be omsagent behavior if customer puts 1 as proc limit and how that can impact omiagent?

lagalbra: I have not tested with numbers as small as 1; I will set a minimum limit of 5 to safeguard against unwise input.

test/installer/scripts/test_omsadmin.sh
lagalbra: PR description updated.

test/installer/scripts/omsadmin_systest.rb
lagalbra: Added with `test_set_proc_limit_below_min`; testing now

lagalbra:
@keikhara Can you please review, since you were part of the design discussions?

lagalbra:
Test for minimum ulimit added and passed. @NarineM Please review.

installer/scripts/omsadmin.sh
```diff
@@ -346,6 +362,28 @@ set_proxy_setting()
...
```
Atchub: can we apply at group level or do we need to include nxautomation user too as it is also a user created by OMS agent

Atchub: does nproc restrict for shell commands executed with root permissions? just checking if there is a constraint

lagalbra: The scope of this task was to control the number of processes under the omsagent user specifically.
The hybrid worker should be able to set a different limit for the nxautomation user if they need to in the future.

NarineM: Good point, when reviewing I only taken omsconfig and omsagent process related users into an account.
@lagalbra , can you look into Atchut's concern?

lagalbra: I believe that would run as root, so this limit wouldn't restrict it, no.

lagalbra: To answer Atchut's question a little further, yes, these limits can be placed at the group level.
To answer Narine's question, no. The owners of the the nxautomation user should place limits if desired, since they have any and all context for what the limits should be. For now, the customer is only looking for limits on the omsagent user.

NarineM: By omsagent user they most likely meant the users that are used to perform operations within or by omsagent related process , so I would think processed spinned up by nxautomation will also count.

lagalbra: Spoke with @keikhara ; our aim is to make the resource limits as specific as possible. If we place the limit at the group level instead of the user level, the omsagent could be pointlessly throttled when the nxautomation user misbehaves.

I will send out an email asking the automation team to add a process limit for the nxautomation worker, after they have performed similar testing for their user's processes.

Given this, please approve this PR.

lagalbra:
We do not have any system tests against the agent service; I have verified this end-to-end myself.

The agent service does not stop if you set the limit to a small number. It's already running, so it keeps running, but new processes over that number cannot be started.

NarineM:
we should be adding new tests for this feature work.I do not think unit test level coverage is enough.

lagalbra:
Rebased against master. Will merge once Travis completes.

lagalbra:
I've verified on a test machine that the agent can be successfully restarted when the ulimit is set to 25.

PR #548, opened 2017-08-11
Use omiconfigeditor to determine state of port 1270

vimish:
@Microsoft/omsagent-devs 
@jeffaco 

omiconfigeditor now supports query option. Existing
logic, to determine if port 1270 is open or not, is
updated to use omiconfigeditor.

PBUILD has passed for the changes.

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -95,6 +95,7 @@ PROC_LIMIT_CONF=/etc/security/limits.conf
 # SCOM variables
 SCX_SSL_CONFIG=/opt/microsoft/scx/bin/tools/scxsslconfig
 OMI_CONF_FILE=/etc/opt/omi/conf/omiserver.conf
+OMI_CONF_EDITOR=/opt/omi/bin/omiconfigeditor
 
 # Error codes and categories:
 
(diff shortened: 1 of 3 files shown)


installer/scripts/omsadmin.sh
```diff
@@ -445,8 +428,8 @@ onboard_scom()
     is_scom_port_open
     if [ $? -eq 1 ]; then
         echo "Opening port 1270"
-        /opt/omi/bin/omiconfigeditor httpsport -a 1270 < /etc/opt/omi/conf/omiserver.conf > /etc/opt/omi/conf/omiserver.conf_temp
-        mv /etc/opt/omi/conf/omiserver.conf_temp /etc/opt/omi/conf/omiserver.conf
+        $OMI_CONF_EDITOR httpsport -a 1270 < $OMI_CONF_FILE > /etc/opt/omi/conf/omiserver.conf_temp
+        mv /etc/opt/omi/conf/omiserver.conf_temp $OMI_CONF_FILE
```
jeffaco: This could be rewritten as:

```
$OMI_CONF_EDITOR httpsport -a 1270 < $OMI_CONF_FILE > ${OMI_CONF_FILE}_temp
mv ${OMI_CONF_FILE}_temp $OMI_CONF_FILE
```

Just another way to do the same thing. I'm not asking for the change, just pointing it out.

installer/scripts/omsadmin.sh
```diff
@@ -95,6 +95,7 @@ PROC_LIMIT_CONF=/etc/security/limits.conf
 # SCOM variables
 SCX_SSL_CONFIG=/opt/microsoft/scx/bin/tools/scxsslconfig
 OMI_CONF_FILE=/etc/opt/omi/conf/omiserver.conf
+OMI_CONF_EDITOR=/opt/omi/bin/omiconfigeditor
```
lagalbra: @jeffaco Is there a minimum version of OMI that supports this? If this was only added in a recent version, then we may want to update our dependencies: https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/datafiles/linux_rpm.data#L10 and https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/datafiles/linux_dpkg.data#L9

jeffaco: That is true. @vimish added this to OMI v1.3. So OMS now has a minimum dependency of OMS v1.3 since it depends on this capability. @vimish, I'm going to mark my review "Waiting" for this change.

vimish:
@jeffaco @lagalbra 

Updated the min version of omi to 1.3.0-2.
Also updated min version of scx to 1.6.3-212 as omsagent bundle supports --enable-opsmgr flag. This flag is introduced in scx 1.6.3-212 bundle.

lagalbra:
@jeffaco Can you approve so that Vishal can merge? I've tested this briefly myself, and it seems to work.

PR #551, opened 2017-08-15
Remove omsagent process limit when omsagent is uninstalled

lagalbra:
Without this change, /etc/security/limits.conf never gets cleaned up after the OMSAgent is uninstalled. Additionally, a customer can set a limit low enough which restricts proper cleanup.

Build, unit, and system tests pass.
Verified that limit is removed after installing and removing.

@Microsoft/omsagent-devs


lagalbra:

PR #552, opened 2017-08-16
Make omsadmin.sh script option available to call is_scom_port_open

lagalbra:
This will allow SCOM detection through the Azure VM extension.

Verified locally.

@Microsoft/omsagent-devs


lagalbra:
@vimish Yes, the purpose of this PR is merely to check that port 1270 is open. The check for the certificate will be done in the OmsAgentForLinux extension code (a little easier to do in python then sh)

lagalbra:
Rebased against master. @vimish Let me know if my comment above is not clear, or you have any other concerns before I merge this to master.

PR #559, opened 2017-08-24
Revert "Move AGENT_GUID to non-workspace-specific location to allow for single-machine agent ID"

lagalbra:
This reverts commit 1e5a6b78f35ecf6cc4369c7d0027a5cb7f9cc506.
Do not revert changes from diagnostic and NPM plugins - these changes were backwards-compatible.
This change will not be released because it violates the OMS security model.

@Microsoft/omsagent-devs @chparimi


lagalbra:
Unit and system tests passed.
pBuild passed.

PR #562, opened 2017-08-25
Ensure omi_mapping_path is specified for performance metric config

lagalbra:
@Microsoft/omsagent-devs 

This change does not affect performance metrics that are collect at onboarding time; however, this will match the expectations of the nxOMSPerfCounter resource.

### DIFF
--- a/docs/OMS-Agent-for-Linux.md
+++ b/docs/OMS-Agent-for-Linux.md
@@ -418,6 +418,7 @@ Each Object (or category) of performance metrics to collect should be defined in
   instance_regex ".*"
   counter_name_regex ".*"
   interval 30s
+  omi_mapping_path /etc/opt/microsoft/omsagent/<workspace id>/conf/omsagent.d/omi_mapping.json
 </source>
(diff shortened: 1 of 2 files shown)


NarineM:
Is this change backwards compatible? Older agents even after upgrade will have confs written in old style with no omi_mapping property. How would the plugin work in that case?

lagalbra:
The plugin provides a default value for the omi_mapping_path property: https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/source/code/plugins/in_oms_omi.rb#L18 This plugin [has always had the omi_mapping_path property](https://github.com/Microsoft/OMS-Agent-for-Linux/commit/1373be7d53d07bda8c1b3244de165eb4d0f41236).
has always had the omi_mapping_path property.

The changes in this PR are for commented-out code and documentation, thus they won't be used and don't need to be backwards-compatible. However, the DSC resource change I'm making will be backwards-compatible with pre-multi-homing agent versions.

PR #567, opened 2017-09-01
Xcshowplim

v-xecamp:
Two items:
1.  Adds -c switch to display limit value, and in cases of missing entry or multiple entries provide logging too.
2.  Adds feature to insert new $AGENT_USER entry 2 lines before the end of the file to avoid making the typical two line sequence indicating the end of this configuration file end up out of place.

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -162,7 +162,7 @@ usage()
     echo "Detect if omiserver is listening to SCOM port:"
     echo "$basename -o"
     echo
-    echo "Show the configured process limit in /etc/limits.conf for user $AGENT_USER."
+    echo "Show the configured process limit in $PROC_LIMIT_CONF for user $AGENT_USER."
     echo "$basename -c"
 }
 
@@ -242,7 +242,7 @@ parse_args()
 {
     local OPTIND opt
 
-    while getopts "h?s:w:d:vp:u:a:lx:XUm:Nrn:oRc" opt; do
+    while getopts "h?s:w:d:vp:u:a:lx:XUm:cNrn:oR" opt; do
         case "$opt" in
         h|\?)
             usage
@@ -422,7 +422,7 @@ unset_omsagent_proc_limit()
 find_current_omsagent_proc_limit()
 {
     #cat $PROC_LIMIT_CONF | grep -E "$LIMIT_LINE_REGEX" > /dev/null 2>&1
-    get_current_omsagent_proc_limit >/dev/null 2>/dev/null
+    get_current_omsagent_proc_limit > /dev/null 2>&1
     return $?
 }
 
@@ -841,7 +841,8 @@ show_omsagent_proc_limit()
         result_limit=`get_current_omsagent_proc_limit | tail -1 | awk '{print $4}'`
         if [ "$result_limit" -eq "-1" ]; then
             result_limit="No Limit"
-            log_info "There is NO present limit to number of processes for $AGENT_USER."
+            #log_info "There is NO present limit to number of processes for $AGENT_USER."
+            log_info "The number of processes for $AGENT_USER is unlimited."
         elif [ "$result_limit" -lt "$MIN_OMSAGENT_PROC_LIMIT" ]; then
             log_warning "$AGENT_USER process limit setting of '$result_limit' is less than minimum of $MIN_OMSAGENT_PROC_LIMIT."
         fi
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -150,6 +150,9 @@ usage()
     echo "Azure resource ID:"
     echo "$basename -a <Azure resource ID>"
     echo
+    echo "Show the configured process limit in $PROC_LIMIT_CONF for user $AGENT_USER."
+    echo "$basename -c"
+    echo
     echo "Set process limit for OMSAgent:"
     echo "$basename -n <specific number limit>"
     echo
@@ -161,9 +164,6 @@ usage()
     echo
     echo "Detect if omiserver is listening to SCOM port:"
     echo "$basename -o"
-    echo
-    echo "Show the configured process limit in $PROC_LIMIT_CONF for user $AGENT_USER."
-    echo "$basename -c"
 }
 
 set_user_agent()
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -90,7 +90,7 @@ MONITOR_AGENT_PORT=$DEFAULT_MONITOR_AGENT_PORT
 DF_OMSAGENT_PROC_LIMIT=75
 MIN_OMSAGENT_PROC_LIMIT=5
 PROC_LIMIT_CONF=/etc/security/limits.conf
-LIMIT_LINE_REGEX="^[[:space:]]*$AGENT_USER[[:space:]]+(hard|soft)[[:space:]]+nproc[[:space:]]+[0-9]+[[:space:]]*$"
+LIMIT_LINE_REGEX="^[[:space:]]*$AGENT_USER[[:space:]]+(hard|soft)[[:space:]]+nproc[[:space:]]+-?[0-9]+[[:space:]]*$"
 
 # SCOM variables
 SCX_SSL_CONFIG=/opt/microsoft/scx/bin/tools/scxsslconfig
@@ -398,7 +398,7 @@ set_omsagent_proc_limit()
 
     log_info "Setting process limit for the $AGENT_USER user in $PROC_LIMIT_CONF to $NEW_OMSAGENT_PROC_LIMIT..."
     local new_omsagent_line="$AGENT_USER  hard  nproc  $NEW_OMSAGENT_PROC_LIMIT"
-    find_current_omsagent_proc_limit
+    get_current_omsagent_proc_limit >/dev/null 2>/dev/null
     if [ $? -eq 0 ]; then
         old_omsagent_line=`cat $PROC_LIMIT_CONF | grep -E "$LIMIT_LINE_REGEX" | tail -1`
         sed -i s,"$old_omsagent_line","$new_omsagent_line",1 $PROC_LIMIT_CONF
@@ -409,7 +409,7 @@ set_omsagent_proc_limit()
 
 unset_omsagent_proc_limit()
 {
-    find_current_omsagent_proc_limit
+    get_current_omsagent_proc_limit >/dev/null 2>/dev/null
     if [ $? -eq 0 ]; then
         # Remove all lines for the omsagent from the limit file
         log_info "Removing process limit for the $AGENT_USER user in $PROC_LIMIT_CONF..."
@@ -421,7 +421,8 @@ unset_omsagent_proc_limit()
 
 find_current_omsagent_proc_limit()
 {
-    cat $PROC_LIMIT_CONF | grep -E "$LIMIT_LINE_REGEX" > /dev/null 2>&1
+    #cat $PROC_LIMIT_CONF | grep -E "$LIMIT_LINE_REGEX" > /dev/null 2>&1
+    get_current_omsagent_proc_limit >/dev/null 2>/dev/null
     return $?
 }
 
@@ -821,6 +822,7 @@ show_workspace_status()
 
 show_omsagent_proc_limit()
 {
+    local result_limit="No Limit"
     if [ ! -f $PROC_LIMIT_CONF ]; then
         log_error "Limits configuration file '$PROC_LIMIT_CONF' missing!"
         return 1
@@ -829,7 +831,6 @@ show_omsagent_proc_limit()
 
     if [ "$count" -eq "0" ]; then
         log_info "There is NO present limit to number of processes for $AGENT_USER."
-        echo ""
     else
         if [ "$count" -gt "1" ]; then
             log_info "$count $AGENT_USER entries found in $PROC_LIMIT_CONF."
@@ -837,12 +838,15 @@ show_omsagent_proc_limit()
             get_current_omsagent_proc_limit
             log_info "End of the $count duplicate entry info Message"
         fi
-        local result_limit=`get_current_omsagent_proc_limit | tail -1 | awk '{print $4}'`
-        if [ "$result_limit" -lt "$MIN_OMSAGENT_PROC_LIMIT" ]; then
+        result_limit=`get_current_omsagent_proc_limit | tail -1 | awk '{print $4}'`
+        if [ "$result_limit" -eq "-1" ]; then
+            result_limit="No Limit"
+            log_info "There is NO present limit to number of processes for $AGENT_USER."
+        elif [ "$result_limit" -lt "$MIN_OMSAGENT_PROC_LIMIT" ]; then
             log_warning "$AGENT_USER process limit setting of '$result_limit' is less than minimum of $MIN_OMSAGENT_PROC_LIMIT."
         fi
-        echo $result_limit
     fi
+    echo $result_limit
     return 0
 }
 
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -242,7 +242,7 @@ parse_args()
 {
     local OPTIND opt
 
-    while getopts "ch?s:w:d:vp:u:a:lx:XUm:Nrn:oR" opt; do
+    while getopts "h?s:w:d:vp:u:a:lx:XUm:Nrn:oRc" opt; do
         case "$opt" in
         h|\?)
             usage
@@ -422,6 +422,12 @@ unset_omsagent_proc_limit()
 find_current_omsagent_proc_limit()
 {
     cat $PROC_LIMIT_CONF | grep -E "$LIMIT_LINE_REGEX" > /dev/null 2>&1
+    return $?
+}
+
+get_current_omsagent_proc_limit()
+{
+    cat $PROC_LIMIT_CONF | grep -E "$LIMIT_LINE_REGEX"
     return $?
 }
 
@@ -815,35 +821,28 @@ show_workspace_status()
 
 show_omsagent_proc_limit()
 {
-    local result_limit=
-
     if [ ! -f $PROC_LIMIT_CONF ]; then
         log_error "Limits configuration file '$PROC_LIMIT_CONF' missing!"
         return 1
     fi
-    local count=`cat $PROC_LIMIT_CONF | grep -E "$LIMIT_LINE_REGEX" | wc -l`
+    local count=`get_current_omsagent_proc_limit | wc -l`
 
     if [ "$count" -eq "0" ]; then
-        log_info "Process limit not currently set for $AGENT_USER."
-    elif [ "$count" -eq "1" ]; then
-        result_limit=`cat $PROC_LIMIT_CONF | grep -E "$LIMIT_LINE_REGEX" | awk '{print $4}'`
+        log_info "There is NO present limit to number of processes for $AGENT_USER."
+        echo ""
     else
-        log_info "$count $AGENT_USER entries found in $PROC_LIMIT_CONF.  This may yield problems.  Last entry should apply."
-        log_info ">>Last entry presumed to apply<<"
-        local limit_lines=`cat $PROC_LIMIT_CONF | grep -E "$LIMIT_LINE_REGEX"`
-        log_info $limit_lines
-        log_info "End of the $count duplicate entry WARNING Message"
-        local result_limit=`cat $PROC_LIMIT_CONF | grep -E "$LIMIT_LINE_REGEX" | tail -1 | awk '{print $4}'`
+        if [ "$count" -gt "1" ]; then
+            log_info "$count $AGENT_USER entries found in $PROC_LIMIT_CONF."
+            log_info ">>Last entry should apply<<" # NOTE:  This will not pick up -1 entries.
+            get_current_omsagent_proc_limit
+            log_info "End of the $count duplicate entry info Message"
+        fi
+        local result_limit=`get_current_omsagent_proc_limit | tail -1 | awk '{print $4}'`
+        if [ "$result_limit" -lt "$MIN_OMSAGENT_PROC_LIMIT" ]; then
+            log_warning "$AGENT_USER process limit setting of '$result_limit' is less than minimum of $MIN_OMSAGENT_PROC_LIMIT."
+        fi
+        echo $result_limit
     fi
-
-    if [ "$result_limit" -eq "-1" ]; then
-        log_info "$AGENT_USER process count unlimited."
-        result_limit=
-    elif [ "$result_limit" -lt "$MIN_OMSAGENT_PROC_LIMIT" ]; then
-        log_warning "$AGENT_USER process limit setting of '$result_limit' is less than minimum of $MIN_OMSAGENT_PROC_LIMIT."
-    fi
-
-    echo $result_limit
     return 0
 }
 
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -161,6 +161,9 @@ usage()
     echo
     echo "Detect if omiserver is listening to SCOM port:"
     echo "$basename -o"
+    echo
+    echo "Show the configured process limit in /etc/limits.conf for user $AGENT_USER."
+    echo "$basename -c"
 }
 
 set_user_agent()
@@ -239,7 +242,7 @@ parse_args()
 {
     local OPTIND opt
 
-    while getopts "h?s:w:d:vp:u:a:lx:XUm:Nrn:oR" opt; do
+    while getopts "ch?s:w:d:vp:u:a:lx:XUm:Nrn:oR" opt; do
         case "$opt" in
         h|\?)
             usage
@@ -300,6 +303,9 @@ parse_args()
             ;;
         R)
             RECONSTRUCT_WORKSPACE_STATE=1
+            ;;
+        c)
+            show_omsagent_proc_limit
             ;;
         esac
     done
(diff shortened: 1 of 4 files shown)


installer/scripts/omsadmin.sh
```diff
@@ -787,6 +804,36 @@ show_workspace_status()
...
+        log_warning ">>Last entry presumed to apply<<"
+        grep -E "$LIMIT_LINE_REGEX" $PROC_LIMIT_CONF 1>&2
+        log_warning "End of the $count duplicate entry WARNING Message"
+        llimit=`grep -E "$LIMIT_LINE_REGEX" $PROC_LIMIT_CONF | tail -1 | awk '{print $4}'`
+    fi
+
+    if [ "$llimit" -eq -1 ]; then
+        log_info "$AGENT_USER proc count unlimited."
+    elif [ "$llimit" -lt "$MIN_OMSAGENT_PROC_LIMIT" ]; then
+        log_error "$AGENT_USER proc limit setting of '$llimit' is less than minimum of $MIN_OMSAGENT_PROC_LIMIT."
```
lagalbra: nit: spell out "process"

installer/scripts/omsadmin.sh
```diff
@@ -787,6 +804,36 @@ show_workspace_status()
...
+        resultcode=1
+    fi
+
+    echo $llimit
+    return $resultcode
```
lagalbra: This method is a get() call; it should always succeed unless we don't find the file to read from. In my opinion, this should always return successful unless you want to check for a case (which I don't think should occur) of /etc/security/limits.conf not being present.
You log errors when an unexpected case happens, but in those cases you have still successfully shown the current limit.

test/installer/scripts/omsadmin_systest.rb
```diff
@@ -136,6 +136,7 @@ def test_set_proc_limit_non_default
     set_omsagent_proc_limit(5000)
     assert(FileUtils.compare_file(@proc_limits_conf, "#{@omsadmin_test_dir}/limits-non-default.conf"),
            "The user process limit was not set correctly to the new value")
+
```
lagalbra: nit: eliminate this line addition

test/installer/scripts/omsadmin_systest.rb
```diff
@@ -173,6 +174,13 @@ def test_unset_proc_limit_when_not_set
            "Process limit conf file was wrongly changed")
   end
 
+  def test_new_proc_limit
```
lagalbra: nit: make this test name more descriptive: `test_set_proc_limit_with_end_label`

test/installer/scripts/omsadmin_systest.rb
```diff
@@ -194,11 +202,36 @@ def set_omsagent_proc_limit(val = nil, should_succeed = true)
     return output
   end
 
+  def show_omsagent_proc_limit
+    lstderrprefix='/tmp/show_omsagent_proc_limit'
```
lagalbra: Since all of your logging for show_omsagent_proc_limit are log_* methods, they will print out "error" or "warning" or "info" with the message that you can check for. This stdout/stderr split creates more problems than I believe it solves, especially because it creates files in /tmp/ that may conflict when a machine has multiple tests running.

test/installer/scripts/omsadmin_systest.rb
```diff
@@ -194,11 +202,36 @@ def set_omsagent_proc_limit(val = nil, should_succeed = true)
...
     output = `#{@omsadmin_script} -r`
     ret_code = $?
     assert_equal(0, ret_code.to_i, "The command to unset the proc limit failed")
     return output
   end
 
+  def test_show_omsagent_proc_limit_default
+    set_omsagent_proc_limit()
+    lso, lse = show_omsagent_proc_limit()
+    assert_match(/^-?\d+$/,lso,"The command to show what should be the default proc limit did not yield an integer.")
```
lagalbra: nit: put spaces in all of your parameter/return lists

test/installer/scripts/omsadmin_systest.rb
```diff
@@ -194,11 +202,36 @@ def set_omsagent_proc_limit(val = nil, should_succeed = true)
...
+    assert_equal(lse, "", "Unexpected error: '#{lse}")
+  end
+
+  def test_show_omsagent_proc_limit_5000
+    set_omsagent_proc_limit(5000)
+    lso, lse = show_omsagent_proc_limit()
+    assert_match(/^5000$/,lso,"Should have been 5000 process limit.  Instead #{lso}.")
```
lagalbra: Ruby's assert_match will automatically print out what the expected pattern was and what it found instead; you don't have to do it yourself.

installer/scripts/omsadmin.sh
```diff
@@ -787,6 +804,36 @@ show_workspace_status()
...
+    else
+        log_warning "$count $AGENT_USER entries found in $PROC_LIMIT_CONF.  This may yield problems.  Last entry should apply."
+        log_warning ">>Last entry presumed to apply<<"
+        grep -E "$LIMIT_LINE_REGEX" $PROC_LIMIT_CONF 1>&2
+        log_warning "End of the $count duplicate entry WARNING Message"
+        llimit=`grep -E "$LIMIT_LINE_REGEX" $PROC_LIMIT_CONF | tail -1 | awk '{print $4}'`
+    fi
+
+    if [ "$llimit" -eq -1 ]; then
+        log_info "$AGENT_USER proc count unlimited."
```
lagalbra: Return in this case; you should not print out a limit of -1

test/installer/scripts/omsadmin_systest.rb
```diff
@@ -194,11 +202,36 @@ def set_omsagent_proc_limit(val = nil, should_succeed = true)
...
+    assert_equal(lse, "", "Unexpected error: '#{lse}")
+  end
+
```
lagalbra: Tests you should add:
- showing the limit set in a file with 2+ omsagent lines (you can use `#{@omsadmin_test_dir}/limits-two-settings.conf` found [here](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/test/installer/scripts/prep_omsadmin.sh#L73))
- showing the limit when there is no line for omsagent (you can use `#{@omsadmin_test_dir}/limits-no-settings.conf` found [here](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/test/installer/scripts/prep_omsadmin.sh#L83))

installer/scripts/omsadmin.sh
```diff
@@ -239,7 +242,7 @@ parse_args()
 {
     local OPTIND opt
 
-    while getopts "h?s:w:d:vp:u:a:lx:XUm:Nrn:o" opt; do
+    while getopts "ch?s:w:d:vp:u:a:lx:XUm:Nrn:o" opt; do
```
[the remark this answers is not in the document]
v-xecamp: I'm not able to see what's going on with this problem.  The new character is in my branch version, and as per my email I show I am pushed successfully.  We may need to compare notes on this directly.

installer/scripts/omsadmin.sh
```diff
@@ -787,6 +804,36 @@ show_workspace_status()
     fi
 }
 
+show_omsagent_proc_limit()
+{
+    llimit=-1
+    resultcode=0
```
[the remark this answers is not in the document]
v-xecamp: How about result_limit and returnvalue?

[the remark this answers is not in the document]
v-xecamp: Also I made them local.

installer/scripts/omsadmin.sh
```diff
@@ -787,6 +804,36 @@ show_workspace_status()
...
+
+    count=`cat $PROC_LIMIT_CONF | grep -E "$LIMIT_LINE_REGEX" | wc -l`
+
+    if [ "$count" -eq "0" ]; then
+        log_info "INFO:  OMS Agent process limit NOT Presently Set."
+    elif [ $count -eq 1 ]; then
```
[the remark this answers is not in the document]
v-xecamp: Yep.

installer/scripts/omsadmin.sh
```diff
@@ -787,6 +804,36 @@ show_workspace_status()
...
+        llimit=`cat $PROC_LIMIT_CONF | grep -E "$LIMIT_LINE_REGEX" | awk '{print $4}'`
+    else
+        log_warning "$count $AGENT_USER entries found in $PROC_LIMIT_CONF.  This may yield problems.  Last entry should apply."
+        log_warning ">>Last entry presumed to apply<<"
```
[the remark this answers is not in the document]
v-xecamp: Cool.

installer/scripts/omsadmin.sh
```diff
@@ -787,6 +804,36 @@ show_workspace_status()
...
+        grep -E "$LIMIT_LINE_REGEX" $PROC_LIMIT_CONF 1>&2
```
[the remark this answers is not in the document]
v-xecamp: 821 is to actually display all the entries in the message.  Other options include:
1.  Leaving it out, as the user can look, though the problem here is without literally stating the problem, more time may be spent addressing it.
2.  log_info $(grep -E "$LIMIT_LINE_REGEX" $PROC_LIMIT_CONF)

installer/scripts/omsadmin.sh
```diff
@@ -787,6 +804,36 @@ show_workspace_status()
...
+        log_warning "End of the $count duplicate entry WARNING Message"
+        llimit=`grep -E "$LIMIT_LINE_REGEX" $PROC_LIMIT_CONF | tail -1 | awk '{print $4}'`
```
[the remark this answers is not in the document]
v-xecamp: Sure.

installer/scripts/omsadmin.sh
```diff
@@ -787,6 +804,36 @@ show_workspace_status()
...
+    fi
+
+    if [ "$llimit" -eq -1 ]; then
```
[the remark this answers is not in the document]
v-xecamp: Yep.

installer/scripts/omsadmin.sh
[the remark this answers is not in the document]
v-xecamp: There is a case, must needs be by hand, that the user could intentionally code the configuration with an explicit -1.  This would pick that up, but the other would not.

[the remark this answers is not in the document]
v-xecamp: Also, the other tells the state of the maintenance, and this tells the behavior.

v-xecamp: Ok.

installer/scripts/omsadmin.sh
v-xecamp: Good point.

installer/scripts/omsadmin.sh
v-xecamp: Ok.  I'll add that.  Sounds good.

installer/scripts/omsadmin.sh
v-xecamp: So in that case, I will also make the initial case result_limit='', as that will fully demonstrate the intention of the routine, and then the single spot of the conditional on -1 will indicate the special case.

test/installer/scripts/omsadmin_systest.rb
v-xecamp: Yes, it only happens when there is none for $AGENT_USER existing, so I think it needs to be more like:  test_set_new_proc_limit_anticipate_end_label.  The existing case does not care about ending labels.

test/installer/scripts/omsadmin_systest.rb
v-xecamp: Good point.  It's a product of my back and forth learning I fear.

test/installer/scripts/omsadmin_systest.rb
v-xecamp: Ok.

test/installer/scripts/omsadmin_systest.rb
v-xecamp: Ok.

installer/scripts/omsadmin.sh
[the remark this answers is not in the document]
lagalbra: What I meant was, What I meant was, move the char "c" to the "Nrn" part of the sequence.

installer/scripts/omsadmin.sh
```diff
@@ -787,6 +804,40 @@ show_workspace_status()
...
 
+show_omsagent_proc_limit()
+{
+    local result_limit=
+
+    if [ ! -f $PROC_LIMIT_CONF ]; then
+        log_error "Limits configuration file '$PROC_LIMIT_CONF' missing!"
+        return 1
+    fi
+    local count=`cat $PROC_LIMIT_CONF | grep -E "$LIMIT_LINE_REGEX" | wc -l`
```
lagalbra: I just remembered; there is already a helper function [find_current_omsagent_proc_limit](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/scripts/omsadmin.sh#L405) which cats the file and greps it, but does not count the number of lines. Please use that function in place of all of the cat/grep lines.

installer/scripts/omsadmin.sh
```diff
@@ -787,6 +804,40 @@ show_workspace_status()
...
+    if [ "$count" -eq "0" ]; then
+        log_info "Process limit not currently set for $AGENT_USER."
+    elif [ "$count" -eq "1" ]; then
+        result_limit=`cat $PROC_LIMIT_CONF | grep -E "$LIMIT_LINE_REGEX" | awk '{print $4}'`
+    else
+        log_info "$count $AGENT_USER entries found in $PROC_LIMIT_CONF.  This may yield problems.  Last entry should apply."
+        log_info ">>Last entry presumed to apply<<"
+        local limit_lines=`cat $PROC_LIMIT_CONF | grep -E "$LIMIT_LINE_REGEX"`
+        log_info $limit_lines
+        log_info "End of the $count duplicate entry WARNING Message"
```
lagalbra: This is now an info message; change to "End of the $count duplicate entry INFO Message"

test/installer/scripts/omsadmin_systest.rb
```diff
@@ -296,4 +310,28 @@ def test_reconstruct_from_partial_empty_omsadmin_conf
...
+  def test_show_omsagent_proc_limit_default
+    set_omsagent_proc_limit()
+    lso = show_omsagent_proc_limit()
+    assert_match(/^-?\d+$/, lso, "The command to show what should be the default proc limit did not yield an integer.")
+  end
+
+  def test_show_omsagent_proc_limit_5000
+    set_omsagent_proc_limit(5000)
+    lso = show_omsagent_proc_limit()
+    assert_match(/^5000$/, lso, "Should have been 5000 process limit.")
```
lagalbra: It would be more explicit to change this to `assert_equal("5000", lso, "Should have been 5000 process limit.")` since you know the exact string that should get printed out; same for `test_show_omsagent_proc_limit_two_entries`

test/installer/scripts/omsadmin_systest.rb
```diff
@@ -296,4 +310,28 @@ def test_reconstruct_from_partial_empty_omsadmin_conf
...
+  end
+
+  def test_show_omsagent_proc_limit_two_entries
+    FileUtils.cp("#{@omsadmin_test_dir}/limits-two-settings.conf", @proc_limits_conf)
+    lso = show_omsagent_proc_limit()
+    assert_match(/^100$/, lso, "Multiple entries should still report last entry, in compliance with specification.")
```
lagalbra: Add a line to check that the "2 $TEST_USER entries found" message is printed.

test/installer/scripts/omsadmin_systest.rb
```diff
@@ -296,4 +310,28 @@ def test_reconstruct_from_partial_empty_omsadmin_conf
...
+  end
+
+  def test_show_omsagent_proc_limit_no_entry
+    FileUtils.cp("#{@omsadmin_test_dir}/limits-no-settings.conf", @proc_limits_conf)
+    lso = show_omsagent_proc_limit()
+    assert_equal("", lso, "The integer associated with unlimited is -1, but we return blank for reporting.")
```
lagalbra: Add a check that the "Process limit not currently set for $TEST_USER." message is printed

test/installer/scripts/omsadmin_systest.rb
v-xecamp: Yep.

installer/scripts/omsadmin.sh
v-xecamp: Okay, so this is in my latest branch check in.  I haven't squashed yet, but tell me if you think we are getting close, then I will.

installer/scripts/omsadmin.sh
v-xecamp: Yes, of course.

test/installer/scripts/omsadmin_systest.rb
v-xecamp: Sorry I missed this Friday.  Looking now.

installer/scripts/omsadmin.sh
```diff
@@ -161,6 +161,9 @@ usage()
     echo
     echo "Detect if omiserver is listening to SCOM port:"
     echo "$basename -o"
+    echo
+    echo "Show the configured process limit in /etc/limits.conf for user $AGENT_USER."
```
lagalbra: Use the `PROC_LIMIT_CONF` variable here, especially since this is not the correct path to look for the limit in.

installer/scripts/omsadmin.sh
```diff
@@ -239,7 +242,7 @@ parse_args()
 {
     local OPTIND opt
 
-    while getopts "h?s:w:d:vp:u:a:lx:XUm:Nrn:oR" opt; do
+    while getopts "h?s:w:d:vp:u:a:lx:XUm:Nrn:oRc" opt; do
```
lagalbra: nit: it's a very small thing, but what I had previously thought was that the "c" should be included with N, r, and n which also pertain to the process limit; like so: `...:Nrcn:...` . The "c" char cannot be right before a colon, because that indicates that it needs to have a following parameter (like n does for setting the limit).

installer/scripts/omsadmin.sh
```diff
@@ -161,6 +161,9 @@ usage()
     echo
```
lagalbra: I think it would be clearer if the added echo lines were here, with the rest of the echo lines about the process limit

installer/scripts/omsadmin.sh
```diff
@@ -381,18 +398,18 @@ set_omsagent_proc_limit()
 
     log_info "Setting process limit for the $AGENT_USER user in $PROC_LIMIT_CONF to $NEW_OMSAGENT_PROC_LIMIT..."
     local new_omsagent_line="$AGENT_USER  hard  nproc  $NEW_OMSAGENT_PROC_LIMIT"
-    find_current_omsagent_proc_limit
+    get_current_omsagent_proc_limit >/dev/null 2>/dev/null
```
lagalbra: Redirecting stdout and stderr is standardized in this file as `> /dev/null 2>&1`; do not use the format `>/dev/null 2>/dev/null`

installer/scripts/omsadmin.sh
```diff
@@ -404,7 +421,14 @@ unset_omsagent_proc_limit()
 
 find_current_omsagent_proc_limit()
```
lagalbra: Since you've removed all usage of this method, you should delete it.

installer/scripts/omsadmin.sh
```diff
@@ -796,6 +820,36 @@ show_workspace_status()
...
+        if [ "$count" -gt "1" ]; then
+            log_info "$count $AGENT_USER entries found in $PROC_LIMIT_CONF."
+            log_info ">>Last entry should apply<<" # NOTE:  This will not pick up -1 entries.
+            get_current_omsagent_proc_limit
+            log_info "End of the $count duplicate entry info Message"
+        fi
+        result_limit=`get_current_omsagent_proc_limit | tail -1 | awk '{print $4}'`
+        if [ "$result_limit" -eq "-1" ]; then
+            result_limit="No Limit"
+            log_info "There is NO present limit to number of processes for $AGENT_USER."
```
lagalbra: I think the messaging for this case should be slightly different than the messaging on line 833 when there are no lines in limits.conf for omsagent, since this case represents that a limit is set, but the "limit" is infinity. Perhaps "The number of allowed processes for $AGENT_USER is unlimited"?

test/installer/scripts/omsadmin_systest.rb
```diff
@@ -296,4 +310,33 @@ def test_reconstruct_from_partial_empty_omsadmin_conf
...
+    l = linearray.length
+    v = linearray[l-1].chomp
+    assert_equal(6, l, "This example should have six output lines from stdout.")
+    assert_equal("100", v, "Multiple entries should still report last entry, in compliance with specification.")
+  end
+
+  def test_show_omsagent_proc_limit_no_entry
+    FileUtils.cp("#{@omsadmin_test_dir}/limits-no-settings.conf", @proc_limits_conf)
+    lso = show_omsagent_proc_limit()
+    assert_match(/info/, lso, "The integer associated with unlimited is -1, but we return only logging.")
```
lagalbra: You should also check that "error" does not match in the output.

test/installer/scripts/omsadmin_systest.rb
v-xecamp: Got that too.  BTW I realized I changed these in my cleanup bout last week too, so the messages try to be more grammatically clear and more correct at the same time now.  I think this will be okay, but sorry that's one more thing for you to review.

installer/scripts/omsadmin.sh
v-xecamp: Got it.

installer/scripts/omsadmin.sh
v-xecamp: That's why I like alphabetical.  Much of reference needs are primarily around finding things, rather than buttressing meaning.  However, in this case yes makes good sense given how we do it.  I'd say since reporting is often most upfront, I'd say cNrn might be fine.  I'll do that, but tell me if you can think of a better one.

installer/scripts/omsadmin.sh
v-xecamp: Ok.

installer/scripts/omsadmin.sh
v-xecamp: Ok.

test/installer/scripts/omsadmin_systest.rb
v-xecamp: Yes, I already changed that out just a bit ago.  I'll add the error check.  Thanks.

v-xecamp: Please see that I added return of "No Limit" rather than nothing, because I not only found it looked puzzling and assymetric, but I felt it might actually be confusing to not have a parallel value.  You're final call of course.

installer/scripts/omsadmin.sh
lagalbra: This method is still here; are you still making changes, or do you want to keep it?

installer/scripts/omsadmin.sh
```diff
@@ -796,6 +820,37 @@ show_workspace_status()
...
+        if [ "$count" -gt "1" ]; then
+            log_info "$count $AGENT_USER entries found in $PROC_LIMIT_CONF."
+            log_info ">>Last entry should apply<<" # NOTE:  This will not pick up -1 entries.
+            get_current_omsagent_proc_limit
+            log_info "End of the $count duplicate entry info Message"
+        fi
+        result_limit=`get_current_omsagent_proc_limit | tail -1 | awk '{print $4}'`
+        if [ "$result_limit" -eq "-1" ]; then
+            result_limit="No Limit"
+            #log_info "There is NO present limit to number of processes for $AGENT_USER."
```
lagalbra: You should delete this line entirely instead of just commenting it out.

test/installer/scripts/omsadmin_systest.rb
```diff
@@ -296,4 +310,45 @@ def test_reconstruct_from_partial_empty_omsadmin_conf
...
+    assert_equal("5000", nostr, "Should have been 5000 process limit.")
+  end
+
+  def test_show_omsagent_proc_limit_two_entries
+    FileUtils.cp("#{@omsadmin_test_dir}/limits-two-settings.conf", @proc_limits_conf)
+    lso = show_omsagent_proc_limit()
+    linearray = lso.split("\n")
+    l = linearray.length
+    m = linearray[0].chomp
+    v = linearray[l-1].chomp
```
lagalbra: nit: these single-char variables could be named more descriptively; reading the code at the moment does not immediately reveal what their purpose is.

installer/scripts/omsadmin.sh
v-xecamp: I don't know why I didn't see your EXACT sequence the first time.  If there are any more changes, I'll make it like that, but for now I have a working build.

test/installer/scripts/omsadmin_systest.rb
```diff
@@ -296,4 +310,45 @@ def test_reconstruct_from_partial_empty_omsadmin_conf
...
+    FileUtils.cp("#{@omsadmin_test_dir}/limits-no-settings.conf", @proc_limits_conf)
+    lso = show_omsagent_proc_limit()
+    linearray = lso.split("\n")
+    l = linearray.length
+    m = linearray[0].chomp
+    v = linearray[l-1].chomp
+    assert_equal(2, l, "This example should have six output lines from stdout.")
+    expectedline0 = "There is NO present limit to number of processes for #{TEST_USER}."
+    assert_not_match(/error/i,m,"Should not be an error")
+    assert_match(/#{expectedline0}/,m,"Expected first line message does not match that output.")
```
lagalbra: nit: use spaces in the parameter lists

installer/scripts/omsadmin.sh
v-xecamp: Got it.
I don't get a comment for the item on find_current_*() for some reason.  I'll take it out as your question implies it is not yet used outside the script.  Thanks.

installer/scripts/omsadmin.sh
v-xecamp: Taking it out.

installer/scripts/omsadmin.sh
```diff
@@ -161,6 +161,9 @@ usage()
     echo
     echo "Detect if omiserver is listening to SCOM port:"
     echo "$basename -o"
+    echo
+    echo "Show the configured process limit in $PROC_LIMIT_CONF for user $AGENT_USER."
+    echo "$basename -c"
```
lagalbra: Moving this up next to the other process limit lines?

lagalbra:
@v-xecamp Can you please confirm all the testing that you have performed with the changes including your latest commit?

PR #573, opened 2017-09-11
Change service_control return values for caller needs.  

v-xecamp:
Intermediate state with traces in bundle_skel.sh

Validate state of omsadmin.conf.

Modify omsadmin.conf validation.

### DIFF
--- a/installer/scripts/service_control
+++ b/installer/scripts/service_control
@@ -22,6 +22,8 @@
 OMSAGENT_START_WAIT_MAX=18
 OMSAGENT_STOP_WAIT_MAX=15
 STOP_PROCESS_WAIT_MAX=5
+WAIT_UNTIL_RUNNING_ITERATION_SLEEP=0.5
+WAIT_UNTIL_STOPPED_ITERATION_SLEEP=0.5
 
 # Space seperated list of non oms workspaces
 NON_OMS_WS="scom"
@@ -33,12 +35,19 @@ BIN_DIR=$COMMONSEGMENT/bin
 ETC_DIR=/etc$COMMONSEGMENT
 VAR_DIR=/var$COMMONSEGMENT
 
-# Error codes and categories:
+# Error codes
 
 # User configuration/parameters:
 
-INVALID_OPTION_PROVIDED=2
-NO_OPTION_PROVIDED=3
+INVALID_OPTION_PROVIDED=72
+NO_OPTION_PROVIDED=75
+RUN_AS_ROOT=77
+
+# Service-related:
+
+ERROR_OS_FACILITY_NOT_PRESENT=67
+ERROR_UNEXPECTED_SYSTEM_INFO=68
+ERROR_UNEXPECTED_STATE=69
 
 # Internal errors:
 
@@ -46,34 +55,30 @@ INTERNAL_ERROR=30
 
 # Files and I/O
 
-FILE_NOT_FOUND=1
-FILE_NOT_GONE=2
+FILE_NOT_FOUND=21
+FILE_NOT_GONE=22
 
-# 
+#  End of error codes
+
 REGEX_UUID='[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}'
 WORKSPACE_REGEX="^$REGEX_UUID$"
 
 # Begin Lower Level Utilities
 
-validate_file_has_line_pattern()
+check_file_has_line_pattern()
 {
-    # Does regular expression check for entire line.  As a validaytor routine,
-    # it exits with a non-zero value if the validation failes.  Specifying any
+    # Does regular expression check for entire line.  As a validator routine,
+    # it exits with a non-zero value if the validation fails.  Specifying any
     # third argument stifles messaging.
     local _fSpec=$1
     local _pattA="$2"
-    local _boolOnly=$3
     cat $_fSpec | grep -Eq "^$_pattA$"
     if [ $? -eq 0 ]; then
         return 0
     fi
-    if [ -z "$_boolOnly" ]; then
-        echo "INTERNAL ERROR:  $_fSpec missing required pattern '^$_pattA\$'."
-    fi
-    exit $INTERNAL_ERROR
+    echo "INTERNAL ERROR:  $_fSpec missing required pattern '^$_pattA\$'." >&2
+    return $INTERNAL_ERROR
 }
-
-# End Lower Level Utilities
 
 # Begin FailCount Singleton Interface
 
@@ -87,18 +92,23 @@ fcso_increment()
     if [ $# -ne 3 ]; then
         echo "INTERNAL ERROR: Incorrect number of parameters passed to fcso_increment" >&2
         echo "$# arguments seen." >&2
-        exit $INTERNAL_ERROR
+        notification_exit $INTERNAL_ERROR
     fi
     local _actStr=$1
     local _wId=$2
     local _sCode=$3
+    local _failError="$4"
 
     if [ "$_sCode" -eq "0" ]; then
         return
     else
         FCSO_FailCount=`expr $FCSO_FailCount + 1`
     fi
-    echo "ERROR:  $_actStr failed with result '$_sCode' on WS Id $_wId."
+    if [ $# -gt 3 ]; then
+        echo "$_failError" >&2
+    else
+        echo "ERROR:  $_actStr failed with result '$_sCode' on workspace $_wId." >&2
+    fi
 }
 
 fcso_new()
@@ -107,6 +117,22 @@ fcso_new()
 }
 
 # End FailCount Singleton Interface
+
+notification_exit()
+{
+    local _exitStatus=$1
+
+    if [ -n `echo $_exitStatus | grep '[0-9][0-9]*'` ]; then
+        echo "service_control exiting with code $_exitStatus"
+        exit $_exitStatus
+    else
+        echo "INTERNAL ERROR:  service_control notification_exit routine NOT USED Correctly."
+        echo "Exiting with code $INTERNAL_ERROR"
+        exit $INTERNAL_ERROR
+    fi
+}
+
+# End Lower Level Utilities
 
 setup_variables()
 {
@@ -118,54 +144,58 @@ setup_variables()
         initial_conf_dir="$ETC_DIR/$1/conf/omsadmin.conf"
     fi
 
-    validate_omsadmin_conf $initial_conf_dir
-    . $initial_conf_dir
-
-    VAR_DIR_WS=$VAR_DIR/$WORKSPACE_ID
-    ETC_DIR_WS=$ETC_DIR/$WORKSPACE_ID
-
-    TMP_DIR=$VAR_DIR_WS/tmp
-    STATE_DIR=$VAR_DIR_WS/state
-    RUN_DIR=$VAR_DIR_WS/run
-    LOG_DIR=$VAR_DIR_WS/log
-    CERT_DIR=$ETC_DIR_WS/certs
-    CONF_DIR=$ETC_DIR_WS/conf
-
-    LOGFILE=$LOG_DIR/omsagent.log
-    PIDFILE=$RUN_DIR/omsagent.pid
-    CONFFILE=$CONF_DIR/omsagent.conf
-    OMSADMIN_CONF=$CONF_DIR/omsadmin.conf
-
-    OMSAGENT_WS=omsagent-$WORKSPACE_ID
-
     WS_STATUS=0
+    check_omsadmin_conf $initial_conf_dir
+    WS_STATUS=$?
+    if [ "$WS_STATUS" -eq "0" ]; then
+        . $initial_conf_dir
+        WS_STATUS=$?
+    fi
+
+    if [ "$WS_STATUS" -eq "0" ]; then
+
+        VAR_DIR_WS=$VAR_DIR/$WORKSPACE_ID
+        ETC_DIR_WS=$ETC_DIR/$WORKSPACE_ID
+
+        TMP_DIR=$VAR_DIR_WS/tmp
+        STATE_DIR=$VAR_DIR_WS/state
+        RUN_DIR=$VAR_DIR_WS/run
+        LOG_DIR=$VAR_DIR_WS/log
+        CERT_DIR=$ETC_DIR_WS/certs
+        CONF_DIR=$ETC_DIR_WS/conf
+
+        LOGFILE=$LOG_DIR/omsagent.log
+        PIDFILE=$RUN_DIR/omsagent.pid
+        CONFFILE=$CONF_DIR/omsagent.conf
+        OMSADMIN_CONF=$CONF_DIR/omsadmin.conf
+
+        OMSAGENT_WS=omsagent-$WORKSPACE_ID
+    fi
+    return $WS_STATUS
 }
 
-validate_omsadmin_conf()
+check_omsadmin_conf()
 {
     # Checks for a series of identifier assignment lines which should always be present
     # for an onboarded workspace to be complete and healthy.
-    local _fSpec=$1
-    local _boolOnly=$2
+    local _fs=$1
+    local ec=0
     VALIDATION_SUPPLEMENT_MESSAGE="Onboarding incorrect or incomplete."
-    if [ -f $_fSpec ]; then
-        validate_file_has_line_pattern $_fSpec "WORKSPACE_ID=$REGEX_UUID" $_boolOnly
-        validate_file_has_line_pattern $_fSpec "AGENT_GUID=$REGEX_UUID" $_boolOnly
-        validate_file_has_line_pattern $_fSpec "LOG_FACILITY=.*" $_boolOnly
-        validate_file_has_line_pattern $_fSpec "CERTIFICATE_UPDATE_ENDPOINT=.*" $_boolOnly
-        validate_file_has_line_pattern $_fSpec "URL_TLD=.*" $_boolOnly
-        validate_file_has_line_pattern $_fSpec "DSC_ENDPOINT=.*" $_boolOnly
-        validate_file_has_line_pattern $_fSpec "OMS_ENDPOINT=.*" $_boolOnly
-        validate_file_has_line_pattern $_fSpec "AZURE_RESOURCE_ID=.*" $_boolOnly
-        validate_file_has_line_pattern $_fSpec "OMSCLOUD_ID=.*" $_boolOnly
-        validate_file_has_line_pattern $_fSpec "UUID=$REGEX_UUID" $_boolOnly
-        return 0
+    if [ -f $_fs ]; then
+        if ! check_file_has_line_pattern $_fs "WORKSPACE_ID=$REGEX_UUID";       then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "AGENT_GUID=$REGEX_UUID";         then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "LOG_FACILITY=.*";                then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "CERTIFICATE_UPDATE_ENDPOINT=.*"; then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "URL_TLD=.*";                     then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "DSC_ENDPOINT=.*";                then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "OMS_ENDPOINT=.*";                then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "AZURE_RESOURCE_ID=.*";           then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "OMSCLOUD_ID=.*";                 then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "UUID=$REGEX_UUID";               then ec=$(expr $ec + 1); fi
+        return $ec
     else
-        if [ -z "$_boolOnly" ]; then
-            echo "INTERNAL_ERROR:  OMS Admin configuration file $_fSpec NOT Found!" >&2
-            echo $VALIDATION_SUPPLEMENT_MESSAGE
-        fi
-        exit $INTERNAL_ERROR
+        echo "ERROR:  OMS Admin configuration file $_fs NOT Found!" >&2
+        return $FILE_NOT_FOUND
     fi
 }
 
@@ -173,7 +203,7 @@ verify_privileges()
 {
     if [ `id -u` -ne 0 ]; then
         echo "Must have root privileges for this operation" >& 2
-        exit 1
+        notification_exit $RUN_AS_ROOT
     fi
 }
 
@@ -181,8 +211,27 @@ this_omsagent_running()
 {
     WS_STATUS=0
     if [ -f $PIDFILE ]; then
-        ps -p `cat $PIDFILE` | grep -q omsagent
-        WS_STATUS=$?
+        local omsagent_pid=`cat $PIDFILE 2>/dev/null`
+        local ps_state=`ps --no-header -o state -p $omsagent_pid`
+        if [ -z "$ps_state" ]; then
+            WS_STATUS=1 # Not There; FALSE
+        else
+            case "$ps_state" in
+            D)  echo "Uninterruptable Sleep State Seen in omsagent process.";;
+            R)  ;;
+            S)  ;;
+            T)  echo "Stopped State Seen in omsagent process."
+                WS_STATUS=$ERROR_UNEXPECTED_STATE;;
+            W)  echo "Paging State Seen in omsagent process."
+                WS_STATUS=$ERROR_UNEXPECTED_STATE;;
+            X)  echo "Dead State Seen in omsagent process."
+                WS_STATUS=$ERROR_UNEXPECTED_STATE;;
+            Z)  echo "Defunct State Seen in omsagent process."
+                WS_STATUS=$ERROR_UNEXPECTED_STATE;;
+            *)  echo "ERROR:  '$ps_state' is not a known ps flag."
+                notification_exit ERROR_UNEXPECTED_SYSTEM_INFO;;
+            esac
+        fi
     else
         WS_STATUS=$FILE_NOT_FOUND
     fi
@@ -206,7 +255,9 @@ is_omsagent_running()
 
 remove_PIDfile_unless_omsagent_running()
 {
-
+    # IMPORTANT:  Remember this procedure does not determine if omsagent is
+    # running, but only clears out the PIDFILE when it is not.  It may be
+    # successful in both a running and a non-running state.
     if this_omsagent_running; then
         return 0
     fi
@@ -218,33 +269,27 @@ remove_PIDfile_unless_omsagent_running()
     return $WS_STATUS
 }
 
-exit_if_agent_not_onboarded()
-{
-    if [ ! -f $OMSADMIN_CONF ]; then
-        echo "FATAL: Agent is not onboarded. omsagent cannot be registered as a service."
-        WS_STATUS=$FILE_NOT_FOUND
-        exit $WS_STATUS
-    fi
-}
-
 wait_until_omsagent_is_running()
 {
     # Required parameter: Number of seconds to wait for agent to become running.
-    if [ -z "$1" -o "$1" -le 0 ]; then
-        echo "Function \"wait_until_omsagent_stops\" called with invalid parameter"
-        WS_STATUS=1
-        exit $WS_STATUS
+    if [ -z "$1" -o "$1" -le "0" ]; then
+        echo "INTERNAL ERROR:  Function \"wait_until_omsagent_is_running\" called with invalid parameter"
+        WS_STATUS=$INTERNAL_ERROR
+        notification_exit $WS_STATUS
     fi
 
     COUNTER=$(( $1 * 2 )) # Since we sleep 0.5 seconds, compute number of seconds
     while [ $COUNTER -gt 0 ]; do
         if ! this_omsagent_running; then
             COUNTER=$(( $COUNTER - 1 ))
-            sleep 0.5
+            sleep $WAIT_UNTIL_RUNNING_ITERATION_SLEEP
         else
             return 0
         fi
     done
+    if this_omsagent_running; then
+        return 0
+    fi
 
     return 1 # Never came up.
 }
@@ -252,21 +297,25 @@ wait_until_omsagent_is_running()
 wait_until_omsagent_stopped()
 {
     # Required parameter: Number of seconds to wait for agent to stop
-    if [ -z "$1" -o "$1" -le 0 ]; then
-        echo "INTERNAL ERROR:  Function \"wait_until_omsagent_stops\" called with invalid parameter"
-        exit $INTERNAL_ERROR
+    if [ -z "$1" -o "$1" -le "0" ]; then
+        echo "INTERNAL ERROR:  Function \"wait_until_omsagent_stopped\" called with invalid parameter"
+        notification_exit $INTERNAL_ERROR
     fi
 
     COUNTER=$(( $1 * 2 )) # Since we sleep 0.5 seconds, compute number of seconds
     while [ $COUNTER -gt 0 ]; do
         if this_omsagent_running; then
             COUNTER=$(( $COUNTER - 1 ))
-            sleep 0.5
+            sleep $WAIT_UNTIL_STOPPED_ITERATION_SLEEP
         else
             remove_PIDfile_unless_omsagent_running
             return $? # Should only be non-zero if a PIDFILE remains.
         fi
     done
+    if ! this_omsagent_running; then
+        remove_PIDfile_unless_omsagent_running
+        return 0
+    fi
 
     return 1 # Never stopped.
 }
@@ -289,31 +338,68 @@ stop_omsagent_process()
     return $WS_STATUS
 }
 
-    #  XC on 20170824; Laura says this is no longer used:
-start_omsagent_process()
+generateProcedureName()
 {
-    if remove_PIDfile_unless_omsagent_running; then
-        $BIN_DIR/omsagent -d $PIDFILE --no-supervisor -o $LOGFILE -c $CONFFILE
-    fi
-    WS_STATUS=$?
-    return $WS_STATUS
+    case "$1" in
+        disable)    echo "disable_omsagent_service";;
+        enable)     echo "enable_omsagent_service";;
+        reload)     echo "restart_omsagent";;
+        restart)    echo "restart_omsagent";;
+        start)      echo "start_omsagent";;
+        stop)       echo "stop_omsagent";;
+        *)          echo "INTERNAL ERROR:  '$1' is not a valid procedural activity."
+                    notification_exit INTERNAL_ERROR;;
+    esac
+}
+
+# ## Iterators #### 
+
+check_non_oms_and_invoke()
+{
+    # Space seperated list of non oms workspaces
+    NON_OMS_WS="scom"
+
+    local _cmdArg=$1
+    local ws_id
+    for ws_id in $NON_OMS_WS
+    do
+        ls -1 $ETC_DIR | grep -w ${ws_id} >/dev/null 2>&1
+        if [ $? -eq 0 ]; then
+            if setup_variables $ws_id; then
+                $_cmdArg
+            fi
+            fcso_increment $_cmdArg $ws_id $WS_STATUS
+        fi
+    done
+}
+
+check_oms_and_invoke()
+{
+    local _cmdArg=$1
+    local ws_id
+    for ws_id in `ls -1 $ETC_DIR | grep -E $WORKSPACE_REGEX`
+    do
+        if setup_variables $ws_id; then
+            $_cmdArg
+        fi
+        fcso_increment $_cmdArg $ws_id $WS_STATUS
+    done
 }
 
 #
-# Normal usage functions (used by everything except service control scripts)
+# Main Interface Procedures:  Those Major Functions called from Main
 #
 
 start_omsagent()
 {
-    exit_if_agent_not_onboarded
     enable_omsagent_service
     remove_PIDfile_unless_omsagent_running
-    if [ $WS_STATUS -ne 0 ]; then
+    if [ "$WS_STATUS" -ne "0" ]; then
         return $WS_STATUS
     fi
 
     # If systemd lives here, then we have a systemd unit file
-    if pidof systemd 1> /dev/null 2> /dev/null; then
+    if pidof systemd 1>/dev/null 2>&1; then
         /bin/systemctl start $OMSAGENT_WS
     else
         if [ -x /usr/sbin/invoke-rc.d ]; then
@@ -324,11 +410,11 @@ start_omsagent()
             /bin/systemctl start $OMSAGENT_WS
         else
             echo "ERROR:  Unrecognized service controller to start OMS Agent service" 1>&2
-            exit 1
+            notification_exit $ERROR_OS_FACILITY_NOT_PRESENT
         fi
     fi
     WS_STATUS=$?
-    if [ $WS_STATUS -eq 0 ]; then
+    if [ "$WS_STATUS" -eq "0" ]; then
         wait_until_omsagent_is_running $OMSAGENT_START_WAIT_MAX
         WS_STATUS=$?
     fi
@@ -340,7 +426,7 @@ stop_omsagent()
     remove_PIDfile_unless_omsagent_running
     if this_omsagent_running; then
         # If systemd lives here, then we have a systemd unit file
-        if pidof systemd 1> /dev/null 2> /dev/null; then
+        if pidof systemd 1> /dev/null 2>&1; then
             /bin/systemctl stop $OMSAGENT_WS
         else
             if [ -x /usr/sbin/invoke-rc.d ]; then
@@ -351,7 +437,7 @@ stop_omsagent()
                 /bin/systemctl stop $OMSAGENT_WS
             else
                 echo "ERROR:  Unrecognized service controller to stop OMS Agent service" 1>&2
-                exit 1
+                notification_exit $ERROR_OS_FACILITY_NOT_PRESENT
             fi
         fi
         WS_STATUS=$?
@@ -359,7 +445,7 @@ stop_omsagent()
         WS_STATUS=0
         return 0
     fi
-    if [ $WS_STATUS -eq 0 ]; then
+    if [ "$WS_STATUS" -eq "0" ]; then
         wait_until_omsagent_stopped $OMSAGENT_STOP_WAIT_MAX
         WS_STATUS=$?
     fi
@@ -368,8 +454,6 @@ stop_omsagent()
 
 restart_omsagent()
 {
-    exit_if_agent_not_onboarded
-
     remove_PIDfile_unless_omsagent_running
     if this_omsagent_running; then
         enable_omsagent_service
@@ -379,7 +463,7 @@ restart_omsagent()
     fi
 
     # If systemd lives here, then we have a systemd unit file
-    if pidof systemd 1> /dev/null 2> /dev/null; then
+    if pidof systemd 1>/dev/null 2>&1; then
         /bin/systemctl restart $OMSAGENT_WS
     else
         if [ -x /usr/sbin/invoke-rc.d ]; then
@@ -390,11 +474,11 @@ restart_omsagent()
             /bin/systemctl restart $OMSAGENT_WS
         else
             echo "ERROR:  Unrecognized service controller to restart OMS Agent service" 1>&2
-            exit 1
+            notification_exit $ERROR_OS_FACILITY_NOT_PRESENT
         fi
     fi
     WS_STATUS=$?
-    if [ $WS_STATUS -eq 0 ]; then
+    if [ "$WS_STATUS" -eq "0" ]; then
         wait_until_omsagent_is_running $OMSAGENT_START_WAIT_MAX
         WS_STATUS=$?
     fi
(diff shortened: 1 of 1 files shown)


installer/bundle/bundle_skel.sh
```diff
@@ -1000,6 +1000,7 @@ case "$installMode" in
         ;;
 
     U)
+        echo "trace 0 installMode U" >&2
```
lagalbra: Can you explain why you want to commit these trace lines? They're great for development debugging, but may not be appropriate for public release.

installer/scripts/service_control
```diff
@@ -20,29 +23,75 @@ VAR_DIR=/var/opt/microsoft/omsagent
 ETC_DIR=/etc/opt/microsoft/omsagent
 BIN_DIR=/opt/microsoft/omsagent/bin
 
-WORKSPACE_REGEX='^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$'
+REGEX_UUID='[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}'
```
lagalbra: For style consistency, can you use double quotes around this string?

installer/scripts/service_control
```diff
@@ -20,29 +23,75 @@ VAR_DIR=/var/opt/microsoft/omsagent
...
-{
+# Begin Lower Level Utilities
+
+validate_file_has_line_pattern() {
+	local _fSpec=$1
+	local _pattA="$2"
+	local _boolOnly=$3
+	cat $_fSpec | grep -Eq "^$_pattA$"
+	if [ $? -eq 0 ]; then
+		return 0
```
lagalbra: Please use four spaces instead of tabs for indentation.

installer/scripts/service_control
```diff
@@ -20,29 +23,75 @@ VAR_DIR=/var/opt/microsoft/omsagent
...
 
-setup_variables()
-{
+# Begin Lower Level Utilities
+
+validate_file_has_line_pattern() {
+	local _fSpec=$1
+	local _pattA="$2"
+	local _boolOnly=$3
+	cat $_fSpec | grep -Eq "^$_pattA$"
```
lagalbra: If the pattern is matched, this should not print anything to stdout; can you validate that nothing gets printed out on this line?

installer/scripts/service_control
```diff
@@ -20,29 +23,75 @@ VAR_DIR=/var/opt/microsoft/omsagent
...
+	if [ $? -eq 0 ]; then
+		return 0
+	fi
+	if [ -z "$_boolOnly" ]; then
+		echo "$VALIDATION_SEVERITY:  $_fSpec missing required pattern '$_pattA'."
+        echo $VALIDATION_SUPPLEMENT_MESSAGE
```
lagalbra: Fix indentation on this line; it's not aligned with the line above it.

installer/scripts/service_control
```diff
@@ -20,29 +23,75 @@ VAR_DIR=/var/opt/microsoft/omsagent
...
+VALIDATION_SEVERITY=WARNING
 
 # Space seperated list of non oms workspaces
 NON_OMS_WS="scom"
 
-setup_variables()
-{
+# Begin Lower Level Utilities
+
+validate_file_has_line_pattern() {
```
lagalbra: Can you add a line or two of explanation about how this method is used?

installer/scripts/service_control
```diff
@@ -20,29 +23,75 @@ VAR_DIR=/var/opt/microsoft/omsagent
...
+
+fcso_any_failures() {
+    return $FCSO_FailCount
+}
+
+fcso_increment() {
+    if [ $# -ne 3 ]; then
+        echo "FATAL:  Programmer error; 3 arguments required for fcso_increment!"
+        echo "$# arguments seen."
+        exit 99
```
lagalbra: In other files, this is an "INTERNAL_ERROR" case for us, and we have a specific exit code defined at the top of the file for it ([example](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/bundle/bundle_skel.sh#L58)). If you want to exit with a code other than 1, please define that number at the top of the file for clarity.

lagalbra: Also, you may want to mimic the message of the other example as well: `echo "INTERNAL ERROR: Incorrect number of parameters passed to fcso_increment" >&2`

installer/scripts/service_control
```diff
@@ -20,29 +23,75 @@ VAR_DIR=/var/opt/microsoft/omsagent
...
+    fi
+    local _actStr=$1
+    local _wId=$2
+    local _sCode=$3
+
+    if [ "$_sCode" -eq 0 ]; then
```
lagalbra: Both types should be string: `"$_sCode" -eq "0"`

installer/scripts/service_control
```diff
@@ -59,6 +108,33 @@ setup_variables()
     OMSADMIN_CONF=$CONF_DIR/omsadmin.conf
 
     OMSAGENT_WS=omsagent-$WORKSPACE_ID
+
+    WS_STATUS=0
+}
+
+validate_omsadmin_conf() {
```
lagalbra: Please use the method definition style with a newline before the opening bracket:
```
method()
{
...
}
```

installer/scripts/service_control
```diff
@@ -59,6 +108,33 @@ setup_variables()
...
+        validate_file_has_line_pattern $_fSpec "AZURE_RESOURCE_ID=.*" $_boolOnly
+        validate_file_has_line_pattern $_fSpec "OMSCLOUD_ID=.*" $_boolOnly
+        validate_file_has_line_pattern $_fSpec "UUID=$REGEX_UUID" $_boolOnly
+        return 0
+    else
+        if [ -z "$_boolOnly" ]; then
+            echo "WARNING:  OMS Admin configuration file $OMSADMIN_CONF NOT Found!" >&2
+            echo $VALIDATION_SUPPLEMENT_MESSAGE
+        fi
+        exit 2
```
lagalbra: Same comment about defining what exit codes mean earlier in file

installer/scripts/service_control
```diff
@@ -69,29 +145,53 @@ verify_privileges()
     fi
 }
 
+this_omsagent_running() {
+    if [ -f $PIDFILE ]; then
+        if ps -p `cat $PIDFILE` | grep -q omsagent
```
lagalbra: Can you make this check more explicit? I would rather have you execute the ps/grep command and then check $? like `is_omsagent_running()` did previously. As it is, it looks a little unclear.

installer/scripts/service_control
```diff
@@ -69,29 +145,53 @@ verify_privileges()
...
-
-    # Kill PID file if we're not running any longer
-    if [ $STATUS -ne 0 ]; then
-        rm -f $PIDFILE
-        return 0
-    else
-        return 1
-    fi
+	#### Keeping return value on this as it originally was.  I deduce
+    #### that because this value is returned with the is-running qualifier
```
lagalbra: Fix indentation on this line

installer/scripts/service_control
```diff
@@ -69,29 +145,53 @@ verify_privileges()
...
+	#### to service_control, that there are scripts in other languages
+	#### that use the more conventional values for booleans.  Meanwhile,
+	#### I recommend deprecation in all but the external interface.  XC
+    if this_omsagent_running; then
```
lagalbra: Fix indentation on this line

installer/scripts/service_control
```diff
@@ -100,18 +200,26 @@ wait_until_omsagent_stops()
...
     while [ $COUNTER -gt 0 ]; do
-        is_omsagent_running && return $?
-        COUNTER=$(( $COUNTER - 1 ))
-        sleep 0.5
+        if is_omsagent_running; then
+            COUNTER=$(( $COUNTER - 1 ))
+            sleep 0.5
+        else
+            remove_PIDfile_unless_omsagent_running
+            WS_STATUS=$? # Should only be non-zero of the removal of a PIDFILE failed for some reason.
```
lagalbra: The `remove_PIDfile_unless_omsagent_running` method is somewhat confusing; you say here that it should return 0 if the PIDFILE was correctly deleted, but you return 1 [here](https://github.com/Microsoft/OMS-Agent-for-Linux/pull/573/files#diff-c1d234f8111f79a335f4309684025d80R186) when the rm command returns 0. Please comment on and validate the expected behavior.

installer/scripts/service_control
```diff
@@ -69,29 +145,53 @@ verify_privileges()
     fi
 }
 
+this_omsagent_running() {
+    if [ -f $PIDFILE ]; then
+        if ps -p `cat $PIDFILE` | grep -q omsagent
+        then
```
lagalbra: Please use the same style throughout for your if/then lines; most of our code uses the same line:
```
if [ check ]; then
...
fi
```

v-xecamp:
I do not.  My apologies.  Those need to come out.

installer/scripts/service_control
v-xecamp: We could have a unit test for that.  I may not understand the question. -q is a standard grep switch.

installer/scripts/service_control
v-xecamp: I'll go through the entire file and fix these.  I realized it's from bringing in the routines from my unit-tested areas where I didn't have the python tabbing.

installer/scripts/service_control
v-xecamp: Good.  I'll follow that.  Sorry about the flub first time.  
BTW I was gonna suggest generally we move to logs always to stderr, and that is specifically an interesting question in omsadmin.sh.  Let's digress on that soon.

installer/scripts/service_control
lagalbra: Ah, I wasn't familiar with the grep -q switch. That's fine.

v-xecamp: I actually saw it first somewhere in this codebase.  It is useful under non-Bash because we need grep to do regular expressions.  In modern Bash there are regular expressions in the [[ ]] segments available.

installer/scripts/service_control
v-xecamp: Yep.  Thanks.

installer/scripts/service_control
v-xecamp: I think I got all those in this source file.  See next push.

installer/scripts/service_control
v-xecamp: ok

installer/scripts/service_control
v-xecamp: Sure.

installer/scripts/service_control
v-xecamp: Yep.

installer/scripts/service_control
```diff
@@ -69,49 +176,107 @@ verify_privileges()
...
 }
 
 exit_if_agent_not_onboarded()
 {
     if [ ! -f $OMSADMIN_CONF ]; then
-        echo "Warning: Agent is not onboarded. omsagent cannot be registered as a service."
-        exit 0 
+        echo "FATAL: Agent is not onboarded. omsagent cannot be registered as a service."
+        WS_STATUS=$FILE_NOT_FOUND
+        exit $WS_STATUS
```
lagalbra: This is another method that gets called in sequences for multiple workspaces, so I would advise against exiting rather than returning an error here.

installer/scripts/service_control
```diff
@@ -69,49 +176,107 @@ verify_privileges()
...
     fi
 }
 
+wait_until_omsagent_is_running()
+{
+    # Required parameter: Number of seconds to wait for agent to become running.
+    if [ -z "$1" -o "$1" -le 0 ]; then
```
lagalbra: integer comparison: use `$1 -le 0`
I would also check first that $# -eq 1, since you're expecting exactly one parameter

installer/scripts/service_control
```diff
@@ -69,49 +176,107 @@ verify_privileges()
...
+        echo "Function \"wait_until_omsagent_stops\" called with invalid parameter"
+        WS_STATUS=1
```
lagalbra: This would count as an internal error case, even though this file can be sourced and called with an invalid parameter. We're the only ones sourcing it.

installer/scripts/service_control
```diff
@@ -69,49 +176,107 @@ verify_privileges()
...
+        exit $WS_STATUS
```
lagalbra: Same multiple workspaces concern: this method should return, not exit

installer/scripts/service_control
```diff
@@ -69,49 +176,107 @@ verify_privileges()
...
+    fi
+
+    COUNTER=$(( $1 * 2 )) # Since we sleep 0.5 seconds, compute number of seconds
+    while [ $COUNTER -gt 0 ]; do
+        if ! this_omsagent_running; then
+            COUNTER=$(( $COUNTER - 1 ))
```
lagalbra: Do you have a preference between this method of arithmetic and using `expr`? I know a certain amount of our arithmetic code that has been in use for a while uses the former method (see [this example](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/scripts/omsadmin.sh#L312)). I would suggest using only only method, and I personally prefer the `$(())` method.

installer/scripts/service_control
```diff
@@ -69,49 +176,107 @@ verify_privileges()
...
+            sleep 0.5
```
lagalbra: Add this as a constant with the rest of your sleep times.

installer/scripts/service_control
```diff
@@ -69,49 +176,107 @@ verify_privileges()
...
+        else
+            WS_STATUS=0
+            return $WS_STATUS
+        fi
+    done
+
+    # One final try for accurate return status (just return status from the call)
+    remove_PIDfile_unless_omsagent_running
+    WS_STATUS=$? # Should only be non-zero of the removal of a PIDFILE failed for some reason.
```
lagalbra: As I mentioned before, this comment indicates that `remove_PIDfile_unless_omsagent_running` only returns non-zero if the PIDFILE was attempted to be removed, but that failed. This contradicts [your code](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/7dc009e88cea05af7603cda59c2c6a4aa1700a5e/installer/scripts/service_control#L213), which reads:
```
    rm -f $PIDFILE # NOTE:  definition of PIDFILE is dependent on WS.
    if [ $? -ne 0 ]; then
        WS_STATUS=2
        return $WS_STATUS
    fi
    return 1
```
If the `rm -f $PIDFILE` line returns 0, you return 1 from `remove_PIDfile_unless_omsagent_running`! If you want to use that method as a trinary state return (0 for omsagent_running, 1 for omsagent not running and PIDFILE removed, 2 for omsagent not running and PIDFILE not removed, than you need to change the rest of your code that uses its return value.

installer/scripts/service_control
```diff
@@ -69,49 +176,107 @@ verify_privileges()
...
+}
+
 wait_until_omsagent_stops()
 {
     # Required parameter: Number of seconds to wait for agent to stop
     if [ -z "$1" -o "$1" -le 0 ]; then
         echo "Function \"wait_until_omsagent_stops\" called with invalid parameter"
-        exit 1
+        WS_STATUS=1
+        exit $WS_STATUS
```
lagalbra: Same multiple workspaces concern: this method should return, not exit

lagalbra: Also, I would consider this an internal error case

installer/scripts/service_control
```diff
@@ -69,49 +176,107 @@ verify_privileges()
...
     fi
 
     COUNTER=$(( $1 * 2 )) # Since we sleep 0.5 seconds, compute number of seconds
     while [ $COUNTER -gt 0 ]; do
-        is_omsagent_running && return $?
-        COUNTER=$(( $COUNTER - 1 ))
-        sleep 0.5
+        if this_omsagent_running; then
```
lagalbra: I think you can condense this block: since `remove_PIDfile_unless_omsagent_running` calls `this_omsagent_running` itself, you can just call `remove_PIDfile_unless_omsagent_running`, and sleep if the return value indicates that omsagent is still running.

installer/scripts/service_control
```diff
@@ -121,20 +286,30 @@ stop_omsagent_process()
     #
     # TODO: omsagent doesn't currently gracefully shut down, we're investigating this later
 
-    is_omsagent_running && return $?
+    if ! remove_PIDfile_unless_omsagent_running; then
+        WS_STATUS=$?
+        return $WS_STATUS
+    fi
 
+    #  XC on 20170824; Looks like the following two lines can be removed:
```
lagalbra: Yes, you can remove the following two lines

installer/scripts/service_control
```diff
@@ -121,20 +286,30 @@ stop_omsagent_process()
...
     echo -n "(Forced) "
     kill -sigkill `cat $PIDFILE`
-    wait_until_omsagent_stops 5
-    return $?
+    wait_until_omsagent_stops $STOP_PROCESS_WAIT_MAX
+    WS_STATUS=$?
+    return $WS_STATUS
 }
 
+    #  XC on 20170824; Laura says this is no longer used:
```
lagalbra: Yes, I think you can remove this whole method

installer/scripts/service_control
```diff
@@ -283,12 +471,16 @@ enable_omsagent_service()
             elif [ -x /sbin/chkconfig ]; then
                 /sbin/chkconfig --add $OMSAGENT_WS > /dev/null
             else
-                echo "Could not find a service controller to configure the OMS Agent Service."
-                exit 1
+                echo "FATAL:  Could not find a service controller to configure the OMS Agent Service."
+                exit 9
```
lagalbra: Make this code constant
Same multiple workspaces concern: this method should return, not exit

installer/scripts/service_control
```diff
@@ -283,12 +471,16 @@ enable_omsagent_service()
...
             fi
         fi
-        touch  $CONF_DIR/.service_registered
+        WS_STATUS=$?
+        if [ "$WS_STATUS" -eq 0 ]; then
```
lagalbra: integer comparison: `$WS_STATUS -eq 0`

installer/scripts/service_control
```diff
@@ -302,107 +494,143 @@ disable_omsagent_service()
...
         if [ -f /usr/sbin/update-rc.d ]; then
             /usr/sbin/update-rc.d -f $OMSAGENT_WS remove
         elif [ -x /usr/lib/lsb/remove_initd ]; then
             /usr/lib/lsb/remove_initd $omsagent_initd
         elif [ -x /sbin/chkconfig ]; then
             /sbin/chkconfig --del $OMSAGENT_WS > /dev/null
         else
-            echo "Unrecognized Service Controller to unregister OMS Agent Service."
+            echo "FATAL:  Unrecognized Service Controller to unregister OMS Agent Service."
             exit 1
```
lagalbra: Assign a specific error code to this.
This may be the only fatal case where we should immediately exit; if we find no way to start a service, than no workspace service will be able to be started.

installer/scripts/service_control
```diff
@@ -302,107 +494,143 @@ disable_omsagent_service()
...
 
         rm $omsagent_initd
     fi
     rm -f $CONF_DIR/.service_registered
+    return $WS_STATUS
 }
 
 check_ws_and_invoke()
 {
+    fcso_new
```
lagalbra: I don't think you should be reinitializing your error count here; take `stop_all_omsagent` for example. Currently, you increment the error count for all the OMS workspaces, then you call `check_ws_and_invoke`, which overwrites the error count you were storing, which is bad.

installer/scripts/service_control
```diff
@@ -16,33 +19,107 @@
...
+COMMONSEGMENT=/opt/microsoft/omsagent
+BIN_DIR=$COMMONSEGMENT/bin
+ETC_DIR=/etc$COMMONSEGMENT
+VAR_DIR=/var$COMMONSEGMENT
+
+# Error codes and categories:
+
+# User configuration/parameters:
+
+INVALID_OPTION_PROVIDED=2
```
lagalbra: Since your error counts will generally be smaller numbers (+1 for every workspace with an error), I would suggest making your other explicit error codes >=10.

v-xecamp: Yes, I'd like to use standard stuff.  Can we begin an omsagent_errors.env file so we all use the same ones?

installer/scripts/service_control
```diff
@@ -16,33 +19,107 @@
...
+NO_OPTION_PROVIDED=3
+
+# Internal errors:
+
+INTERNAL_ERROR=30
+
+# Files and I/O
+
+FILE_NOT_FOUND=1
```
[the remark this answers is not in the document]
v-xecamp: Yes, I used that because it seemed to be what sh returns.  Same question as above.

installer/scripts/service_control
```diff
@@ -16,33 +19,107 @@
...
+
+# 
+REGEX_UUID='[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}'
+WORKSPACE_REGEX="^$REGEX_UUID$"
+
+# Begin Lower Level Utilities
+
+validate_file_has_line_pattern()
+{
+    # Does regular expression check for entire line.  As a validaytor routine,
```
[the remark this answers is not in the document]
v-xecamp: OOh.  finger faltering I see.  Got it.

installer/scripts/service_control
```diff
@@ -16,33 +19,107 @@
...
+    # it exits with a non-zero value if the validation failes.  Specifying any
```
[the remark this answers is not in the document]
v-xecamp: Got it.

installer/scripts/service_control
```diff
@@ -16,33 +19,107 @@
...
+    local _actStr=$1
+    local _wId=$2
+    local _sCode=$3
+
+    if [ "$_sCode" -eq "0" ]; then
+        return
+    else
+        FCSO_FailCount=`expr $FCSO_FailCount + 1`
+    fi
+    echo "ERROR:  $_actStr failed with result '$_sCode' on WS Id $_wId."
```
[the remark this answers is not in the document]
v-xecamp: Yes, of course.  It is used across all so I will take out ID.

installer/scripts/service_control
```diff
@@ -59,6 +136,36 @@ setup_variables()
...
+    WS_STATUS=0
+}
+
+validate_omsadmin_conf()
+{
+    # Checks for a series of identifier assignment lines which should always be present
+    # for an onboarded workspace to be complete and healthy.
+    local _fSpec=$1
+    local _boolOnly=$2
+    VALIDATION_SUPPLEMENT_MESSAGE="Onboarding incorrect or incomplete."
```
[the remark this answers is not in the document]
v-xecamp: Yes, I took those out.  The SUPPLEMENT items were experimental in my libraries I originally made these from and there is a better way to do this anyway.  That being said, as I imply above, I think an OMS_Messaging_Library might be a good idea.  I also have a log library that may be well to do the same time.

installer/scripts/service_control
```diff
@@ -59,6 +136,36 @@ setup_variables()
...
+        validate_file_has_line_pattern $_fSpec "AZURE_RESOURCE_ID=.*" $_boolOnly
+        validate_file_has_line_pattern $_fSpec "OMSCLOUD_ID=.*" $_boolOnly
+        validate_file_has_line_pattern $_fSpec "UUID=$REGEX_UUID" $_boolOnly
+        return 0
+    else
+        if [ -z "$_boolOnly" ]; then
+            echo "INTERNAL_ERROR:  OMS Admin configuration file $_fSpec NOT Found!" >&2
+            echo $VALIDATION_SUPPLEMENT_MESSAGE
+        fi
+        exit $INTERNAL_ERROR
```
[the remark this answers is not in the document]
v-xecamp: Ok.  Sounds good.

installer/scripts/service_control
lagalbra: It may be possible to have a file like that to store constants such as error codes; in my mind, it would go under `/opt/microsoft/omsagent/bin/`.

Our files which current use constant error codes are:
- [bundle skeleton](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/bundle/bundle_skel.sh); this won't be able to use a packaged file like the rest of the code might, since the bundle has to run before it can assume the file will exist. Pulling constant error codes from this file to the bundle code may require some work; perhaps it can be inserted dynamically by our [create_bundle script](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/bundle/create_bundle.sh)?
- [omsadmin.sh](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/scripts/omsadmin.sh); this can simply execute the error codes file to get its values
- service_control script
- [Ruby maintenance helper script](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/source/code/plugins/agent_maintenance_script.rb) - reading an error code file meant for shell will probably require a little more work here

installer/scripts/service_control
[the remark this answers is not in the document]
v-xecamp: I think the way to go with this is leave the exits always on validation, and instead when multiples are handled, execute a subshell at that point so that you keep the solid consistency of a real validation routine, which by my experience exits on failure, and yet allows multiple processing by use of the subprocess that can just do the exiting without leaving the entire set.  This makes the code cleaner too.

installer/scripts/service_control
v-xecamp: Yes.  Same comment about constructs.

installer/scripts/service_control
v-xecamp: Yes, I think I already made that change.  Thanks for the reminder.  I'll make sure.

installer/scripts/service_control
v-xecamp: Same response.

installer/scripts/service_control
v-xecamp: I agree.  Please let's talk about that.  Some of these I looked for an existing one, and did not find it.  There may be more than one area, and I really am preferring to start a library for them, otherwise I can picture gobs of multiply duplicated work here.

installer/scripts/service_control
v-xecamp: Yes, that sounds right.  I'll check it.

v-xecamp: Sorry again for the email mess.

installer/scripts/service_control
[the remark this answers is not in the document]
lagalbra: That design is up to you. If you can implement it and test it successfully, then that sounds good.

installer/scripts/service_control
lagalbra: Please remove this whole method; it isn't used.

installer/scripts/service_control
```diff
@@ -159,16 +349,22 @@ start_omsagent()
...
+        WS_STATUS=$?
+    fi
+    return $WS_STATUS
 }
 
 stop_omsagent()
 {
-    is_omsagent_running
-    if [ $? -ne 0 ]; then
+    remove_PIDfile_unless_omsagent_running
```
lagalbra: Same "react if remove_PIDfile_unless_omsagent_running returns non-zero here" comment here

installer/scripts/service_control
```diff
@@ -159,16 +349,22 @@ start_omsagent()
         elif [ -x /bin/systemctl ]; then
             /bin/systemctl start $OMSAGENT_WS
         else
-            echo "Unrecognized service controller to start OMS Agent service" 1>&2
+            echo "ERROR:  Unrecognized service controller to start OMS Agent service" 1>&2
             exit 1
```
lagalbra: We can discuss this tomorrow, but this should be assigned a specific error code.

installer/scripts/service_control
```diff
@@ -180,23 +376,32 @@ stop_omsagent()
             elif [ -x /bin/systemctl ]; then
                 /bin/systemctl stop $OMSAGENT_WS
             else
-                echo "Unrecognized service controller to stop OMS Agent service" 1>&2
+                echo "ERROR:  Unrecognized service controller to stop OMS Agent service" 1>&2
                 exit 1
```
lagalbra: Same error code comment here

installer/scripts/service_control
```diff
@@ -210,10 +415,16 @@ restart_omsagent()
         elif [ -x /bin/systemctl ]; then
             /bin/systemctl restart $OMSAGENT_WS
         else
-            echo "Unrecognized service controller to restart OMS Agent service" 1>&2
+            echo "ERROR:  Unrecognized service controller to restart OMS Agent service" 1>&2
             exit 1
```
lagalbra: Same error code comment here

installer/scripts/service_control
lagalbra: Let's discuss this tomorrow

installer/scripts/service_control
```diff
@@ -283,12 +495,16 @@ enable_omsagent_service()
...
-                echo "Could not find a service controller to configure the OMS Agent Service."
-                exit 1
+                echo "FATAL:  Could not find a service controller to configure the OMS Agent Service."
+                exit 9
             fi
         fi
-        touch  $CONF_DIR/.service_registered
+        WS_STATUS=$?
+        if [ "$WS_STATUS" -eq 0 ]; then
+            touch  $CONF_DIR/.service_registered
```
lagalbra: nit: remove extra space in this line

installer/scripts/service_control
```diff
@@ -302,167 +518,158 @@ disable_omsagent_service()
...
-        if [ $? -eq 0 ]; then
-            setup_variables ${ws_id}
-            $1
-        fi
-    done
-}
+# ## Interface Sub-Submethods
 
-start_all_omsagent()
+generateProcedureName()
```
lagalbra: nit: name this using the same style as the rest of the method names in this file: `generate_procedure_name`

installer/scripts/service_control
```diff
@@ -302,167 +518,158 @@ disable_omsagent_service()
...
-       setup_variables ${ws_id}
-       stop_omsagent
-    done
+# ## Iterators #### 
 
-    check_ws_and_invoke stop_omsagent
-}
-
-restart_all_omsagent()
+check_non_oms_and_invoke()
```
lagalbra: Fix indentation in this method.

installer/scripts/service_control
```diff
@@ -302,167 +518,158 @@ disable_omsagent_service()
...
+            fi
+            fcso_increment $_cmdArg $ws_id $WS_STATUS
+        fi
     done
-
-    check_ws_and_invoke enable_omsagent_service
 }
 
-disable_all_omsagent_services()
+check_oms_and_invoke()
```
lagalbra: Fix indentation in this method.

installer/scripts/service_control
```diff
@@ -302,167 +518,158 @@ disable_omsagent_service()
...
 
-    restart)
-        if [ -z $2 ]; then
-            restart_all_omsagent
-        else
-            setup_variables $2
-            restart_omsagent
-        fi
-        ;;
+service_template() {
```
lagalbra: nit: method declaration styling:
```
service_template()
{
...
}
```

lagalbra: Also, fix indentation in this method.

installer/scripts/service_control
```diff
@@ -302,167 +518,158 @@ disable_omsagent_service()
...
-    reload)
-        # TODO: Due to a bug in OMS right now, we can't reload via a signal
-        if [ -z $2 ]; then
-            restart_all_omsagent
-        else
-            setup_variables $2
-            restart_omsagent
+    verify_privileges
+	l_procedure=$(generateProcedureName $_activityName)
+	if [ -n "$_wsId" ]
```
lagalbra: Same if statement format comment

installer/scripts/service_control
```diff
@@ -302,167 +518,158 @@ disable_omsagent_service()
...
-        if [ -z $2 ]; then
-	    disable_all_omsagent_services
-        else
-            setup_variables $2
-	    disable_omsagent_service
-        fi
-        ;;
+# ## Main Procedure
 
+case "$1" in
```
lagalbra: Please use styling for this case block like bundle_skel uses as well. As it is below, the spacing is inconsistent, and some cases span multiple lines while others do not.

installer/scripts/service_control
```diff
@@ -302,167 +518,158 @@ disable_omsagent_service()
...
+                            fi;;
+    #TODO: Due to a bug in OMS right now, we can't reload via a signal (Stated Before 2017/08)
+    reload)					service_template restart $2;;
+    restart)				service_template restart $2;;
+    start)					service_template start $2;;
+    stop)					service_template stop $2;;
     *)
         echo "Unknown parameter : $1" 1>&2
+        printUsage
         exit 1
```
lagalbra: Should be specific error code; probably INVALID_OPTION_PROVIDED

installer/scripts/service_control
v-xecamp: Okay, I have headway on this, and it passes tests.  next.

installer/scripts/service_control
v-xecamp: Yes, I think what I've heard is unless you are always sure, like with $?, it is better to do "$WS_STATUS" -eq "0" or equivalent.  I'm trying to stick to that as long as it's POSIX and not Bash.  In Bash, which I don't think we should follow here, [[]] constructs are recommended, which are much nicer.

installer/scripts/service_control
```diff
@@ -16,49 +19,164 @@
...
+        if ! check_file_has_line_pattern $_fs "WORKSPACE_ID=$REGEX_UUID";       then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "AGENT_GUID=$REGEX_UUID";         then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "LOG_FACILITY=.*";                then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "CERTIFICATE_UPDATE_ENDPOINT=.*"; then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "URL_TLD=.*";                     then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "DSC_ENDPOINT=.*";                then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "OMS_ENDPOINT=.*";                then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "AZURE_RESOURCE_ID=.*";           then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "OMSCLOUD_ID=.*";                 then ec=$(expr $ec + 1); fi
+        if ! check_file_has_line_pattern $_fs "UUID=$REGEX_UUID";               then ec=$(expr $ec + 1); fi
```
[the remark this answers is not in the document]
v-xecamp: As per above I'm making them all in higher number categories.  I'm not paying attention to the value consistency with other scripts for now, as that looks like a lot of work, but that is a reasonable long term ideal.  See next push.

installer/scripts/service_control
```diff
@@ -16,49 +19,164 @@
...
+        return $ec
+    else
+        echo "INTERNAL_ERROR:  OMS Admin configuration file $_fs NOT Found!" >&2
+        return $INTERNAL_ERROR
```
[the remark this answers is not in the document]
v-xecamp: Yep..

installer/scripts/service_control
```diff
@@ -69,72 +187,142 @@ verify_privileges()
...
+    local omsagent_pid=`cat $PIDFILE 2>/dev/null`
+    local ps_state=
+    WS_STATUS=0
+    if [ -f $PIDFILE ]; then
+        ps_state=`ps --no-header -o state -p $omsagent_pid`
+        if [ -z "$ps_state" ]; then
+            WS_STATUS=1 # No.
+        else
+            case "$ps_state" in
+            D)  echo "Uninterruptable Sleep State Seen in omsagent process.";;
```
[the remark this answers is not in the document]
v-xecamp: Yes, I find constructing short cases with simple columnar alignments optimizes maintainability.  Looking at this now, but it may require a risky addition to get that.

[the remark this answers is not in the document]
v-xecamp: That one was different than I thought.  Got the alignment as you suggested.

installer/scripts/service_control
```diff
@@ -69,72 +187,142 @@ verify_privileges()
...
+            R)  ;;
+            S)  ;;
+            T)  echo "Stopped State Seen in omsagent process.";                 WS_STATUS=$UNEXPECTED_STATE;;
+            W)  echo "Paging State Seen in omsagent process.";                  WS_STATUS=$UNEXPECTED_STATE;;
+            X)  echo "Dead State Seen in omsagent process.";                    WS_STATUS=$UNEXPECTED_STATE;;
+            Z)  echo "Defunct State Seen in omsagent process.";                 WS_STATUS=$UNEXPECTED_STATE;;
+            *)  echo "INTERNAL ERROR:  '$ps_state' is not a known ps flag."
+                exit INTERNAL_ERROR;;
```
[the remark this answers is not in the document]
v-xecamp: Sounds good.

installer/scripts/service_control
```diff
@@ -69,72 +187,142 @@ verify_privileges()
...
+            COUNTER=$(( $COUNTER - 1 ))
+            sleep 0.5
+        else
+            return 0
+        fi
     done
 
-    # One final try for accurate return status (just return status from the call)
-    is_omsagent_running
+    return 1 # Never came up.
```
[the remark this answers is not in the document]
v-xecamp: I had to add to your comments above.  

I do not see the sleep.  It may have been a leftover I already took out from debug code.

installer/scripts/service_control
```diff
@@ -69,72 +187,142 @@ verify_privileges()
...
+            sleep 0.5
+        else
+            remove_PIDfile_unless_omsagent_running
+            return $? # Should only be non-zero if a PIDFILE remains.
+        fi
+    done
 
-    # kill -sigterm `cat $PIDFILE`
-    # wait_until_omsagent_stops 10 && return $?
+    return 1 # Never stopped.
```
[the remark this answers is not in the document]
v-xecamp: Good catch on the sleeps.  Thank you.

installer/scripts/service_control
```diff
@@ -16,137 +19,399 @@
...
+ETC_DIR=/etc$COMMONSEGMENT
+VAR_DIR=/var$COMMONSEGMENT
+
+# Error codes and categories:
+
+
+# User configuration/parameters:
+# User configuration/parameters:
+
+INVALID_OPTION_PROVIDED=72
```
lagalbra: Please remove all of these that you do not use in this file.

installer/scripts/service_control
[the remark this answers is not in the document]
v-xecamp: The better way is to have a little routine that does those pieces.  That is a little higher maintenence in shell since there are no embedded routines, but it's easy then to test the thing.  I suggest we wait until we separate out the api from the main and at that time, we can go through and divide things out for better unit testing anyway.

installer/scripts/service_control
```diff
@@ -302,167 +590,108 @@ disable_omsagent_service()
...
+    #TODO: Due to a bug in OMS right now, we can't reload via a signal (Stated Before 2017/08)
+    reload)					service_template restart $2;;
+    restart)				service_template restart $2;;
+    start)					service_template start $2;;
+    stop)					service_template stop $2;;
     *)
         echo "Unknown parameter : $1" 1>&2
-        exit 1
+        printUsage
+        notification_exit $NO_OPTION_PROVIDED
```
lagalbra: This should be "invalid operation provided" because a parameter $1 was provided.

installer/scripts/service_control
v-xecamp: I should have posted my commend about the remove_* procedure in here somewhere.  Please see that Laura and I found the original state worked better because remove_PID* procedure succeeds both at times when omsagent is running, and at times when it is not, by design.  Comments have been added.

installer/scripts/service_control
v-xecamp: No, I like to see consistent usage.  I saw that in one location, so I used it here.  I realize now we have both.  I suggest, again, that we regularize such things when we divide the library from the script.

installer/scripts/service_control
```diff
@@ -16,137 +19,399 @@
...
+        start)      echo "start_omsagent";;
+        stop)       echo "stop_omsagent";;
+        *)          echo "INTERNAL ERROR:  '$1' is not a valid procedural activity."
+                    notification_exit INTERNAL_ERROR;;
+    esac
+}
+
+# ## Iterators #### 
+
+check_non_oms_and_invoke()
```
lagalbra: Fix indentation in this method - please scan your code before check-in and make sure that everything is aligned properly.

installer/scripts/service_control
lagalbra: This is still comparing a string to an integer - have you fixed this and I'm looking at the wrong commit?

installer/scripts/service_control
v-xecamp: See email exchange on use of remove_PID* procedure.

lagalbra:
@v-xecamp Please ping me directly once you are done adding small changes and have a copy which you are confident you can check in. This should minimize back-and-forth.

installer/scripts/service_control
```diff
@@ -16,137 +19,382 @@
...
+# Internal errors:
+
+INTERNAL_ERROR=30
+
+# Files and I/O
+
+FILE_NOT_FOUND=21
+FILE_NOT_GONE=22
+
+#
```
lagalbra: nit: delete empty comment line

installer/scripts/service_control
```diff
@@ -16,137 +19,382 @@
...
 
-exit_if_agent_not_onboarded()
+wait_until_omsagent_is_running()
 {
-    if [ ! -f $OMSADMIN_CONF ]; then
-        echo "Warning: Agent is not onboarded. omsagent cannot be registered as a service."
-        exit 0 
+    # Required parameter: Number of seconds to wait for agent to become running.
+    if [ -z "$1" -o "$1" -le 0 ]; then
+        echo "INTERNAL ERROR:  Function \"wait_until_omsagent_stops\" called with invalid parameter"
```
lagalbra: You forgot to paste in the correct method name: wait_until_omsagent_is_running

installer/scripts/service_control
```diff
@@ -16,137 +19,382 @@
...
 }
 
-wait_until_omsagent_stops()
+wait_until_omsagent_stopped()
 {
     # Required parameter: Number of seconds to wait for agent to stop
     if [ -z "$1" -o "$1" -le 0 ]; then
-        echo "Function \"wait_until_omsagent_stops\" called with invalid parameter"
-        exit 1
+        echo "INTERNAL ERROR:  Function \"wait_until_omsagent_stops\" called with invalid parameter"
```
lagalbra: You forgot to change this to the new method name: wait_until_omsagent_stopped

installer/scripts/service_control
```diff
@@ -16,137 +19,382 @@
...
+        stop)       echo "stop_omsagent";;
+        *)          echo "INTERNAL ERROR:  '$1' is not a valid procedural activity."
+                    notification_exit INTERNAL_ERROR;;
+    esac
 }
 
-start_omsagent_process()
+# ## Iterators #### 
+
+check_non_oms_and_invoke()
```
lagalbra: Indentation in this method is still inconsistent - please fix these methods: check_non_oms_and_invoke, check_oms_and_invoke, service_template, and the main procedure case statement!

installer/scripts/service_control
```diff
@@ -222,11 +490,11 @@ find_systemd_dir()
     # Various distributions have different paths for systemd unit files ...
     local UNIT_DIR_LIST="/usr/lib/systemd/system /lib/systemd/system"
 
-    if pidof systemd 1> /dev/null 2> /dev/null; then
+    if pidof systemd 1> /dev/null 2>&1; then
```
lagalbra: Why have you changed this line? If the system does not have systemd, it should have something else, and that's expected.

installer/scripts/service_control
lagalbra: This is still comparing a string to an integer - please fix this!

v-xecamp:
Yes, I had them squashed.  I just cannot get to github any longer from there.  I may try to bring the tree down to local…

installer/scripts/service_control
v-xecamp: I think what is most omnipresent is probably the most important, and after that what is clearer or more acceptable to the group, as the project group will mostly read this, despite its being public.  I have been trying to pull stuff up I've seen elsewhere in the code, so I probably saw both somewhere.

installer/scripts/service_control
v-xecamp: Good point.

installer/scripts/service_control
v-xecamp: Same.

installer/scripts/service_control
v-xecamp: Going through the entire file...Okay, I took all those kinds out (anything with -eq 0, -ne 0, -gt, -lt -ge -le 0, EXCEPT those where we know they are numeric like `id -u`, $COUNTER, and $?).

installer/scripts/service_control
```diff
@@ -69,72 +187,142 @@ verify_privileges()
     fi
 }
 
-is_omsagent_running()
+this_omsagent_running()
 {
-    verify_privileges
+    local omsagent_pid=`cat $PIDFILE 2>/dev/null`
+    local ps_state=
```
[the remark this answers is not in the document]
v-xecamp: yes, good catch.  Thank you.

installer/scripts/service_control
v-xecamp: These were discussed around oddments with remove routine.  I think they way it is now is good.

installer/scripts/service_control
v-xecamp: Got it.  That was where I lost my build machine.  ;^)

installer/scripts/service_control
v-xecamp: Ok, I checked and took out all tabs.  I will look once more, but I think I have this.  Next push will be around 1:30pm 9/27.

installer/scripts/service_control
v-xecamp: I think I have all those now.  Pushing any second.

lagalbra: Please answer my question here; this line is still changed.

v-xecamp: I don't understand the question.  The intent is to use 2>&1; instead of the duplicated form.  I apologize if there is something else I am not seeing.  I'll come over.

PR #574, opened 2017-09-22
Improve system tests and remove_all logic

lagalbra:
No extraneous messages are printed out during system testing
Improvement of remove_all workspaces logic for non-OMS workspace cases
Localized unit test for containerized agent hostname

@Microsoft/omsagent-devs

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -97,6 +97,9 @@ SCX_SSL_CONFIG=/opt/microsoft/scx/bin/tools/scxsslconfig
 OMI_CONF_FILE=/etc/opt/omi/conf/omiserver.conf
 OMI_CONF_EDITOR=/opt/omi/bin/omiconfigeditor
 
+# Space seperated list of non oms workspaces
+NON_OMS_WS="scom LAD"
+
 # Error codes and categories:
 
 # User configuration/parameters:
@@ -592,6 +595,7 @@ onboard()
         echo
         echo "Generated request:"
         cat $BODY_ONBOARD
+        echo
     fi
 
     set_proxy_setting
@@ -632,7 +636,7 @@ onboard()
         cleanup_certs
         return $INVALID_PROXY
     elif [ $error -ne 0 ]; then
-        log_error "Error during the onboarding request. Check the correctness of the workspace ID and shared key or run omsadmin.sh with '-v'"
+        log_error "Error during the onboarding request: curl returned $error. Check the correctness of the workspace ID and shared key or run omsadmin.sh with '-v'"
         cleanup_certs
         return $ERROR_ONBOARDING
     fi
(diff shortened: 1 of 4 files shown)


lagalbra:
pBuild running now

lagalbra:
pBuild, unit tests, and system tests pass.

lagalbra:
@v-xecamp Please review. This should be checked in before https://github.com/Microsoft/OMS-Agent-for-Linux/pull/575

installer/scripts/omsadmin.sh
```diff
@@ -592,6 +595,7 @@ onboard()
         echo
         echo "Generated request:"
         cat $BODY_ONBOARD
+        echo
```
v-xecamp: I could completely change this to a here document as it is much more readable and maintainable:
cat <<EOGREQ # or whatever

Generated request:
$BODY_ONBOARD

EOGREQ

installer/scripts/omsadmin.sh
```diff
@@ -632,7 +636,7 @@ onboard()
         cleanup_certs
         return $INVALID_PROXY
     elif [ $error -ne 0 ]; then
-        log_error "Error during the onboarding request. Check the correctness of the workspace ID and shared key or run omsadmin.sh with '-v'"
+        log_error "Error during the onboarding request: curl returned $error. Check the correctness of the workspace ID and shared key or run omsadmin.sh with '-v'"
         cleanup_certs
         return $ERROR_ONBOARDING
```
v-xecamp: I thought I saw somewhere that there are curl specific errors.  Are those pertinent here?  Perhaps even to embed in the log message if not to return.

test/installer/scripts/maintenance_systestbase.rb
```diff
@@ -72,16 +72,26 @@ def prep_proxy(proxy_setting)
     check_proxy_server_working
   end
 
-  def do_onboard(workspace_id, shared_key, should_succeed = true)
+  def do_onboard(workspace_id, shared_key, should_succeed = true, verbose = false)
     check_test_keys()
-    onboard_out = `#{@omsadmin_script} -w #{workspace_id} -s #{shared_key}`
-    result = $?
+    onboard_cmd = "#{@omsadmin_script}"
+    if !workspace_id.nil?
```
v-xecamp: unless workspace_id.nil?

test/installer/scripts/maintenance_systestbase.rb
```diff
@@ -72,16 +72,26 @@ def prep_proxy(proxy_setting)
...
+       onboard_cmd = "#{onboard_cmd} -w #{workspace_id}"
+    end
+    if !shared_key.nil?
```
v-xecamp: unless shared_key.nil?

installer/scripts/omsadmin.sh
lagalbra: As you know, I'm changing how the generated request is printed in https://github.com/Microsoft/OMS-Agent-for-Linux/pull/575, and I don't think this comment applies once those changes are taken into consideration.

installer/scripts/omsadmin.sh
lagalbra: You can see the curl-specific error codes here: https://help.github.com/articles/testing-your-ssh-connection/

If you look in the lines of code above these, we check if any of those curl-specific error codes apply; if not, we fall to this case. This changed line is just adding curl's return code to the printout for diagnostic purposes.

test/installer/scripts/maintenance_systestbase.rb
lagalbra: Will change

test/installer/scripts/maintenance_systestbase.rb
lagalbra: Will change

PR #576, opened 2017-09-26
Update README.md

mabicca:
Adding more clarity about where to find workspace key on the portal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 Welcome to the OMS Agent for Linux! The OMS Agent for Linux enables rich and real-time analytics for operational data (Syslog, Performance, Alerts, Inventory) from Linux servers, Docker Containers and monitoring tools like Nagios, Zabbix and System Center.
 
 ## Quick Install guide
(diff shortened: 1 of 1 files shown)


lagalbra:
@mabicca Please see our [Contributing Guide](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/CONTRIBUTING.md); we are not accepting pull requests from outside the team at this time.
Please be assured that we are actively working on improving our documentation.

NarineM:
@lagalbra , this is a good addition and it is a small change. Can you merge this  with master?

lagalbra:
@keikhara for FYI - this should be considered with the doc changes Matt is making

@NarineM The exact changes proposed here are not entirely clear/correct:
In the Azure portal, the workspace ID and primary key can be found under Log Analytics > <workspace> > Settings > Advanced Settings
In the OMS portal, the workspace ID and primary key can be found under Settings > Connected Sources.

The proposed changes combine parts of these two sources, but not completely correct for either portal. Docs work is still ongoing with Keiko and Matt.

NarineM:
I am OK with just leaving Azure portal way, or actually have both. In any case I do not see any harm adding this info. Many folks who are new to omsagent onboarding experience have hard time finding these values.

PR #582, opened 2017-09-29
Refactor oms_common.rb

v-xecamp:
Additions to handle address patterns were added.  Corresponding supporting tests also added.

### DIFF
--- a/source/code/plugins/oms_common.rb
+++ b/source/code/plugins/oms_common.rb
@@ -1,5 +1,9 @@
 module OMS
 
+  MSDockerCImprovHostnameFilePath = '/var/opt/microsoft/docker-cimprov/state/containerhostname'
+  IPV6_REGEX = '\h{4}:\h{4}:\h{4}:\h{4}:\h{4}:\h{4}:\h{4}:\h{4}'
+  IPV4_Approximate_REGEX = '\d+\.\d+\.\d+\.\d+'
+
   class RetryRequestException < Exception
     # Throw this exception to tell the fluentd engine to retry and
     # inform the output plugin that it is indeed retryable
@@ -22,7 +26,7 @@ module OMS
     @@OSName = nil
     @@OSVersion = nil
     @@Hostname = nil
-    @@HostnameFilePath = '/var/opt/microsoft/docker-cimprov/state/containerhostname'
+    @@HostnameFilePath = MSDockerCImprovHostnameFilePath
     @@FQDN = nil
     @@InstalledDate = nil
     @@AgentVersion = nil
@@ -478,6 +482,111 @@ module OMS
     @@tzRightFolder = 'right/'
 
     class << self
+
+      # Internal methods
+      # (left public for easy testing, though protected may be better later)
+
+      def clean_hostname_string(hnBuffer)
+        return "" if hnBuffer.nil? # So give the rest of the program a string to deal with.
+        hostname_buffer = hnBuffer.strip
+        return hostname_buffer
+      end
+
+      def has_designated_hostnamefile?
+        return false if @@HostnameFilePath.nil?
+        return false unless @@HostnameFilePath =~ /\w/
+        return false unless File.exist?(@@HostnameFilePath)
+        return true
+      end
+
+      def is_dot_separated_string?(hnBuffer)
+        return true if /[^.]+\.[^.]+/ =~ hnBuffer
+        return false
+      end
+
+      def is_hostname_compliant?(hnBuffer)
+        # RFC 2181:
+        #   Size limit is 1 to 63 octets, so probably bytesize is appropriate method.
+        return false if hnBuffer.nil?
+        return false if /\./ =~ hnBuffer # Hostname by definition may not contain a dot.
+        return false if /:/ =~ hnBuffer # Hostname by definition may not contain a colon.
+        return false unless 1 <= hnBuffer.bytesize && hnBuffer.bytesize <= 63
+        return true
+      end
+
+      def is_like_ipv4_string?(hnBuffer)
+        return false unless /\A#{IPV4_Approximate_REGEX}\z/ =~ hnBuffer
+        qwa = hnBuffer.split('.')
+        return false unless qwa.length == 4
+        return false if qwa[0].to_i == 0
+        qwa.each do |quadwordstring|
+            bi = quadwordstring.to_i
+            # This may need more detail if 255 octets are sometimes allowed, but I don't think so.
+            return false unless 0 <= bi and bi < 255
+        end
+        return true
+      end
+
+      def is_like_ipv6_string?(hnBuffer)
+        return true if /\A#{IPV6_REGEX}\z/ =~ hnBuffer
+        return false
+      end
+
+      def look_for_socket_class_host_address
+        hostname_buffer = nil
+
+        begin
+          hostname_buffer = Socket.gethostname
+        rescue => error
+          OMS::Log.error_once("Unable to get the Host Name using socket facility: #{error}")
+          return
+        end
+        @@Hostname = clean_hostname_string(hostname_buffer)
+
+        return # Thwart accidental return to force correct use.
+      end
+
+      def look_in_designated_hostnamefile
+        # Issue:
+        #   When omsagent runs inside a container, gethostname returns the hostname of the container (random name)
+        #   not the actual machine hostname.
+        #   One way to solve this problem is to set the container hostname same as machine name, but this is not
+        #   possible when host-machine is a private VM inside a cluster.
+        # Solution:
+        #   Share/mount ‘/etc/hostname’ as '/var/opt/microsoft/omsagent/state/containername' with container and
+        #   omsagent will read hostname from shared file.
+        hostname_buffer = nil
+
+        unless File.readable?(@@HostnameFilePath)
+            OMS::Log.warn_once("File '#{@@HostnameFilePath}' exists but is not readable.")
+            return
+        end
+
+        begin
+          hostname_buffer = File.read(@@HostnameFilePath)
+        rescue => error
+          OMS::Log.warn_once("Unable to read the hostname from #{@@HostnameFilePath}: #{error}")
+        end
+        @@Hostname = clean_hostname_string(hostname_buffer)
+        return # Thwart accidental return to force correct use.
+      end
+
+      def validate_hostname_equivalent(hnBuffer)
+        # RFC 1123 and 2181
+        # Note that for now we are limiting the earlier maximum of 63 for fqdn labels and thus
+        # hostnames UNTIL we are assured azure will allow 255, as specified in RFC 1123, or
+        # we are otherwise instructed.
+        rfcl = "RFCs 1123, 2181 with hostname range of {1,63} octets for non-root item."
+        return if is_hostname_compliant?(hnBuffer)
+        return if is_like_ipv4_string?(hnBuffer) 
+        return if is_like_ipv6_string?(hnBuffer)
+        msg = "Hostname '#{hnBuffer}' not compliant (#{rfcl}).  Not IP Address Either."
+        OMS::Log.warn_once(msg)
+        raise NameError, msg
+      end
+
+      # End of Internal methods
+
       # get the unified timezone id by absolute file path of the timezone file
       # file path: the absolute path of the file
       def get_unified_timezoneid(filepath)
@@ -575,33 +684,19 @@ module OMS
         return @@OSVersion
       end
 
-      def get_hostname
-        return @@Hostname if !@@Hostname.nil?
+      def get_hostname(ignoreOldValue = false)
 
-        # Issue:
-        #   When omsagent runs inside a container, gethostname returns the hostname of the container (random name)
-        #   not the actual machine hostname.
-        #   One way to solve this problem is to set the container hostname same as machine name, but this is not
-        #   possible when host-machine is a private VM inside a cluster.
-        # Solution:
-        #   Share/mount ‘/etc/hostname’ as '/var/opt/microsoft/omsagent/state/containername' with container and
-        #   omsagent will read hostname from shared file.
+        if not is_hostname_compliant?(@@Hostname) or ignoreOldValue then
 
-        begin
-          if File.exist?(@@HostnameFilePath) && File.readable?(@@HostnameFilePath)
-            @@Hostname = File.read(@@HostnameFilePath).strip
-            return @@Hostname
-          end
-        rescue => error
-          OMS::Log.warn_once("Unable to read the hostname from #{@@HostnameFilePath}: #{error}")
+            look_in_designated_hostnamefile         if has_designated_hostnamefile?
+
+            look_for_socket_class_host_address  unless is_hostname_compliant?(@@Hostname)
         end
 
         begin
-          hostname = Socket.gethostname.split(".")[0]
+          validate_hostname_equivalent(@@Hostname)
         rescue => error
-          OMS::Log.error_once("Unable to get the Host Name: #{error}")
-        else
-          @@Hostname = hostname
+          OMS::Log.warn_once("Hostname '#{@@Hostname}' found, but did NOT validate as compliant.  #{error}.  Using anyway.")
         end
         return @@Hostname
       end
--- a/test/code/plugins/oms_common_test.rb
+++ b/test/code/plugins/oms_common_test.rb
@@ -4,12 +4,64 @@ require 'net/http'
 require 'net/https'
 require 'json'
 require 'fileutils'
+require 'flexmock/test_unit'
 require_relative 'omstestlib'
 require_relative ENV['BASE_DIR'] + '/source/code/plugins/oms_common'
 
+
 module OMS
 
+  TestHostnameList = 
+  [
+          #         Address Type  Compliant Name
+
+    TestHostname.new(:RFC1123Hostname, true,  'onetwo'),
+    TestHostname.new(:RFC1123Hostname, true,  'xxx1234'),
+    TestHostname.new(:RFC1123Hostname, true,  'x'),
+    TestHostname.new(:RFC1123Hostname, true,  'www'),
+    TestHostname.new(:RFC1123Hostname, true,  'Microsoft'),
+    TestHostname.new(:RFC1123Hostname, true,  'xxx123456789012345678901234567890123456789012345678901234567890'),
+
+    TestHostname.new(:RFC1123Hostname, false, ''),
+    TestHostname.new(:RFC1123Hostname, false, 'xxx123456789012345678901234567890123456789012345678901234567890x'),
+    TestHostname.new(:RFC1123Hostname, false, 'one.two'),
+    TestHostname.new(:RFC1123Hostname, false, 'one:two'),
+    TestHostname.new(:RFC1123Hostname, false, 'microsoft.com'),
+    TestHostname.new(:RFC1123Hostname, false, 'microsoft:355'),
+    TestHostname.new(:RFC1123Hostname, false, nil),
+
+    TestHostname.new(:IPv4,            true,  '192.168.0.5'),
+    TestHostname.new(:IPv4,            true,  '1.2.3.4'),
+    TestHostname.new(:IPv4,            true,  '254.254.254.254'),
+    TestHostname.new(:IPv4,            true,  '1.0.0.0'),
+
+    TestHostname.new(:IPv4,            false, ''),
+    TestHostname.new(:IPv4,            false, '192.168.0.256'),
+    TestHostname.new(:IPv4,            false, '1.0.0.444'),
+    TestHostname.new(:IPv4,            false, '192.168.0.A'),
+    TestHostname.new(:IPv4,            false, '0.2.3.4'),
+    TestHostname.new(:IPv4,            false, '1.2.3.255'),
+    TestHostname.new(:IPv4,            false, '255.255.255.255'),
+    TestHostname.new(:IPv4,            false, '192.168.0'),
+    TestHostname.new(:IPv4,            false, '192.168..5'),
+    TestHostname.new(:IPv4,            false, '192.168.0.5.5'),
+    TestHostname.new(:IPv4,            false, '192.168.0.5.A'),
+
+    TestHostname.new(:IPv6,            true,  '2001:0db8:85a3:0000:0000:8a2e:0370:7334'),
+    TestHostname.new(:IPv6,            true,  '1234:DDDD:FFFF:EEEE:3333:8888:7777:1111'),
+    TestHostname.new(:IPv6,            true,  '0db8:85a3:0000:0000:8a2e:0370:7334:f135'),
+
+    TestHostname.new(:IPv6,            false, ''),
+    TestHostname.new(:IPv6,            false, '1000:::::::1000'),
+    TestHostname.new(:IPv6,            false, 'XXXX:0db8:85a3:0000:0000:8a2e:0370:7334'),
+    TestHostname.new(:IPv6,            false, 'xx:0db8:85a3:0000:0000:8a2e:0370:7334'),
+    TestHostname.new(:IPv6,            false, ':0db8:85a3:0000:0000:8a2e:0370:7334'),
+    TestHostname.new(:IPv6,            false, '0db8:85a3:0000:0000:8a2e:0370:7334'),
+    TestHostname.new(:IPv6,            false, ':0db8:85a3:0000:0000:8a2e:0370:7334'),
+  ]
+
   class CommonTest < Test::Unit::TestCase
+    include FlexMock::TestCase
 
     # Extend class to reset OSFullName class variable
     class OMS::Common
@@ -48,6 +100,10 @@ module OMS
 
         def HostnameFilePath=(hostname_file_path)
           @@HostnameFilePath = hostname_file_path
+        end
+
+        def GetHostnameInternalForTest
+          return @@Hostname
         end
       end
     end
@@ -93,6 +149,106 @@ module OMS
       "OSFullName=CentOS Linux 7.0 (x86_64)\n" \
       "OSAlias=UniversalR\n" \
       "OSManufacturer=Central Logistics GmbH\n"
+
+    # Begin:  Tests on Public internals; probably only for use inside the Singular object.
+
+    def test_clean_hostname_string
+      hostname = '        microsoft     '
+      b = Common.clean_hostname_string(hostname)
+      expected = 'microsoft'
+      assert_equal(b, expected, "method clean_hostname_string should return '#{expected}'.")
+      hostname = expected
+      b = Common.clean_hostname_string(hostname)
+      assert_equal(b, expected, "method clean_hostname_string should return '#{expected}'.")
+    end
+
+    def test_has_designated_hostnamefile?
+      container_hostname_tempfile = Tempfile.new('containerhostname')
+      test_hostname = 'test_hostname'
+      Common.HostnameFilePath = container_hostname_tempfile.path
+      File.write(container_hostname_tempfile.path, test_hostname)
+      r = Common.has_designated_hostnamefile?
+      assert(r, "method has_designated_hostnamefile? should see #{container_hostname_tempfile.path} exists at this point.")
+      container_hostname_tempfile.unlink
+      r = Common.has_designated_hostnamefile?
+      assert_false(r, "method has_designated_hostnamefile? should see #{container_hostname_tempfile.path} Does NOT exist at this point.")
+      Common.HostnameFilePath = nil
+      Common.Hostname = nil
+    end
+
+    def test_is_hostname_compliant?
+      TestHostnameList.each do |thno| # thno stands for TestHostname object
+        next unless thno.AddressType == :RFC1123Hostname
+        if thno.SpecCompliant then
+          assert(Common.is_hostname_compliant?(thno.Hostname),       "Common.is_hostname_compliant?(#{thno.Hostname}) should be true.")
+        else
+          assert_false(Common.is_hostname_compliant?(thno.Hostname), "Common.is_hostname_compliant?(#{thno.Hostname}) should NOT be true.")
+        end
+      end
+    end
+
+    def test_is_like_ipv4_string?
+      TestHostnameList.each do |thno| # thno stands for TestHostname object
+        next unless thno.AddressType == :IPv4
+        if thno.SpecCompliant then
+          assert(Common.is_like_ipv4_string?(thno.Hostname),       "Common.is_like_ipv4_string?(#{thno.Hostname}) should be true.")
+        else
+          assert_false(Common.is_like_ipv4_string?(thno.Hostname), "Common.is_like_ipv4_string?(#{thno.Hostname}) should NOT be true.")
+        end
+      end
+    end
+
+    def test_is_like_ipv6_string?
+      TestHostnameList.each do |thno| # thno stands for TestHostname object
+        next unless thno.AddressType == :IPv6
+        if thno.SpecCompliant then
+          assert(Common.is_like_ipv6_string?(thno.Hostname),       "Common.is_like_ipv6_string?(#{thno.Hostname}) should be true.")
+        else
+          assert_false(Common.is_like_ipv6_string?(thno.Hostname), "Common.is_like_ipv6_string?(#{thno.Hostname}) should NOT be true.")
+        end
+      end
+    end
+
+    def test_look_for_socket_class_host_address
+      Common.Hostname = nil
+      Common.look_for_socket_class_host_address
+      hostname = Common.GetHostnameInternalForTest
+      assert(hostname.length > 0, "Hostname internal should be non-zero length.")
+    end
+
+    def test_look_in_designated_hostnamefile
+      container_hostname_tempfile = Tempfile.new('containerhostname')
+      test_hostname = 'test_hostname'
+      Common.HostnameFilePath = container_hostname_tempfile.path
+      File.write(container_hostname_tempfile.path, test_hostname)
+      Common.Hostname = nil
+
+      Common.look_in_designated_hostnamefile
+      hostname = Common.GetHostnameInternalForTest
+      assert_equal(test_hostname, hostname, "Hostname internal should be #{test_hostname} file")
+
+      # Remove container host name file
+      container_hostname_tempfile.unlink
+      Common.Hostname = nil
+      Common.HostnameFilePath = nil
+    end
+
+    def test_validate_hostname_equivalent
+      TestHostnameList.each do |thno| # thno stands for TestHostname object
+        next unless thno.AddressType == :IPv6
+        if thno.SpecCompliant then
+          assert_nothing_raised do
+              Common.validate_hostname_equivalent(thno.Hostname)
+          end
+        else
+          assert_raise NameError do
+              Common.validate_hostname_equivalent(thno.Hostname)
+          end
+        end
+      end
+    end
+
+    # End:  Tests on Public Internals
 
     def test_get_os_full_name()
       File.write(@tmp_conf_file.path, @@OSConf)
@@ -205,8 +361,40 @@ module OMS
       Common.Hostname = nil
       container_hostname_tempfile.unlink
       hostname = Common.get_hostname
-      test_hostname = Socket.gethostname.split(".")[0]
+      test_hostname = Socket.gethostname
       assert_equal(test_hostname, hostname, 'get_hostname should read from Socket.gethostname')
+    end
+
+    def test_get_hostname_as_mocks_from_socket_gethostname
+      result_hostname = nil
+      TestHostnameList.each do |thno| # thno stands for TestHostname object
+        $log = MockLog.new
+        flexmock(Socket, :strict, Socket => :gethostname, :gethostname => thno.Hostname) do
+          load "#{ENV['BASE_DIR']}/source/code/plugins/oms_common.rb"
+          assert_nothing_raised do
+            result_hostname = Common.get_hostname
+          end
+          if thno.Hostname.nil? then
+              assert_equal(result_hostname, '', 'Did not get correct hostname')
+          else
+              assert_equal(result_hostname, thno.Hostname, 'Did not get correct hostname')
+          end
+          if thno.SpecCompliant then
+            assert($log.logs.empty?, "No exception should be logged")
+          else
+            assert_false($log.logs.empty?, "Error should be logged")
+            # Checks on error logs for the validation specifics may be appropriate, but leaving
+            # out for now until there is a task specifying the log text so we don't generate
+            # unknown issues.
+            assert($log.logs[-1].include?('did NOT validate as compliant.'), "Warning in log: #{$log.logs}")
+          end
+        end
+        flexmock(Socket).flexmock_teardown
+        sleep 0.1
+      end
+      # Then back to normal (a validation afterwards would be better, but that is another project)
+      flexmock(Socket).flexmock_teardown
+      load "#{ENV['BASE_DIR']}/source/code/plugins/oms_common.rb"
     end
 
     def test_get_fqdn
@@ -318,7 +506,7 @@ module OMS
       record["DataItems"] = [ {"Message" => "iPhone\xAE"} ];
       parsed_record = Common.parse_json_record_encoding(record);
       assert_equal("{\"DataItems\":[{\"Message\":\"iPhone®\"}]}", parsed_record, "parse json record utf-8 encoding failed");
-    end	
+    end
     
     def test_safe_dump_simple_hash_array_noerror
       $log = MockLog.new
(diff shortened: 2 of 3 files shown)


source/code/plugins/oms_common.rb
```diff
@@ -478,6 +480,141 @@ class Common
...
 
     class << self
+
+      # Internal methods
+      # (left public for easy testing, though protected may be better later)
+
+      def clean_hostname_string(hnBuffer)
+        s = hnBuffer.strip
+        # Take off any port numbers:
+        if /\A(\h{4}:\h{4}:\h{4}:\h{4}:\h{4}:\h{4}:\h{4}:\h{4}):\d+\z/ =~ s
```
lagalbra: Given that this regex is very similar to the one used in `is_like_ipv6_string()`, we should try to unify them. Can we make it a constant regex with the matching group and use it in both?
Otherwise, at least add a comment that this regex is intended to match an IPv6 address; it is not intuitive.

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +480,141 @@ class Common
...
+            return $1
+        elsif not is_like_ipv6_string?(hnBuffer) and /\A(.*):\d+\z/ =~ s
+            return $1
+        end
+        return s
+      end
+
+      def has_Designated_HostnameFile?
+        return false if @@HostnameFilePath.nil?
+        return false if @@HostnameFilePath.length < 8 # 8 is first Guess.  Need policy here.
```
lagalbra: Please explain the need for this line. We currently don't have a policy safeguarding against our own file path strings.

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +480,141 @@ class Common
...
+        # Take off any port numbers:
+        if /\A(\h{4}:\h{4}:\h{4}:\h{4}:\h{4}:\h{4}:\h{4}:\h{4}):\d+\z/ =~ s
+            return $1
+        elsif not is_like_ipv6_string?(hnBuffer) and /\A(.*):\d+\z/ =~ s
+            return $1
+        end
+        return s
+      end
+
+      def has_Designated_HostnameFile?
```
lagalbra: Please use standard casing: `has_designated_hostname_file`

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +480,141 @@ class Common
...
+        return false
+      end
+
+      def is_hostname_compliant?(hnBuffer)
+        # RFC 2181:
+        #   Size limit is 1 to 63 octets, so probably bytesize is appropriate method.
+        return false if hnBuffer.nil?
+        return false if /\./ =~ hnBuffer # Hostname by definition may not contain a dot.
+        return false if /:/ =~ hnBuffer # Hostname by definition may not contain a colon.
+        return true if 1 <= hnBuffer.bytesize && hnBuffer.bytesize <= 63
```
lagalbra: This might read better if this was also a "return false" case, allowing the last line of the method to have a unique return:
```
return false if 1 > hnBuffer.bytesize  || hnBuffer.bytesize > 63
return true
```

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +480,141 @@ class Common
...
+        return false
+      end
+
+      def is_like_ipv4_string?(hnBuffer)
+        return false unless /\A[12]?\d{1,2}\.[12]?\d{1,2}\.[12]?\d{1,2}\.[12]?\d{1,2}\z/ =~ hnBuffer
```
lagalbra: For reference, where did you find this regex? Are we sure it won't exclude valid cases?

lagalbra: If someday the "Computer/Hostname" fields are used for IPs of network devices forwarding data, this may need to be more broad. Simply validating that it has four parts of all digits should be our bar because of strange cases like this.

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +480,141 @@ class Common
...
+            return false if i == 0 and bi == 0
+            i += 1
+        end
+        return false unless qwa[0].to_i != 0
+        return true
+      end
+
+      def is_like_ipv6_string?(hnBuffer)
+        return true if /\A\h{4}:\h{4}:\h{4}:\h{4}:\h{4}:\h{4}:\h{4}:\h{4}\z/ =~ hnBuffer
+        return false
```
lagalbra: Why don't you do extra-thorough checks in this method like in `is_like_ipv4_string?()` ?

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +480,141 @@ class Common
...
+        if is_dot_separated_string?(hnb) then
+          if is_like_ipv4_string?(hnb) then
+            # If Socket.gethostname returns an IPv4 address,
+            # use the entire string.
+            @@Hostname = hnb
+          else
+            # Here we presume we got a fully qualified domain name.
+            labelarray = hnb.split(".")
+            if is_recognized_prefix_label?(labelarray[0]) then
+              # Presume first label is a catagory aspect like www, and use the second.
```
lagalbra: Can you provide more detail as to why we're checking that an FQDN starts with "www"? When is this the case?

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +480,141 @@ class Common
...
+        return false
+      end
+
+      def is_recognized_prefix_label?(fqdnLabel)
+        s = fqdnLabel.strip.downcase
+        return true if s == 'www'
+        return false
+      end
+
+      def look_for_sockety_host_address
```
lagalbra: nit: sockety?

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +480,141 @@ class Common
...
+        hnb = nil
+
+        begin
+          socket_hostname = Socket.gethostname
+          hnb = socket_hostname.strip
```
lagalbra: This line should be outside of the error catch.

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +480,141 @@ class Common
...
+        end
+
+        if is_dot_separated_string?(hnb) then
+          if is_like_ipv4_string?(hnb) then
+            # If Socket.gethostname returns an IPv4 address,
+            # use the entire string.
+            @@Hostname = hnb
+          else
+            # Here we presume we got a fully qualified domain name.
+            labelarray = hnb.split(".")
```
lagalbra: Your variable-naming is a little confusing. Some are very short abbreviations like "hnb" (what does that stand for?) and some are more descriptive (labelarray). For readable code, I would much prefer longer variable names.

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +480,141 @@ class Common
...
+          end
+        else
+          # Here we presume we got a usable hostname,
+          # though it could also be an IPV6, or other things,
+          # and more attention may be needed here later.
+          @@Hostname = clean_hostname_string(hnb)
+        end
+      end
+
+      def look_in_Designated_HostnameFile
```
lagalbra: Same casing concern here: `look_in_designated_hostname_file`

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +480,141 @@ class Common
...
+            return
+        end
+
+        begin
+          hnb = File.read(@@HostnameFilePath)
+          @@Hostname = clean_hostname_string(hnb)
+        rescue => error
+          OMS::Log.warn_once("Unable to read the hostname from #{@@HostnameFilePath}: #{error}")
+        end
+
```
lagalbra: Remove this extra line

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +480,141 @@ class Common
...
+        #   omsagent will read hostname from shared file.
+
+        unless File.readable?(@@HostnameFilePath)
+            OMS::Log.warn_once("File '#{@@HostnameFilePath}' exists but is not readable.")
+            return
+        end
+
+        begin
+          hnb = File.read(@@HostnameFilePath)
+          @@Hostname = clean_hostname_string(hnb)
```
lagalbra: This line should not be in error rescue block

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +480,141 @@ class Common
...
+
+      end
+
+      def validate_hostname_equivalent(hnBuffer)
+        # RFC 2181
+        rfcl = "RFC 2181 specified hostname range of {1,63} octets for non-root item."
+        return if is_hostname_compliant?(hnBuffer)
+        return if is_like_ipv4_string?(hnBuffer) 
+        return if is_like_ipv6_string?(hnBuffer)
+        msg = "Hostname '#{hnBuffer}' length of #{hnBuffer} out of range (#{rfcl})."
```
lagalbra: You print hnBuffer twice here, and not the length of hnBuffer.

source/code/plugins/oms_common.rb
```diff
@@ -575,34 +712,22 @@ def get_os_version(conf_path = "/etc/opt/microsoft/scx/conf/scx-release")
...
+            look_for_sockety_host_address      unless is_hostname_compliant?(@@Hostname)
 
-        begin
-          hostname = Socket.gethostname.split(".")[0]
-        rescue => error
-          OMS::Log.error_once("Unable to get the Host Name: #{error}")
-        else
-          @@Hostname = hostname
+            unless is_hostname_compliant?(@@Hostname)
+            # Give up, and take whatever the shell's got.
```
lagalbra: Indent this comment

source/code/plugins/oms_common.rb
```diff
@@ -575,34 +712,22 @@ def get_os_version(conf_path = "/etc/opt/microsoft/scx/conf/scx-release")
...
+              @@Hostname = clean_hostname_string(`hostname`)
+            end
         end
+        validate_hostname_equivalent(@@Hostname)
```
lagalbra: If this raises a non-OMS-specific error, FluentD will catch it and the data being handled to be sent to OMS will not be sent. If you intend to allow errors being thrown from this validate routine, then please test what happens to data that is being handled when the error is thrown. It is not acceptable if the data gets dropped.

source/code/plugins/oms_common.rb
```diff
@@ -575,34 +712,22 @@ def get_os_version(conf_path = "/etc/opt/microsoft/scx/conf/scx-release")
         return @@OSVersion
       end
 
-      def get_hostname
-        return @@Hostname if !@@Hostname.nil?
+      def get_hostname(ignoreOldValue=false)
+        # Notes:  This implementation does not try to address multi-byte character aspects,
+        # but merely tries to address simple exigencies.  There may be multi-byte issues
+        # we will need to address in the future here.
+        if ! is_hostname_compliant?(@@Hostname) or ignoreOldValue then
```
lagalbra: style nit: place ! directly next to is_hostname_compliant?: `!is_hostname_compliant?(@@Hostname)`

test/code/plugins/oms_common_test.rb
```diff
@@ -94,6 +100,256 @@ def teardown
       "OSAlias=UniversalR\n" \
       "OSManufacturer=Central Logistics GmbH\n"
 
+    # Begin:  Unit Tests for some simple methods.
+    # May best be put in another file later.
+
+    def test_clean_hostname_string
+      hn = ' www.five.com:333     '
+      b = Common.clean_hostname_string(hn)
```
lagalbra: You can eliminate at least 4 lines in this method; you should not declare your test string as a meaningless variable rather than just directly passing it to `clean_hostname_string()`. In this case, repeated code will make people less likely to check that your test cases are correct and thorough.

test/code/plugins/oms_common_test.rb
```diff
@@ -94,6 +100,256 @@ def teardown
...
+      expected = 'www.five.com'
+      assert_equal(b,expected,"method clean_hostname_string should return '#{expected}'.")
+      hn = '        microsoft     '
+      b = Common.clean_hostname_string(hn)
+      expected = 'microsoft'
+      assert_equal(b,expected,"method clean_hostname_string should return '#{expected}'.")
+      hn = expected
+      b = Common.clean_hostname_string(hn)
+      assert_equal(b,expected,"method clean_hostname_string should return '#{expected}'.")
```
lagalbra: style nit: add spaces to all of your parameter lists

test/code/plugins/oms_common_test.rb
```diff
@@ -94,6 +100,256 @@ def teardown
       "OSAlias=UniversalR\n" \
       "OSManufacturer=Central Logistics GmbH\n"
 
+    # Begin:  Unit Tests for some simple methods.
+    # May best be put in another file later.
+
+    def test_clean_hostname_string
+      hn = ' www.five.com:333     '
```
lagalbra: Can you comment on how we know that all of these test cases (particularly www.* and IP addresses) are confirmed possibilities when `Socket.gethostname()` returns? My main concern here is that we might be writing code for an unconfirmed issue. Do we have references (besides our "10" telemetry) showing that these are anomalies we have to account for?

test/code/plugins/oms_common_test.rb
```diff
@@ -94,6 +100,256 @@ def teardown
...
+      b = Common.clean_hostname_string(hn)
+      expected = 'microsoft'
+      assert_equal(b,expected,"method clean_hostname_string should return '#{expected}'.")
+      hn = expected
+      b = Common.clean_hostname_string(hn)
+      assert_equal(b,expected,"method clean_hostname_string should return '#{expected}'.")
+    end
+
+    def test_has_Designated_HostnameFile?
+      file_hostname = '/tmp/containerhostname'
```
lagalbra: We should use Ruby's Tempfile utilities to create a file for testing here. On our pBuild machines, if more than one user runs this code, it can collide. That's why I made this change here: https://github.com/Microsoft/OMS-Agent-for-Linux/commit/8bc0525ec5b08c725665d79d69f12feedc199330#diff-3c82e3bf99b11c51027f17b4f05413ddR195

test/code/plugins/oms_common_test.rb
```diff
@@ -94,6 +100,256 @@ def teardown
...
+      assert(!r,"Common.is_dot_separated_string?(#{s}) should be true.")
+      s = ".one:two:three:four"
+      r = Common.is_dot_separated_string?(s)
+      assert(!r,"Common.is_dot_separated_string?(#{s}) should be true.")
+      s = "1234"
+      r = Common.is_dot_separated_string?(s)
+      assert(!r,"Common.is_dot_separated_string?(#{s}) should be true.")
+      s = "192:168:223:222."
+      r = Common.is_dot_separated_string?(s)
+      assert(!r,"Common.is_dot_separated_string?(#{s}) should be true.")
```
lagalbra: "192:168:223:222." is not actually a dot-separated string, since it has no content after the dot. If you want to assert something is false, it is much clearer to use `assert_false` and correlate the message.

test/code/plugins/oms_common_test.rb
```diff
@@ -94,6 +100,256 @@ def teardown
...
+    end
+
+    def test_is_hostname_compliant?
+      s = "onetwo"
+      r = Common.is_hostname_compliant?(s)
+      assert(r,"Common.is_dot_separated_string?(#{s}) should be true.")
```
lagalbra: Again, I think you've copy-pasted these tests too enthusiastically. This line should be checking if `Common.is_hostname_compliant?(s)` returns true, not `Common.is_dot_separated_string?`, as the message seems to indicate.

test/code/plugins/oms_common_test.rb
```diff
@@ -94,6 +100,256 @@ def teardown
...
+      assert(!r,"Common.is_like_ipv6_string?(#{s}) should NOT be true.")
+      s = "xx:0db8:85a3:0000:0000:8a2e:0370:7334"
+      r = Common.is_like_ipv6_string?(s)
+      assert(!r,"Common.is_like_ipv6_string?(#{s}) should NOT be true.")
+      s = ":0db8:85a3:0000:0000:8a2e:0370:7334"
+      r = Common.is_like_ipv6_string?(s)
+      assert(!r,"Common.is_like_ipv6_string?(#{s}) should NOT be true.")
+      s = "0db8:85a3:0000:0000:8a2e:0370:7334"
+      r = Common.is_like_ipv6_string?(s)
+      assert(!r,"Common.is_like_ipv6_string?(#{s}) should NOT be true.")
```
lagalbra: Each of these sections of 3 lines should be reduced as much as possible; ideally into one line for each assert, since you're doing so many of them. I would recommend making an interface for each of the methods you're trying to test, like these system tests: https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/test/installer/scripts/omsadmin_systest.rb#L214
Especially for these, a `should_succeed` parameter can be very handy. Remember, if the test fails, it will print out something like "r was expected to be true, but was false", so having a message like "it should NOT be true" is redundant. Instead, both provide a useful message to the tester and document the reason for the test by saying something like "This IPv6 address is invalid because xxxx"

test/code/plugins/oms_common_test.rb
```diff
@@ -94,6 +100,256 @@ def teardown
...
+    def test_is_recognized_prefix_label?
+      s = "www"
+      r = Common.is_recognized_prefix_label?(s)
+      assert(r,"Common.is_recognized_prefix_label?(#{s}) should be true.")
+      s = "moose"
+      r = Common.is_recognized_prefix_label?(s)
+      assert(!r,"Common.is_recognized_prefix_label?(#{s}) should NOT be true.")
+    end
+
+    def test_look_for_sockety_host_address
```
lagalbra: What exactly is this test trying to accomplish?

test/code/plugins/oms_common_test.rb
```diff
@@ -94,6 +100,256 @@ def teardown
...
+      Common.Hostname = nil
+      Common.look_for_sockety_host_address
+      hostname = Common.GetHostnameInternalForTest
+      assert(hostname.length > 0, "Hostname internal should be non-zero length.")
+    end
+
+    def test_look_in_Designated_HostnameFile
+      file_hostname = '/tmp/containerhostname'
```
lagalbra: Use Tempfile

test/code/plugins/oms_common_test.rb
```diff
@@ -94,6 +100,256 @@ def teardown
       "OSAlias=UniversalR\n" \
       "OSManufacturer=Central Logistics GmbH\n"
 
+    # Begin:  Unit Tests for some simple methods.
```
lagalbra: This entire test file is all unit tests; this comments don't clarify much.

test/code/plugins/oms_common_test.rb
```diff
@@ -209,6 +465,15 @@ def test_get_hostname_in_containerized_agent
       assert_equal(test_hostname, hostname, 'get_hostname should read from Socket.gethostname')
     end
 
+    def test_get_hostname_as_ip
+      $log = MockLog.new
+      hostname = Common.get_hostname # TODO add flexmock thing to have Socket.gethostname return an IP (or better yet, return the results of Socket.ip_address_list
```
lagalbra: If you have a TODO comment, it's not ready for review :)

source/code/plugins/oms_common.rb
v-xecamp: Yes, I agree.  I was focused on a working draft and I kept dropping that task.  Good catch.

source/code/plugins/oms_common.rb
v-xecamp: Semantics dictates it gives good information.  If that is a bad spec, I just need the good one.  I suspect it would be good to check non-zero string at least.

source/code/plugins/oms_common.rb
v-xecamp: Good point.  Usually I do that.  Thanks.

source/code/plugins/oms_common.rb
lagalbra: Sure, checking that the string has a length > 0 would be fine.

source/code/plugins/oms_common.rb
v-xecamp: That is something I came up with.  It is not all encompassing.  It will include many things that are outside the proper range, and now that I have the loop below it checking the octets, it is probably better to just to back to \d+ and just use the loop to check the ranges.  Unless you have other suggestions that's what I'll do.

source/code/plugins/oms_common.rb
v-xecamp: Yes, the format I found for ipv6 was much simpler, because \h{4} is always fixed, and near as I can tell, since it has a 1 to 1 correspondence with the possible values already, there is no further need.  If I am missing something, please don't hesitate.  I only did some review of RFCs before I got started with these drafts, and could easily be making a mistake interpreting specfications.  Also, I did not look at all for any MS specifications that may further restrict our version of the protocol.

source/code/plugins/oms_common.rb
v-xecamp: I don't mind "socket_like"  or "Ruby_Socket_Class_hostname", but was in the short run thinking sockety might be adequately illustrative.  The main thing of interest that I want to convey here is that this is the hostname information that is typical from the Ruby Socket class as limited by our parameters used.  I see this as helpful because we want to be clear for test and usage maintenance activities.

lagalbra: Makes sense. I just wasn't sure if it was a typo, or if you intended to convey "socket-like"

source/code/plugins/oms_common.rb
v-xecamp: I may be over-thinking this from experiences with sockets in the 90s programming C.  My scenario was, given it's a network situation, it could return without a proper exception, yielding a nil or some other odd class, which would then get tripped as an error when trying to run strip.  If you think that'll never happen, I'll accept your judgement.  I may just be rusty.  We could also make a separate block just for it, but it should never get to the assignment if the exception thrown is in fact proper, so the only case is error with bad data returned.

source/code/plugins/oms_common.rb
v-xecamp: label is the name of the components of the fully qualified domain name, as per readings in RFCs.  We could call it fqdn_label_array.

source/code/plugins/oms_common.rb
v-xecamp: If we are dividing out from a fqdn mistakenly conveyed to us anyway, presumably you could get something like www.myhostname.washington.edu, which is something I have seen as typical.  This anticipation may not be valid for some reason, so if so, it's a bad check.

source/code/plugins/oms_common.rb
v-xecamp: Sure.  Out of style.  Leftovers from previous shop habits.

source/code/plugins/oms_common.rb
v-xecamp: Got it.

source/code/plugins/oms_common.rb
v-xecamp: I put a "not".

source/code/plugins/oms_common.rb
lagalbra: Hmm, good point. In case of those possibilities, I would normally have a separate catch for nil cases, but you can handle it this way. I would recommend leaving a comment, or clarifying the error message; something like "Unable to get string value for hostname ..."?

source/code/plugins/oms_common.rb
lagalbra: Yes, the part I'm looking for clarification on is "something I have seen as typical" - where have you seen this?

test/code/plugins/oms_common_test.rb
v-xecamp: No.  I presumed this.  We may be able to trim out the concerns of FQDN.  Socket.gethostname should never return such a thing.  I likely went overboard here.  Rationale is as follows:

1. In my mind, I conceive both from experiences and knowledge of configurations many mis-configuration combinations of DNS and hostname.  This may just be my problem.
2.  At the very least, there is a clear justification for good logging on any situation which is out of the ordinary.  However, checking for www and FQDN inside a hostname clearly go beyond this need.
3.  Once I had the function programmed, I needed to write tests for it.  It is perfectly reasonable at this point to take the www check and the FQDN check out, as long as all odd cases are logged with warnings, I believe.

test/code/plugins/oms_common_test.rb
v-xecamp: The advantage of buffer variables is that trace prints can be embedded quickly.  In these tests extra lines do not hurt runs.  One can also re-use them in some tests for comparisons.  However, readability is certainly of concern.  I'll go through it with that in mind.  Thanks.

test/code/plugins/oms_common_test.rb
v-xecamp: Sure.

test/code/plugins/oms_common_test.rb
v-xecamp: Thanks.

test/code/plugins/oms_common_test.rb
v-xecamp: Trying "Public Internals, which is closer to what I labeled the code.  

My tendency is to collate things in alphabetical order, or functional order if that ends up being more helpful.  I may not be familiar enough with the domain, but I didn't see an existing order, and wanted to categorized somehow.  Were I refactoring these tests, I would literally put every one in alphabetical order, and probably supplement with some glossary on top of area categories, but if I could just understand an existing categorization I'd be more comfortable.

test/code/plugins/oms_common_test.rb
v-xecamp: I changed this a bit and took out literally all the FQDN stuff.  It's around, so we can bring it back later (and it was passing all the tests, too, but it still has risks, and strongly questionable need, so good for now).  Concise is good.

test/code/plugins/oms_common_test.rb
v-xecamp: I took some of these out and made better names to some.

test/code/plugins/oms_common_test.rb
v-xecamp: Good point about assert false.  I'm used to the shoulda framework.

test/code/plugins/oms_common_test.rb
lagalbra: "Socket.gethostname should never return such a thing" - this is what I wanted to know. If that is the case, then we should not work on checking for it.

test/code/plugins/oms_common_test.rb
v-xecamp: Okay, I didn't see this last until just now, and I cleaned it up the way I'm used to seeing it.  I'll keep this pattern in mind though, as I recall we are doing that elsewhere.

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +482,118 @@ class Common
...
+        return false if /:/ =~ hnBuffer # Hostname by definition may not contain a colon.
+        return false unless 1 <= hnBuffer.bytesize && hnBuffer.bytesize <= 63
+        return true
+      end
+
+      def is_like_ipv4_string?(hnBuffer)
+        return false unless /\A#{IPV4_Approximate_REGEX}\z/ =~ hnBuffer
+        qwa = hnBuffer.split('.')
+        return false unless qwa.length == 4
+        return false unless qwa[0].to_i != 0
```
lagalbra: This check has an unnecessary double-negative: to me, it reads easier as `return false if qwa[0].to_i == 0`

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +482,118 @@ class Common
...
+        i = 0
+        qwa.each do |quadwordstring|
+            bi = quadwordstring.to_i
+            # This may need more detail if 255 octets are sometimes allowed, but I don't think so.
+            return false unless 0 <= bi and bi < 255
```
lagalbra: Let's allow 255 octets; as far as I know, they are valid.

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +482,118 @@ class Common
...
+            i += 1
```
lagalbra: What does this i variable do? I don't see it being used.

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +482,118 @@ class Common
...
+
+        if is_like_ipv4_string?(hostname_buffer) then
+          # If Socket.gethostname returns an IPv4 address,
+          # use the entire string.
+          @@Hostname = hostname_buffer
+        else
+          # Here we presume we got a usable hostname,
+          # though it could also be an IPV6, or other things,
+          # and more attention may be needed here later.
+          @@Hostname = clean_hostname_string(hostname_buffer)
```
lagalbra: You already called `clean_hostname_string(hostname_buffer)` on line 545, so both the `if is_like_ipv4_string?(hostname_buffer)` case and the `else` case simply set `@@Hostname = hostname_buffer`. I believe this if/else can be replace with just the set line.

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +482,118 @@ class Common
...
+
+        begin
+          hostname_buffer = Socket.gethostname
+        rescue => error
+          OMS::Log.error_once("Unable to get the Host Name using socket facility: #{error}")
+          return
+        end
+        hostname_buffer = clean_hostname_string(hostname_buffer)
+
+        if is_like_ipv4_string?(hostname_buffer) then
```
lagalbra: Do you need the "then" keyword? I'm not sure I've seen it before in Ruby code.

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +482,118 @@ class Common
...
+        #   When omsagent runs inside a container, gethostname returns the hostname of the container (random name)
+        #   not the actual machine hostname.
+        #   One way to solve this problem is to set the container hostname same as machine name, but this is not
+        #   possible when host-machine is a private VM inside a cluster.
+        # Solution:
+        #   Share/mount ‘/etc/hostname’ as '/var/opt/microsoft/omsagent/state/containername' with container and
+        #   omsagent will read hostname from shared file.
+
+        unless File.readable?(@@HostnameFilePath)
+            OMS::Log.warn_once("File '#{@@HostnameFilePath}' exists but is not readable.")
```
lagalbra: Use two-space indents here, and everywhere (if there are any my eye has missed)

source/code/plugins/oms_common.rb
```diff
@@ -478,6 +482,118 @@ class Common
...
+      end
+
+      def validate_hostname_equivalent(hnBuffer)
+        # RFC 2181
+        rfcl = "RFC 2181 specified hostname range of {1,63} octets for non-root item."
+        return if is_hostname_compliant?(hnBuffer)
+        return if is_like_ipv4_string?(hnBuffer) 
+        return if is_like_ipv6_string?(hnBuffer)
+        msg = "Hostname '#{hnBuffer}' not compliant (#{rfcl}).  Not IP Address Either."
+        OMS::Log.error_once(msg)
```
lagalbra: I believe this should just be a warning - we will still handle this data, but the customer should be aware something is unexpected.

source/code/plugins/oms_common.rb
```diff
@@ -575,33 +691,23 @@ def get_os_version(conf_path = "/etc/opt/microsoft/scx/conf/scx-release")
         return @@OSVersion
       end
 
-      def get_hostname
-        return @@Hostname if !@@Hostname.nil?
+      def get_hostname(ignoreOldValue=false)
```
lagalbra: nit: space out default value: `ignoreOldValue = false`

source/code/plugins/oms_common.rb
```diff
@@ -575,33 +691,23 @@ def get_os_version(conf_path = "/etc/opt/microsoft/scx/conf/scx-release")
...
 
-        begin
-          if File.exist?(@@HostnameFilePath) && File.readable?(@@HostnameFilePath)
-            @@Hostname = File.read(@@HostnameFilePath).strip
-            return @@Hostname
-          end
-        rescue => error
-          OMS::Log.warn_once("Unable to read the hostname from #{@@HostnameFilePath}: #{error}")
-        end
+            look_for_socket_class_host_address      unless is_hostname_compliant?(@@Hostname)
```
lagalbra: style nit: use a single space between the method and conditional here and on line 700

test/code/plugins/oms_common_test.rb
```diff
@@ -94,6 +100,148 @@ def teardown
...
+      b = Common.clean_hostname_string(hostname)
+      expected = 'microsoft'
+      assert_equal(b, expected, "method clean_hostname_string should return '#{expected}'.")
+      hostname = expected
+      b = Common.clean_hostname_string(hostname)
+      assert_equal(b, expected, "method clean_hostname_string should return '#{expected}'.")
+    end
+
+    def test_has_designated_hostnamefile?
+      file_hostname = '/tmp/containerhostname'
```
lagalbra: Again, please use the Ruby Tempfile module, otherwise this can collide with other users on shared build machines!

test/code/plugins/oms_common_test.rb
```diff
@@ -94,6 +100,148 @@ def teardown
...
+      assert_false(Common.is_like_ipv4_string?('0.2.3.4'),         "Common.is_like_ipv4_string?('0.2.3.4') should NOT be true.")
+      assert_false(Common.is_like_ipv4_string?('1.2.3.255'),       "Common.is_like_ipv4_string?('1.2.3.255') should NOT be true.")
+      assert_false(Common.is_like_ipv4_string?('255.255.255.255'), "Common.is_like_ipv4_string?('1.2.3.255') should NOT be true.")
+      assert_false(Common.is_like_ipv4_string?('192.168.0'),       "Common.is_like_ipv4_string?('192.168.0') should NOT be true.")
+      assert_false(Common.is_like_ipv4_string?('192.168..5'),      "Common.is_like_ipv4_string?('192.168..5') should NOT be true.")
+      assert_false(Common.is_like_ipv4_string?('192.168.0.5.5'),   "Common.is_like_ipv4_string?('192.168.0.5.5') should NOT be true.")
+      assert_false(Common.is_like_ipv4_string?('192.168.0.5.A'),   "Common.is_like_ipv4_string?('192.168.0.5.5') should NOT be true.")
+    end
+
+    def test_is_like_ipv6_string?
```
lagalbra: I prefer the layout of this method to the layout of `test_is_like_ipv4_string?` or `test_is_hostname_compliant?`. Can you update those test methods to use a block with constants to check, and then a block of assert statements? It looks very clear that way.

test/code/plugins/oms_common_test.rb
```diff
@@ -94,6 +100,148 @@ def teardown
...
+      assert(Common.is_like_ipv6_string?(ipv6a2), "Common.is_like_ipv6_string?(#{ipv6a2}) should be true.")
+
+      assert_false(Common.is_like_ipv6_string?(ipv6a3), "Common.is_like_ipv6_string?(#{ipv6a3}) should be true.")
+      assert_false(Common.is_like_ipv6_string?(ipv6a4), "Common.is_like_ipv6_string?(#{ipv6a4}) should be true.")
+      assert_false(Common.is_like_ipv6_string?(ipv6a5), "Common.is_like_ipv6_string?(#{ipv6a5}) should be true.")
+      assert_false(Common.is_like_ipv6_string?(ipv6a6), "Common.is_like_ipv6_string?(#{ipv6a6}) should be true.")
+
+    end
+
+    def test_look_for_socket_class_host_address
```
lagalbra: Same question as before: What exactly is this test trying to accomplish?

test/code/plugins/oms_common_test.rb
```diff
@@ -94,6 +100,148 @@ def teardown
...
+      Common.Hostname = nil
+      Common.look_for_socket_class_host_address
+      hostname = Common.GetHostnameInternalForTest
+      assert(hostname.length > 0, "Hostname internal should be non-zero length.")
+    end
+
+    def test_look_in_designated_hostnamefile
+      file_hostname = '/tmp/containerhostname'
```
lagalbra: Again, use Tempfile

test/code/plugins/oms_common_test.rb
```diff
@@ -209,6 +357,38 @@ def test_get_hostname_in_containerized_agent
...
 
+    def test_get_hostname_mocked
+      $log = MockLog.new
+      
+      flexmock(Socket, :strict, Socket => :gethostname, :gethostname => "myhostname") do
+        hostname = Socket.gethostname
+        assert_equal("myhostname", hostname, "Could not get the hostname")
+        assert(hostname.size > 0, "Hostname returned is empty")
+        load "#{ENV['BASE_DIR']}/source/code/plugins/oms_common.rb"
+        hostname = Common.get_hostname # TODO add flexmock thing to have Socket.gethostname return an IP (or better yet, return the results of Socket.ip_address_list
```
lagalbra: Remove TODO commments that you have taken care of

lagalbra: This would also be a good place to assert that nothing gets raised, since "myhostname" is valid.

test/code/plugins/oms_common_test.rb
```diff
@@ -209,6 +357,38 @@ def test_get_hostname_in_containerized_agent
       assert_equal(test_hostname, hostname, 'get_hostname should read from Socket.gethostname')
     end
 
+    def test_get_hostname_mocked
```
lagalbra: You're mocking in `test_get_hostname_as_ipv4_address` as well, so the test name "test_get_hostname_mocked" is not precise. "test_get_hostname_as_alpha_string" might be better

source/code/plugins/oms_common.rb
v-xecamp: Ok.

v-xecamp: Just as a caveat, 255 is admonished with I think the "should" word, but is I think not yet a required top limit.

v-xecamp: Double checked.  You're right.  https://en.wikipedia.org/wiki/IPv4#Addresses_ending_in_0_or_255

source/code/plugins/oms_common.rb
v-xecamp: No longer needed.

source/code/plugins/oms_common.rb
v-xecamp: Last log is a warning on this.  I'll change this for now, but it seemed like an error that we just handled, and that might be well labeled with higher seriousness at the exception point.  I'll set this back to warn for now.

source/code/plugins/oms_common.rb
v-xecamp: Ok.

test/code/plugins/oms_common_test.rb
v-xecamp: Sorry.  Missed that first time through.

test/code/plugins/oms_common_test.rb
v-xecamp: Got it.

test/code/plugins/oms_common_test.rb
```diff
@@ -94,6 +150,106 @@ def teardown
       "OSAlias=UniversalR\n" \
       "OSManufacturer=Central Logistics GmbH\n"
 
+    # Begin:  Tests on Public internals; probably only for use inside the Singular object.
+
+    def test_clean_hostname_string
+      hostname = '        microsoft     '
```
lagalbra: Why isn't this in TestHostnameList?

v-xecamp: Sorry.  GA?  

Here are items I've got:
1.  Looks like successful install tests on CentOS 7, Ubuntu 16.04, and OpenSuSE.  
2.  I know I did have a process fallaway overnight before I started running the sequence:
sudo su -l omsagent -c '/opt/microsoft/omsconfig/Scripts/OMS_MetaConfigHelper.py --disable'
echo "trace result of metaconfig $?"
sudo su - omsagent -c '/opt/microsoft/omsconfig/Scripts/GetDscLocalConfigurationManager.py | grep RefreshMode'
---snip---
3.  It looks like this probably stays as I seem to be turning off the update from Dsc, but I only checked it once, not exhaustively, and the 36 hour test needs about 34 more hours at least.

lagalbra: GA: general availability. Our GA bundles are any which are labelled GA on our releases page: https://github.com/Microsoft/OMS-Agent-for-Linux/releases

v-xecamp: it looks like my centos install got an update somehow to its oms_common.rb under opt.  Perhaps I didn't execute the disable fast enough.  I'll try again.

v-xecamp: D'oh.  I forgot my install script stops all threads at the end of each test.  I need to administer a start to engage the 36 hour test.  It was probably fine all along.

PR #583, opened 2017-09-29
Modified log collection script to add dsc configuration folder contents

amitsara:


tools/LogCollector/source/omslinux_agentlog.py
```diff
@@ -401,6 +401,12 @@ def copyCommonFiles(omsLinuxType):
     cmd='cp /var/opt/microsoft/omsconfig/omsconfig* /tmp/omslogs'
     out=execCommand(cmd)
     writeLogCommand(cmd)
+    cmd='mkdir -p /tmp/omslogs/configuration'
```
lagalbra: Please rename this folder, or copy your logs into another subfolder with a name indicating that it is just DSC config.

amitsara: changing folder name to dscconfiguration

PR #584, opened 2017-09-30
Updating log collector .tgz after DSC config file additions

lagalbra:
Related commit: https://github.com/Microsoft/OMS-Agent-for-Linux/commit/7bfb10eed0c581fec2b9be0a6def9fca63262922

@amitsara FYI


NarineM:
This is OK to overwrite because the change is small.. but let us make sure moving forward we publish the packages under different version folder such as v2, v3 and do not overwrite previously published version but now with a different content.

PR #590, opened 2017-10-16
Don't perform omsadmin.conf check for scom workspace

vimish:
PBUILD was successful.
Verified that agent is getting onboarded to both OMS and SCOM.

### DIFF
--- a/installer/scripts/service_control
+++ b/installer/scripts/service_control
@@ -134,6 +134,18 @@ notification_exit()
 
 # End Lower Level Utilities
 
+should_check_omsadmin_conf()
+{
+    local ws_id
+    for ws_id in $NON_OMS_WS
+    do
+        if [ "$1" = "$ws_id" ]; then
+            return 1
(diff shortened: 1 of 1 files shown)


installer/scripts/service_control
```diff
@@ -145,8 +145,10 @@ setup_variables()
     fi
 
     WS_STATUS=0
-    check_omsadmin_conf $initial_conf_dir
-    WS_STATUS=$?
+    if [ $1 != "scom" ]; then
```
lagalbra: Better practice to check that $1 is not in the `NON_OMS_WS` list (my main concern is that this same check needs to be made for the "LAD" workspace as well). Other than that, this looks good.

vimish:
@lagalbra 
Changed the logic to not check omadmin.conf for all NON_OMS_WS.
Verified agent getting onboarded to both OMS and SCOM.
PBUILD successful.

PR #594, opened 2017-10-24
Make omsagent.ulinux use this_omsagent_running.

v-xecamp:
Simple modifications in omsagent.ulinux to use this_omsagent_running instead of old is_omsagent_running, which had non-standard boolean usage, and was therefore used incorrectly in at least this location.  Most additions were in new subdirectory post_install where I put test scripts I use to run install tests on various hosts in a partly automated fashion.  Adequate automated tests must be run after the bundle install.


v-xecamp:
This is broken, and not what I intended.

PR #599, opened 2017-11-01
Required change to enable OMI and DSC debug build

amal-khalaf:
modified version of omsagent/build/Makefile which have needed change for building DSC as Debug when triggering the build from the OMSagent


NarineM:
Please update to only include DSC option and rename the flag name to be only dsc. As per conversation with Jeff it is not fully supported in OMI.

NarineM:
Hi Amal, can you squash the commits into one? It is hard to review subsequent changes.

NarineM:
Can you check with OMI and see if there is anything you still depending on to be able to merge this? otherwise please close this PR.

abenbachir:
Very old PR, closing it.

PR #605, opened 2017-11-09
WLM discovery and monitors

vimish:
1. Added an input plugin wlm_discovery for discovering computer entities.
2. Updated oms_omi plugin to generate data for wlm monitors.

PBUILD was run successfully.


NarineM:
Taking this review offline .

PR #638, opened 2018-02-06
OMI binding fix

NarineM:
@niroyb 

Nic, can you take a look at this PR and let me know possible side effects of this change? I am planning to test on one dpk and one rpm package with final kit. Let me know if there is anything else I should consider before merging. thanks.


NarineM:
@jeffaco

PR #639, opened 2018-02-07
Adding troubleshooting for OMS extension

superminiek:
I have added two new procedures for my common OMS Agent issues.


docs/Troubleshooting.md
```diff
@@ -289,4 +290,33 @@ Or
 sudo sh ./onboard_agent.sh --purge
 ```
 
-You can continue re-onboarding after using the `--purge` option
\ No newline at end of file
+You can continue re-onboarding after using the `--purge` option
+
+
+### OMS extension on Azure portal is marked as failed state: Provisioning failed
```
NarineM: We recommend azure customer use OMS Linux VM extension to install omsagent and not command line tool. Please also review existing troubleshooting on omsagent extension on possible errors are documented here: https://github.com/Azure/azure-linux-extensions/tree/master/OmsAgent
And move this content to be in extension repo instead of omsagent repo

superminiek:
I mentioned about case when after installation status is failed, command line tool is in fact an installer and in this case it is the only way to change provisioning state (Azure resource health is not affected, I contacted with Azure support regarding that). I think that all my changes are some kind of relation between these two repositories but of course it can be moved there.

PR #654, opened 2018-03-02
Update onboard_agent.sh to fix #653

### LINKED ISSUE #653: Install via onboard_agent.sh fails if run as root without sudo installed
The script assumes `sudo` will be installed, so if the script is run as root, it still tries to use `sudo` and will cause the script to error out and fail.

Reference:

https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/scripts/onboard_agent.sh#L93
`    wget -O ${BUNDLE_X64} ${GITHUB_RELEASE}${BUNDLE_X64} && sudo sh ./${BUNDLE_X64} ${bundleParameters}`

https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/scripts/onboard_agent.sh#L96
`    wget -O ${BUNDLE_X86} ${GITHUB_RELEASE}${BUNDLE_X86} && sudo sh ./${BUNDLE_X86} ${bundleParameters}`

jasonrush:
Only using sudo if not running as root

### DIFF
--- a/installer/scripts/onboard_agent.sh
+++ b/installer/scripts/onboard_agent.sh
@@ -87,11 +87,17 @@ if [ -n "$proxyConf" ]; then
     bundleParameters="${bundleParameters} -p $proxyConf"
 fi
 
+# We need to use sudo for commands in the following block, if not running as root
+SUDO=''
+if (( $EUID != 0 )); then
+    SUDO='sudo'
+fi
+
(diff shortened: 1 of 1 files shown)


Zeromax:
Is there a plan for a new release?

shpimpal:
Yes, next release is planned for mid-April.

PR #658, opened 2018-03-08
 OMI refactoring to install OMI package directly from omi-kits

shpimpal:
Build changes:
  * Include OMI package from omi-kits repo directly
  * Extract SCX, apache and mysql providers from SCX bundle and package with OMS bundle

 Installation changes:
  * Install OMI and above providers from the packages instead of thorugh SCX bundle

@NarineM @jeffaco

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -6,6 +6,7 @@
 SHELL=/bin/bash
 BASE_DIR := $(subst /build,,$(CURDIR))
 OMI_ROOT := $(shell cd ../../omi/Unix; pwd -P)
+OMI_KITS := $(shell cd ../../omi-kits; pwd -P)
 SCX_CORE_KITS := $(shell cd ../../scxcore-kits; pwd -P)
 PAL_DIR := $(shell cd ../../pal; pwd -P)
 DSC_DIR := $(shell cd ../../dsc; pwd -P)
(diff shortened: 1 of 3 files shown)


installer/bundle/bundle_skel.sh
```diff
@@ -899,6 +973,19 @@ if [ -n "${checkVersionAndCleanUp}" ]; then
...
+    versionAvailable=`getVersionNumber $OMI_PKG omi-`
+    if shouldInstall_omi; then shouldInstall="Yes"; else shouldInstall="No"; fi
+    printf '%-15s%-15s%-15s%-15s\n' omi $versionInstalled $versionAvailable $shouldInstall
+
+    # SCX package
+    versionInstalled=`getInstalledVersion scx`
+    versionAvailable=`getVersionNumber $SCX_PKG scx-`
+    if shouldInstall_scx; then shouldInstall="Yes"; else shouldInstall="No"; fi
+    printf '%-15s%-15s%-15s%-15s\n' scx $versionInstalled $versionAvailable $shouldInstall
+
```
jeffaco: Hmm. I'm not sure what you should do about layered packages from SCX, which you did take care to handle when debundling the package, but you're not handling here.

I think it would make sense to run the accompanying script to see if you should install the package and, if so, then check the associated version of the package. That way, the list is more complete and has no surprises.

Thoughts?

Keep in mind that you need to run the associated script to make sure it makes sense to even install the package. If the associated script says "no", then don't bother with it here, of course.

jeffaco:
One more thing. I think the notion of installing a "list" of packages at once was troublesome for SCX, that turned out to be problematic for OMI upgrading. You might want to look at this and see if you can just do direct installs/updates for everything (like SCX did).

build/Makefile
```diff
@@ -368,7 +369,28 @@ $(INTERMEDIATE_DIR)/$(OUTPUT_PACKAGE_PREFIX).tar : $(RUBY_DEST_DIR) $(IN_PLUGINS
...
+	cd $(INTERMEDIATE_DIR)/scxbundle.*; $(COPY) 100/scx-*.universal.$(PF_ARCH).{deb,rpm} $(INTERMEDIATE_DIR)/100
+
+	# Gather apache and mysql cimprov bundle scripts from SCX
+	cd $(INTERMEDIATE_DIR)/scxbundle.*; $(COPY) *.sh $(INTERMEDIATE_DIR)/oss-kits
+
+	# Gather OMI packages from omi-kits
+	cd $(OMI_KITS); $(COPY) release/omi-*.ssl_098.ulinux.$(PF_ARCH).{deb,rpm} $(INTERMEDIATE_DIR)/098
+	cd $(OMI_KITS); $(COPY) release/omi-*.ssl_100.ulinux.$(PF_ARCH).{deb,rpm} $(INTERMEDIATE_DIR)/100
+
+	# Remove ssl_098, ssl_100 from omi filenames
```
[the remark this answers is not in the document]
shpimpal: Sorry for the merge commit in PR, happened by mistake and couldn't quickly figure out how to revert that.

About the recovery to get OMI installed, wasn't sure whether to add to OMS bundle, will add it.

OpenSSL v1.1.* support is in progress of being added to OMS agent, was planning to take care then.

Yet to run 32-bit pbuild, it's in progress.

I've raised a separate PR on Build-OMS-Agent-for-Linux repo to add omi-kits as a submodule.

installer/bundle/bundle_skel.sh
shpimpal: What do you mean by layered packages?

This part is to just print the version info all packages that this bundle script has and whether they are upgradable (yes/no).
This was done by SCX bundle script earlier for OMI and SCX, which now is removed. So substituting this code to do that.
Usage: #sh omsagent-*.sh --version-check

jeffaco: The SCX package ships more than just the SCX package. It also ships MySQL and Apache, and OMS actually has the ability to use those due to it's flexible configuration.

The way the layered packages work:

1. They ship with a run script. You run that to determine if it makes sense to even install the product (i.e. it doesn't make sense to install Apache agent if Apache isn't installed). So, if the run script says that an installation should happen, then

2. The SCX package will attempt to install layered package. See [here](https://github.com/Microsoft/SCXcore/blob/master/installer/bundle/bundle_skel_Linux.sh#L618-L632) for an example.

So, my thoughts here, when listing version information:

For each layered package that got dragged along,

1. Run the script. If the script isn't found, break out immediately (this is important, some providers run under 64-bit and it became tricky to restrict the kits from `make`).

2. If the script says "Yeah, good to go", then consider version information for layered package (and display output). If it says "No, don't install", then still print layered package but say `N/A` or something like that.

That's my thoughts anyway.

Finally:

The SCX package used to queue installs (add them to a list), and this was reverted. It turned out to not fix problems, and thus the "recovery" code was added instead. So the SCX package will just install a package when appropriate.

I think that, unless you have information to the contrary, you should likely do the same thing to OMS. We thought that installing entire sets of packages as one set helped avoid installation issues but then, when faced with hard facts to the contrary, that code was reverted. I think this reversion never made it to OMS because, by that point, OMS was transitioned to the OMS team.

Let me know if you have questions, thanks.

shpimpal: Now the apache and MySQL scripts will also get copied to OMS bundle, and will run them during installation irrespective of whether they exist on the system or not. The providers (apache and MySQL) scripts will take care of the decision to install the providers, they will skip installation of the providers if they do not exist on the system.

Here, the "--version-check" option is taking care of printing version info for OMS agent, omsconfig, OMI and SCX packages. Are you saying we print info for providers too?

About the queue installs (add them to a list), will work on it. Thanks.

build/Makefile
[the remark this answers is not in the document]
jeffaco: Your commit message has lines that are too long. Please limit commit message lines (either the summary line or the detailed lines) to around 70 bytes as per [ostc-docs](https://github.com/Microsoft/ostc-docs/blob/master/workflow-workflow.md).

installer/bundle/bundle_skel.sh
jeffaco: I'm thinking they should. But are these still shell bundles? If so, you probably shouldn't handle the details. If they're not shell bundles, then you own them.

If they are shell bundles, the shell bundles are likely too old to support `--version-check`. Otherwise, you could have just called the shell bundles themselves with the appropriate parameter.

In general, I'd like the `--version-check` parameter to specify a complete list, but it didn't do it before. If Apache and MySQL are still shell bundles, just handle them like before. If not, then you should probably add additional code (not much).

Please re-work the installation mechanism (don't queue package installation) and be certain to handle recovery the same way SCX was. Otherwise, you're likely to inject regressions. Thanks.

shpimpal: Not sure what to call them, but they are shell scripts and are handled and kept [differently]( https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/bundle/bundle_skel.sh#L958-L976) than other bundle scripts. Looks like they do support `--version-check`, but will have to take care of the [headers](https://github.com/Microsoft/SCXcore/blob/44bb092419ae2edc0ff7e40e1e883df7a805f6bc/installer/bundle/bundle_skel_Linux.sh#L452) printed as they will be printed by all the shell bundles multiple times now.

jeffaco: Run the shell script piped to `tail -1`, like this?

```
~> ./apache-cimprov-1.0.1-10.universal.1.x86_64.sh --version-check | tail -1
apache-cimprov None           1.0.1.10       Yes            
~>
```

I think that will do it, and is easier than modifying the other shell bundles.

NarineM:
Please pick up open ssl 1.1 changes I checked in today and update this PR to include ssl 1,1 packages as well. Other than that if x86 builds fine, I  can sign off.

shpimpal:
Added openssl 1.1 changes and few more changes.
x86 and x64 built fine on Jenkins.

NarineM:
@jeffaco 
Hi Jeff, we do not want to mix refactoring work not related to omi package install changes here in a single code review. Shrutika will do a separate PR for refactoring the list install.

NarineM:
Please do not forget to squash all commits into one before merging with master. I sign off.

shpimpal:
Thank you for your insightful review comments @jeffaco. A task has been created to refactor the queued installs. Did not want to mix that work here, so will create another PR for that. Thanks.

PR #662, opened 2018-03-14
Added state for saving network agent data to avoid upload duplicate info

sylar217:
The state is saved in memory and each time we receive NetworkAgent data, we compare with the state and decide whether we want to upload.

### DIFF
--- a/source/code/plugins/in_npmd_server.rb
+++ b/source/code/plugins/in_npmd_server.rb
@@ -127,7 +127,6 @@ module Fluent
             @stderrFileNameHash = Hash.new
             @stop_sync = Mutex.new
             @networkAgentState = nil
-            @netWorkAgentSync = Mutex.new
 
             # Parameters for restart backoffs
             @agent_restart_count = 0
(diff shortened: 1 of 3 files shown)


source/code/plugins/in_npmd_server.rb
```diff
@@ -125,6 +127,8 @@ def start
             @npmdIntendedStop = false
             @stderrFileNameHash = Hash.new
             @stop_sync = Mutex.new
+            @networkAgentState = nil
+            @netWorkAgentSync = Mutex.new
```
chparimi: Might not need this as per code here. The npmd_reader is single thread of execution and I don't see state being altered anywhere else.

source/code/plugins/in_npmd_server.rb
```diff
@@ -353,6 +361,30 @@ def is_valid_dataitem(item)
             end
         end
 
+        def shouldUploadNetworkAgentInfo(item)
+            _retVal = false
+            if(@networkAgentState == nil || @networkAgentState["SubnetId"] == nil || @networkAgentState["AgentIP"] == nil || @networkAgentState["AgentFqdn"] == nil || @networkAgentState["TimeGenerated"] == nil)
```
chparimi: Minor: Can we have the conditions pan multiple lines.

source/code/plugins/in_npmd_server.rb
```diff
@@ -353,6 +361,30 @@ def is_valid_dataitem(item)
...
+                _timeDiff /= 3600
+            end
+            if(!_retVal && _timeDiff > 6 || _timeDiff < 0)
+                _retVal = true
+            end
+            if(!_retVal && (@networkAgentState["SubnetId"] != item["SubnetId"] || @networkAgentState["AgentIP"] != item["AgentIP"] || @networkAgentState["AgentFqdn"] != item["AgentFqdn"]))
+                _retVal = true
+            end
+            if(_retVal)
+                @netWorkAgentSync.synchronize do
```
chparimi: Ideally if we need to protect the networkAgentState variable then have this synchronize for the logic where we are reading as well as updating it.

test/code/plugins/fake_npmd_binary.rb
```diff
@@ -6,6 +6,7 @@ class NPMDTest
     FAKE_PATH_DATA  = '{"DataItems":[{"SourceNetwork":"abcd", "SourceNetworkNodeInterface":"abcd", "SourceSubNetwork":"abcd", "DestinationNetwork":"abcd", "DestinationNetworkNodeInterface":"abcd", "DestinationSubNetwork":"abcd", "RuleName":"abcd", "TimeSinceActive":"abcd", "LossThreshold":"abcd", "LatencyThreshold":"abcd", "LossThresholdMode":"abcd", "LatencyThresholdMode":"abcd", "SubType":"NetworkPath", "HighLatency":"abcd", "MedianLatency":"abcd", "LowLatency":"abcd", "LatencyHealthState":"abcd","Loss":"abcd", "LossHealthState":"abcd", "Path":"abcd", "Computer":"abcd"}]}'
 
     FAKE_AGENT_DATA = '{"DataItems":[{"AgentFqdn":"abcd", "AgentIP":"abcd", "AgentCapability":"abcd", "SubnetId":"abcd", "PrefixLength":"abcd", "AddressType":"abcd", "SubType":"NetworkAgent", "AgentId":"abcd"}]}'
+    FAKE_AGENT_DATA2 = '{"DataItems":[{"AgentFqdn":"abcd", "AgentIP":"abcd", "AgentCapability":"abcd", "SubnetId":"abcd", "PrefixLength":"abcd", "AddressType":"abcd", "SubType":"NetworkAgent", "AgentId":"abcd"}]}'
```
chparimi: Is there any difference between FAKE_AGENT_DATA and FAKE_AGENT_DATA2? if not in the statement on line 40 we can again call puts with FAKE_AGENT_DATA

PR #667, opened 2018-03-21
Add temporary fix for systems running curl versions with default http/2

jamvar:
Add a fix that check for curl version in the system and adds "--http1.1" if needed to the RET_CODE command in omsadmin


installer/scripts/omsadmin.sh
```diff
@@ -628,13 +628,21 @@ onboard()
         esac
     fi
 
+    #This is a temporary fix for Systems with Curl versions using HTTP\2 as default
```
NarineM: Looks good. Can you add a comment about which version of curl is not working for which this fall back is needed? Aslo add a link to the article where curl version 7.4 default protocol version was announced.

PR #675, opened 2018-03-23
Adding file permission data to log

mbreakey3:
We've had a couple of issues with certain files not having the correct permissions set, so we're adding this info to our logs


mbreakey3:
@Indhukrishna @NarineM @nitinbps Can you please review when you get a chance? Thanks!

mbreakey3:
@NarineM You can merge this whenever you're ready

PR #678, opened 2018-03-27
Remove set/list installations of packages

shpimpal:
In OMS agent bundle, OMS and DSC packages used to be installed
together in a list. SCX team had observed issues due to such
installations in a list, probably due to packages being installed parallelly, which they reverted later in this [commit](https://github.com/Microsoft/SCXcore/pull/72).
Hence, removing the set/list installations of packages in OMS agent
bundle too.

@jeffaco  @NarineM


installer/bundle/bundle_skel.sh
```diff
@@ -1126,18 +1058,22 @@ case "$installMode" in
...
+               cleanup_and_exit $OMS_INSTALL_FAILED
+            fi
+
+            # Install DSC
+            if shouldInstall_omsconfig; then
+                pkg_add ${DSC_PKG} omsconfig
+                TEMP_STATUS=$?
+                if [ $TEMP_STATUS -ne 0 ]; then
+                    echo "$DSC_PKG package failed to install and exited with status $TEMP_STATUS"
+                    cleanup_and_exit $DSC_INSTALL_FAILED
```
NarineM: Looks good except I remember when installing bundle we ignored DSC package failure and will print a message saying that DSC is not installed and these functionalities would not be available. Would that still be the case with this change?

NarineM: Other than my comments above I sign off.

NarineM: Have you tested upgrade scenario with this change?

shpimpal: I forgot about the fact that we deliberately don't handle DSC failure and still continue. Do you want me to continue even if DSC fails now too?

Jagadeesh has done 64bit testing for both install and upgrade scenarios, and both worked fine. Could not test 32bit due to Jenkins build break, but for now 64bit testing should be good as changes are not specific to platform.

NarineM: let us keep it consistent with the rest of package installs. I sign off. please merge.

PR #680, opened 2018-03-27
Publish new version of log collector

NarineM:
@jamvar


mbreakey3:
@NarineM  - can you double check that the collected logs look correct? I'm not seeing all of the logs on my machine after running the collector.

PR #682, opened 2018-04-02
Fixed Update Management logging.

kjohn-msft:
Please add description and also the tests covered

### DIFF
--- a/source/code/plugins/patch_management_lib.rb
+++ b/source/code/plugins/patch_management_lib.rb
@@ -21,10 +21,6 @@ class LinuxUpdates
     SCHEDULE_NAME_VARIABLE = "SCHEDULE_NAME"
     APT_GET_START_DATE_TIME_FORMAT = "%Y-%m-%d %H:%M:%S"
 
-    def initialize(log)
-        @log = log
-     end
-
     def getAgentId()
         return OMS::Configuration.agent_id
     end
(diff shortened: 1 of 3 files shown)


test/code/plugins/patch_management_lib_test.rb
```diff
@@ -27,7 +27,7 @@ def setup
     @inventoryPath = File.join(File.dirname(__FILE__), 'InventoryWithUpdates.xml')
     LinuxUpdates.prev_hash = nil
 
-    @linuxUpdatesInstance = LinuxUpdates.new(OMS::MockLog.new)
```
Atchub: Mock Log need to be deleted

source/code/plugins/patch_management_lib.rb
```diff
@@ -278,9 +274,9 @@ def transform_and_wrap(
                     "OSFullName" => osFullName,
                     "Collections"=> collections
                 }]}
-          @log.debug "LinuxUpdates : installedPackages x #{installedPackages.size}, 
-                                        availableUpdates x #{availableUpdates.size}"
+          OMS::Log.info_once("LinuxUpdates : installedPackages x #{installedPackages.size},
```
Atchub: is it not better to leave it as debug as it will help in debugging to log it continously

Atchub: discussed with Koshy and Varad, its clarified that it would of info once and the message will be in debug after that.

abenbachir:
Very old PR, closing it.

PR #689, opened 2018-04-13
Don't upload the files which are greater than specified size.

gauravga:
This checkin, allows file upload to be controlled by settings from AMS.
The files below the specified file size will be uploaded. The specified file size will now come through 'ContentLength' in the change records returned by the resource.


vrdmr:
You'll need to change the Test.

PR #727, opened 2018-06-29
save bundle install logs

jamvar:
Save the installation logs from bundle to /var/log/omsbundle.log


abenbachir:
Very old PR, this change will cause more harm if something goes wrong. closing it.

PR #728, opened 2018-06-29
Remove support for openssl 0.9.8

shpimpal:
Azure Automation DSC team has already removed the support for openssl
0.9.8. So any machines running openssl v0.9.8 will stop working.
Removing support from oms agent too.

### DIFF
--- a/build/Makefile
+++ b/build/Makefile
@@ -40,8 +40,8 @@ SEPOLICY_DIR_EL6=$(INTERMEDIATE_DIR)/selinux.el6
 
 ifeq ($(ULINUX),1)
 # Doesn't matter what version of SSL/Ruby we use to compile or link the OMI plugin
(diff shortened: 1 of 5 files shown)


build/Makefile
```diff
@@ -40,8 +40,8 @@ SEPOLICY_DIR_EL6=$(INTERMEDIATE_DIR)/selinux.el6
 
 ifeq ($(ULINUX),1)
 # Doesn't matter what version of SSL/Ruby we use to compile or link the OMI plugin
```
abhijeetgaiha: What are we building here that doesn't require a specific SSL or Ruby?

abhijeetgaiha: Seems like the OMI input plugin, please confirm.

shpimpal: Yes, this is for OMI input plugin.

PR #749, opened 2018-07-11
Added health check script to log collector

DeveloperTommy:
Adding health check script for update management.

This will allow one to see statuses of the machine such as Network Connectivity, OMS Agent status, multihoming of workspaces, OMS admin conf, hybrid worker status, machine info, etc.


DeveloperTommy:
Looks like it's actually passing: https://travis-ci.org/Microsoft/OMS-Agent-for-Linux/requests

tools/LogCollector/source/update_mgmt_health_check.py
```diff
@@ -0,0 +1,266 @@
...
+def check_network():
+    output = []
+    output.append("Network check: \n\n")
+
+    endpoints = ["bing.com", "google.com"]
+    
+    agent_regions = ["jpe", "eus2", "cc", "scus", "uks", "sea"]
+    agent_endpoint = "-agentservice-prod-arm-1.trafficmanager.net"
+
+    jrds_regions = ["wcus", "ncus"]
```
Atchub: JRDS is available in more regions , the following blog has the list from a year back https://docs.microsoft.com/en-us/azure/log-analytics/log-analytics-oms-gateway  but please check the service xmls.

ideally, if the script can find that from conf file or where ever hybrid worker is picking from , it will need less maintenance from our end to update periodically.

tools/LogCollector/source/update_mgmt_health_check.py
```diff
@@ -0,0 +1,266 @@
...
+    ods_endpoints = ["*.ods.opinsights.azure.com", "*.oms.opinsights.azure.com", "ods.systemcenteradvisor.com"]
+
+    ff_endpoints = ["usge-jobruntimedata-prod-1.usgovtrafficmanager.net", "usge-agentservice-prod-1.usgovtrafficmanager.net", 
+                    "*.ods.opinsights.azure.us", "*.oms.opinsights.azure.us" ]
+
+    workspace = get_workspace()
+
+    for endpoint in endpoints:
+        sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        response = sock.connect_ex((endpoint, 443))
```
Atchub: we can move this to a different method and reuse accross

tools/LogCollector/source/update_mgmt_health_check.py
```diff
@@ -0,0 +1,266 @@
...
+        if len(potential_workspaces) > 1:
+            agent_status.append("OMS Agent may be multihomed. Potential workspaces: " + str(potential_workspaces) + "\n")
+
+    else:
+        agent_status.append("Directory /var/opt/microsoft/omsagent/ does not exist. Is the omsagent installed? \n")
+
+    return "".join(agent_status)
+
+def check_hybrid_worker_running():
+    output = os.popen("ps aux | grep worker").read()
```
Atchub: check if process is running can be converted to a method and we can reuse by passing a string for doing the grep

tools/LogCollector/source/update_mgmt_health_check.py
```diff
@@ -0,0 +1,266 @@
...
+def main(output_path=None):
+
+    output = []
+
+    output.append(check_network() + "\n")
+    output.append(check_oms_agent_installed() + "\n")
+    output.append(check_oms_agent_running() + "\n")
+    output.append(check_oms_admin() + "\n")
+    output.append(check_hybrid_worker_running() + "\n")
+    output.append(get_machine_info() + "\n")
```
Atchub: we can probably reorder the calls so that Machineinfo,OMSadmininfo, is omsagent installed, omsagent running, hybrid worker running,

tools/LogCollector/source/update_mgmt_health_check.py
```diff
@@ -0,0 +1,266 @@
...
+from os import walk
+
+def main(output_path=None):
+
+    output = []
+
+    output.append(check_network() + "\n")
+    output.append(check_oms_agent_installed() + "\n")
+    output.append(check_oms_agent_running() + "\n")
+    output.append(check_oms_admin() + "\n")
```
Atchub: in 50% of the incidents DSC is not running even though  agent works, so its better to add a check for that.

tools/LogCollector/source/update_mgmt_health_check.py
```diff
@@ -0,0 +1,266 @@
...
+
+    print "".join(output)
+
+def check_network():
+    output = []
+    output.append("Network check: \n\n")
+
+    endpoints = ["bing.com", "google.com"]
+    
+    agent_regions = ["jpe", "eus2", "cc", "scus", "uks", "sea"]
```
Atchub: agent end point info will be in the omsadmin.conf, see if we can pick and use that

tools/LogCollector/source/update_mgmt_health_check.py
Atchub: worker.conf will have the jrds info

tools/LogCollector/source/update_mgmt_health_check.py
```diff
@@ -0,0 +1,268 @@
...
+
+    if agent_endpoint is not None:
+        agent_endpoints.append(agent_endpoint)
+    
+    if jrds_endpoint is not None:
+        jrds_endpoints.append(jrds_endpoint)
+
+    ods_endpoints = ["*.ods.opinsights.azure.com", "*.oms.opinsights.azure.com", "ods.systemcenteradvisor.com"]
+
+    ff_endpoints = ["usge-jobruntimedata-prod-1.usgovtrafficmanager.net", "usge-agentservice-prod-1.usgovtrafficmanager.net",
```
Atchub: can we just confirm with  a security expert on exposing us gov endpoints in public file?

tools/LogCollector/source/update_mgmt_health_check.py
```diff
@@ -0,0 +1,268 @@
...
+def get_machine_info():
+    output = []
+
+    output.append("\nMachine Information: \n")
+    hostname_output = os.popen("hostnamectl").read()
+    output.append(hostname_output + "\n")
+
+    return "".join(output)
+
+def check_ff():
```
Atchub: how about mooncake?

tools/LogCollector/source/update_mgmt_health_check.py
DeveloperTommy: per email with Matt Hofacker, he linked this https://docs.microsoft.com/en-us/azure/azure-government/documentation-government-services-monitoringandmanagement . since these seem to have the FF endpoints, I think we are good.

tools/LogCollector/source/update_mgmt_health_check.py
DeveloperTommy: update management currently is not supported in mooncake, so I omitted it.

PR #754, opened 2018-07-20
Updated the WLI module to work with new service changes

vanishrao:


installer/conf/omsagent.d/wlm_baseOS.conf
```diff
@@ -55,3 +55,62 @@
   tag oms.wlm.monitor
 </source>
 
+<source>
```
vimish: Shouldn't the old <source> be removed?...this conf will emit perf data twice (one for each <source> tag.)

installer/conf/omsagent.d/wlm_baseOS.conf
```diff
@@ -55,3 +55,62 @@
...
+
+<source>
+  type wlm_oms_omi
+  object_name "Physical Disk"
+  instance_regex ".*"
+  counter_name_regex ".*"
+  interval 30s
+  omi_mapping_path %CONF_DIR_WS%/wlm_monitor_mapping.json
+  tag oms.wlm.monitor
+  wlm_ip "infrastructureInsights"
```
vimish: <nit> i in the IP name should be in caps.

vanishrao: This is one of the asks from the service to have it as a small "i" in the packet being sent from the Agent.

installer/conf/omsagent.d/wlm_baseOS.conf
vanishrao: The old source tags will be removed once the support for the old data type (which is the only difference between the two packets) is removed from the service.

PR #759, opened 2018-07-31
Updated WLI module for new service changes

vanishrao:


vanishrao:
Will open an another one.

PR #773, opened 2018-08-30
Fetching and updating Azure Resource ID in agent

shpimpal:
The load_configuration method called by all the output plugins create
a thread at first that will keep fetching and updating Azure Res ID.

### DIFF
--- a/source/code/plugins/oms_configuration.rb
+++ b/source/code/plugins/oms_configuration.rb
@@ -18,6 +18,10 @@ module OMS
     @@NotifyBlobODSEndpoint = nil
     @@OmsCloudId = nil
     @@AzureResourceId = nil
+    @@AzureIMDSEndpoint = "http://169.254.169.254/metadata/instance?api-version=2017-12-01"
+    @@AzureResIDThreadLock = Mutex.new
+    @@ProxyConfig = nil
+    @@ProxyConfigFilePath = "/etc/opt/microsoft/omsagent/proxy.conf"
     @@UUID = nil
  
     class << self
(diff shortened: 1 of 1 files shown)


source/code/plugins/oms_configuration.rb
```diff
@@ -135,7 +189,22 @@ def load_configuration(conf_path, cert_path, key_path)
...
           if line =~ /AZURE_RESOURCE_ID/
-            @@AzureResourceId = line.sub("AZURE_RESOURCE_ID=","").strip
+            azure_resource_id = get_azure_resid_from_imds()
+
+            retries = 0
+            until !azure_resource_id.nil? do
+              azure_resource_id = get_azure_resid_from_imds()
+              retries += 1
+              break if retries > 2
+              sleep 2
```
rishabh1jain: Consider exponential backoff

source/code/plugins/oms_configuration.rb
```diff
@@ -75,11 +79,61 @@ def parse_proxy_config(proxy_conf_str)
...
+            res = http_req.start() do |http|
+              http.request(http_get_req)
+            end
+
+            imds_instance_json = JSON.parse(res.body)
+            if !imds_instance_json.has_key?("compute")
+              return nil  #classic vm
+            end
+
+            azure_resource_id = '/subscriptions/' + imds_instance_json['compute']['subscriptionId'] + '/resourceGroups/' + imds_instance_json['compute']['resourceGroupName'] + '/providers/Microsoft.Compute/'
```
rishabh1jain: Consider not iterating through first level key again and again

source/code/plugins/oms_configuration.rb
shpimpal: That would create delay in fluentd. Anyway, I'm planning on moving this logic to the thread function.

source/code/plugins/oms_configuration.rb
```diff
@@ -75,11 +81,75 @@ def parse_proxy_config(proxy_conf_str)
...
+            http_req.open_timeout = 3
+            http_req.read_timeout = 2
+
+            res = http_req.start() do |http|
+              http.request(http_get_req)
+            end
+
+            imds_instance_json = JSON.parse(res.body)
+
+            return nil unless imds_instance_json.has_key?("compute")  #classic vm
```
abhijeetgaiha: Check for other required values being not present or nil.

source/code/plugins/oms_configuration.rb
```diff
@@ -75,11 +81,84 @@ def parse_proxy_config(proxy_conf_str)
...
+
+            imds_instance_json = JSON.parse(res.body)
+
+            return nil if !imds_instance_json.has_key?("compute") || imds_instance_json['compute'].empty? #classic vm
+
+            imds_instance_json_compute = imds_instance_json['compute']
+
+            #guard from missing keys
+            return nil unless imds_instance_json_compute.has_key?("subscriptionId") && imds_instance_json_compute.has_key?("resourceGroupName")
+                           && imds_instance_json_compute.has_key?("virtualMachines") && imds_instance_json_compute.has_key?("virtualMachineScaleSets")
```
abhijeetgaiha: Shouldn't it be either VM or VMSS is present?

shpimpal: sorry, got the keys names wrong. The response has only 'name' and 'vmScaleSetName' keys. The 'name' is name of VM if vmScaleSetName is empty and VMSS instance name otherwise.

PR #794, opened 2018-10-08
change $METACONFIG_PY call to sudo (for hardened images)

andyt530:
Hi,

When using CIS hardened Linux images - you have a problem installing the OMS agent install because:
the user "omsagent" is marked nologin in the /etc/passwd file by the hardening:

omsagent:x:995:992:OMS agent:/var/opt/microsoft/omsagent/run:/sbin/nologin

This causes line 663 
su - $AGENT_USER -c $METACONFIG_PY  

to fail with:
This account is currently not available.

You can workaround by using sudo -u instead.
I found the suggested workaround here:

https://unix.stackexchange.com/questions/358740/this-user-is-currently-not-available-but-allow-to-run-script-by-this-user


abenbachir:
It's time for this to get merged.

PR #797, opened 2018-10-10
Updating docker bundle for provider

rashmichandrashekar:
Updating docker bundle for provider


NarineM:
No longer applicable, we are on version 1.0.0.40.

PR #828, opened 2018-10-19
Oms docker test automation scripts 

jamvar:
OMS Bundle testing with docker images


test/docker-test/omsfiles/custom.log
```diff
@@ -0,0 +1,20 @@
...
+2018-10-18 01:47:10 A purple pig and a green donkey flew a kite in the middle of the night and ended up sunburnt.
+2018-10-18 01:47:10 I am counting my calories, yet I really want dessert.
+2018-10-18 01:47:10 Cats are good pets, for they are clean and are not noisy.
+2018-10-18 01:47:10 Where do random thoughts come from?
+2018-10-18 01:47:10 I am never at home on Sundays.
+2018-10-18 01:47:10 The memory we used to share is no longer coherent.
+2018-10-18 01:47:10 She folded her handkerchief neatly.
+2018-10-18 01:47:10 The memory we used to share is no longer coherent.
+2018-10-18 01:47:10 She was too short to see over the fence.
+2018-10-18 01:47:10 Rock music approaches at high velocity.
```
hestolz: where did you find all of this 😂 @jamvar

jamvar: You have to creative sometimes. I got them using a random sentence generator from ruby libs I used previously.

test/docker-test/omsfiles/oms_run_script.py
```diff
@@ -93,82 +90,75 @@ def disable_dsc():
...
     copy_config_files()
     apache_mysql_conf()
-    generate_data()
-    disable_dsc()
-
-def copy_config_files():
-    os.system('dos2unix /home/temp/omsfiles/perf.conf \
-            && dos2unix /home/temp/omsfiles/rsyslog-oms.conf \
-            && dos2unix /home/temp/omsfiles/apache_logs.conf \
-            && dos2unix /home/temp/omsfiles/mysql_logs.conf \
```
hestolz: so now that we deleted mysql_logs.conf and don't copy anything, how do we set config?
i see we copy a mysql_logs.conf from  /etc/opt/microsoft/omsagent/sysconf/omsagent.d/mysql_logs.conf, does that have paths that work by default? (do the default log file paths match where copy the premade log files to on line 136?):

`os.system('cp /home/temp/omsfiles/mysql.log /var/log/mysql/mysql.log \
                && cp /home/temp/omsfiles/error.log /var/log/mysql/error.log \
                && cp /home/temp/omsfiles/mysql-slow.log /var/log/mysql/mysql-slow.log \
                && cp /home/temp/omsfiles/custom.log /var/log/custom.log')`

just want to clarify

jamvar: The default path for the log file is the same as we are setting it up, so there's no need for mysql_logs.conf file any more to be copied from the host. It can just copy the default one present on the container when it get's installed.

hestolz: Perfect, just wasn't immediately sure

test/docker-test/omsfiles/customlog.conf
```diff
@@ -4,7 +4,7 @@
   type sudo_tail
   path /var/log/custom.log
   pos_file /var/opt/microsoft/omsagent/state/CUSTOM_LOG_BLOB.Custom_Log_CL_<workspace-id>.pos
-  read_from_head false
```
hestolz: FYI @jamvar this caused a regression where oracle7 no longer had custom logs

hestolz: so i have reverted

jamvar: But, It has to be set to false. This change is an update and will be available in all regions next week. Going forward this will be the setting. I did see that oracle 7 custom log is not working with this set to false, but let's keep it at false and try to make it work.

PR #833, opened 2018-10-23
Update in_exec.rb patch for fluentd v0.12.40

jamvar:
Patch to remove ZombieProcess update according to the current fluentd v0.12.40


jvssvarma:
That's the only change. The 'ensure sleep interval' part is now already in the in_exec file for v0.12.40

jamvar:
Thats' the only change. The remaining is now already part of in_exec file in v0.12.40

PR #852, opened 2018-12-05
Custom Log wildcards issue fix

shpimpal:
This will fix the issue with custom log solution where files specified with wildcard pattern do not get picked up until agent restart and sometimes do not picked up at all as files are not readable to omsagent user.

- Moving the code that lists files with given pattern to "tailfilereader.rb" as we need sudo to list them

### DIFF
--- a/source/code/plugins/tailfilereader.rb
+++ b/source/code/plugins/tailfilereader.rb
@@ -20,8 +20,38 @@ module Tailscript
 
     attr_reader :paths
 
+    def file_exists(path)
+      if system("sudo test -f #{path}")
+        @log.info "Following tail of #{path}"
+        return path
+      else
+        @log.warn "#{path} does not exist. Cannot tail the file."
+        return nil
+      end
+    end
+
+    def expand_paths()
+      arr_paths = @paths.split(',').map {|path| path.strip }
+      date = Time.now
+      expanded_paths = []
+      arr_paths.each { |path|
+        path = date.strftime(path)
+        if path.include?('*')
+          Dir.glob(path).select { |p|
+            @log.info "Following tail of #{p}"
+            expanded_paths << p
+          }
+        else
+          file = file_exists(path)
+          expanded_paths << file unless file.nil?
+        end
+      }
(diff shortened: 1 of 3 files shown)


source/code/plugins/tailfilereader.rb
```diff
@@ -365,7 +395,7 @@ def read_inode
     end
   end.parse!
 
-  a = Tailscript::NewTail.new(ARGV)
+  a = Tailscript::NewTail.new(ARGV[0])
```
simathih: are we sure all callers to this script pass an array? and not only sudo was passing Array?

source/code/plugins/tailfilereader.rb
```diff
@@ -20,8 +20,38 @@ def initialize(paths)
...
+          expanded_paths << file unless file.nil?
+        end
+      }
+      return expanded_paths
+    end
+
     def start
-      start_watchers(@paths) unless @paths.empty?
+      paths = expand_paths()
+      start_watchers(paths) unless paths.empty?
```
simathih: earlier @paths passed to start_watchers was a space separated string, now it is space separated?

source/code/plugins/tailfilereader.rb
```diff
@@ -20,8 +20,38 @@ def initialize(paths)
...
+          file = file_exists(path)
+          expanded_paths << file unless file.nil?
+        end
+      }
+      return expanded_paths
+    end
+
     def start
-      start_watchers(@paths) unless @paths.empty?
+      paths = expand_paths()
```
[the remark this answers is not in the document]
shpimpal: In ruby, instance variables start with @ and local variables are plain characters (or start with _). So it should not cause any conflicts.

source/code/plugins/tailfilereader.rb
shpimpal: AFAIK, this plugin itself and its test plugin are the only ones that call this class.
What do you mean by this - "and not only sudo was passing Array?" ?

PR #892, opened 2019-03-08
Add azure region and use custom resource id if present

shkhisti:
Azure region is retrieved from IMDS. If it is not present, we will not set the header in outgoing requests.
Custom id will be set up either by environment variable or by passing in to omsadmin script.
If id from omsadmin contains microsoft.container services, we will not run the periodic resourceid refresh task.

### DIFF
--- a/source/code/plugins/oms_configuration.rb
+++ b/source/code/plugins/oms_configuration.rb
@@ -247,12 +247,10 @@ module OMS
             if @@AzureResourceId.nil? || @@AzureResourceId.empty?              
               @@AzureResourceId = line.sub("AZURE_RESOURCE_ID=","").strip
               if @@AzureResourceId.include? "Microsoft.ContainerService"
-                OMS::Log.info_once("Azure resource id in configuration file is for AKS. It will be used")    
-              end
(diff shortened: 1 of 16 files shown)


source/code/plugins/oms_configuration.rb
```diff
@@ -80,17 +81,42 @@ def parse_proxy_config(proxy_conf_str)
...
+      
       def get_azure_resid_from_imds()
           begin
             uri = URI.parse(@@AzureIMDSEndpoint)
             http_get_req = Net::HTTP::Get.new(uri, initheader = {'Metadata' => 'true'})
 
-            if @@ProxyConfig.empty?
-              http_req = Net::HTTP.new(uri.host, uri.port)
-            else
-              http_req = Net::HTTP.new(uri.host, uri.port, @@ProxyConfig[:addr], @@ProxyConfig[:port], @@ProxyConfig[:user], @@ProxyConfig[:pass])
```
simathih: if proxy shouldn't beused for IMDS, why was this specific case added initially? was there any historical reason?

PR #893, opened 2019-03-11
Use Nokogiri instead of REXML

abenbachir:
Using nokogiri will provide us 3x speedup.
I've only updated out_oms_blob, once PR is validated I can start converting other files.
Adding nokogiri-1.10.1.gem and mini_portile2-2.4.0.gem


installer/scripts/genRubyInstaller.sh
```diff
@@ -67,24 +67,26 @@ echo "%Files" > $OUTPUT_FILE
 # For each destination file, generate the appropriate lines for InstallBuilder
 
 NEW_BASE_DIR=""
+saveIFS="$IFS"; IFS=$'\n'
```
hestolz: Reasons for the changes in this file?

abenbachir: the "find" CMD with look recursively for all files/dir, if a filename contains a whitespace it will ignore the rest of the filename.
ex: ./test/my test.txt
will be : ./test/my
so ./test/my was considered as directory, this issue is very known in bash scripting and its very nasty to catch.

abenbachir:
There is also a bug in Build-OMS-Agent-for-Linux/pal/installer/InstallBuilder/installbuilder.py

Here is how to reproduce it: include a file with whitespace in ruby.data

`${{RUBY_DEST}}/lib/ruby/gems/2.4.0/doc/my test.txt;   ${{RUBY_DEST}}/lib/ruby/gems/2.4.0/doc/my test.txt;; 644; root; root`

The installbuilder.py will only look for ${{RUBY_DEST}}/lib/ruby/gems/2.4.0/doc/my instead of "${{RUBY_DEST}}/lib/ruby/gems/2.4.0/doc/my test.txt".

Right now, I've removed 2 paths containing whitespaces in ruby.data in nokogiri test folder,  it didn't cause any issue because they were only test files, but this issue should be addressed.

simathih (approved):

hestolz:

PR #899, opened 2019-03-19
Use TCP protocol in in_syslog plugin instead of UDP

abenbachir:
Replacing files for rsyslog and syslog-ng for OMS and LAD:
Replace "@" by "@@".
Replace "udp" by "tcp"

This PR should be merged after this DSC [PR#545](https://github.com/Microsoft/PowerShell-DSC-for-Linux/pull/545).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@
 /build/x86
 /test/docker-test/*[.html, .log, .pyc]
 .vscode
+.idea
--- a/installer/conf/omsagent.d/syslog.conf
+++ b/installer/conf/omsagent.d/syslog.conf
@@ -2,7 +2,7 @@
   type syslog
   port %SYSLOG_PORT%
   bind 127.0.0.1
-  protocol_type udp
+  protocol_type tcp
   tag oms.syslog
 </source>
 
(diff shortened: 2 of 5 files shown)


installer/conf/rsyslog-lad.conf
```diff
@@ -1,2 +1,2 @@
 # OMS Syslog collection for workspace LAD (SYSLOG_PORT=%SYSLOG_PORT%)
-# To be added for each facility/min_severity: <facility>.<min_severity>     @127.0.0.1:%SYSLOG_PORT%
```
abenbachir: @canfikret I know this change migh not cause any conflicts with LAD, but just asking if any issue may happend ?

installer/conf/rsyslog-lad.conf
```diff
@@ -1,2 +1,2 @@
...
+# To be added for each facility/min_severity: <facility>.<min_severity>     @@127.0.0.1:%SYSLOG_PORT%
```
simathih: why do we need the extra @?

installer/conf/omsagent.d/syslog.conf
```diff
@@ -2,7 +2,7 @@
   type syslog
   port %SYSLOG_PORT%
   bind 127.0.0.1
-  protocol_type udp
+  protocol_type tcp
```
simathih: how are you planning on verifying this change before release?

installer/conf/rsyslog-lad.conf
ers-msft: +1

.gitignore
```diff
@@ -9,3 +9,4 @@
 /build/x86
 /test/docker-test/*[.html, .log, .pyc]
 .vscode
+.idea
```
ers-msft: why?

hestolz: it was just part of the original commit
it is just .gitignore rule to ignore some IDE generated files i think

PR #900, opened 2019-03-19
doc: fix typo tcp() should be udp() 

abenbachir:


hestolz:
Why do we want this to be UDP? Thought #899 was gonna transition us to TCP.

abenbachir:
This is not related to syslog. Its Security Events that ASI are using through syslog (in different config [file](https://github.com/Microsoft/OMS-Agent-for-Linux/blob/master/installer/conf/omsagent.d/security_events.conf)). Doc says UDP, but they put tcp() in syslog-ng.

mgladi:
@abenbachir we're using UDP as the default since that was the case originally for the generic syslog. Are you transitioning syslog to TCP?
Thanks a lot for the fix BTW

abenbachir:
@mgladi indeed, we are switching to TCP to handle higher loads.

PR #902, opened 2019-03-22
Telemetry Enhancement

hestolz:
Add a new input plugin, "agent_telemetry", that sends telemetry information using the Telemetry class from agent_telemetry_script. 

Moved some code from agent_topology_request_script/agent_maintenance_script to oms_common since it is shared by both the topology request and the telemetry request.


source/code/plugins/agent_telemetry_script.rb
```diff
@@ -0,0 +1,371 @@
...
+    strongtyped_accessor :Region, String
+    strongtyped_accessor :ConfigMgrEnabled, String
+    strongtyped_accessor :ResourceUsage, AgentResourceUsage
+    strongtyped_accessor :QoS, Array # of AgentQoS
+  end
+  
+  class Telemetry
+
+    require 'json'
+    # require 'objspace'
```
hestolz: for now, still have issue with `kernel_require.rb:55:in `require': cannot load such file -- objspace (LoadError)` even after enabling the ObjectSpace extension during ruby build.

source/code/plugins/agent_telemetry_script.rb
```diff
@@ -0,0 +1,371 @@
...
+      end
+
+      if !res.nil?
+        log_info("OMS agent management service topology request response code: #{res.code}") if @verbose
+      
+        if res.code == "200"
+          log_info("OMS agent management service topology request success")
+          return 0
+        else
+          log_error("Error sending OMS agent management service topology request. HTTP code #{res.code}")
```
hestolz: to fix: still getting a 400 response.

source/code/plugins/in_agent_telemetry.rb
```diff
@@ -0,0 +1,117 @@
...
+
+    def start
+      @telemetry_script = OMS::Telemetry.new(@omsadmin_conf_path, @cert_path, @key_path,
+                                             @pid_path, @proxy_path, @os_info, @install_info, @log, false)
+
+      if @query_interval and @poll_interval
+        @query_finished = false
+        @poll_finished = false
+        @query_condition = ConditionVariable.new
+        @poll_condition = ConditionVariable.new
```
hestolz: @shkhisti I am now aware of the new issues with the heartbeats not sending on 1m interval (possibly due to ConditionVariable). Originally I just emulated the in_heartbeat_request pattern, I will alter this e.g. to use sleep instead.

source/code/plugins/agent_maintenance_script.rb
```diff
@@ -262,11 +195,31 @@ def apply_dsc_endpoint(server_resp)
       return dsc_endpoint
     end
 
+    # Update the topology and telemetry request frequencies
+    def apply_query_intervals(server_resp)
```
hestolz: need to modify now that ODS topology response will give both topology query interval and telemetry query interval.

source/code/plugins/agent_telemetry_script.rb
abenbachir: Did you include the new ruby content in ruby.data after removing objspace in configure ?
because I can see ${{RUBY_DEST}}/lib/ruby/2.4.0/${{RUBY_ARCM}}/etc.so;
You should see this path: ${{RUBY_DEST}}/lib/ruby/2.4.0/${{RUBY_ARCM}}/objspace.so;

Will be installed here: /opt/microsoft/omsagent/ruby/lib/ruby/2.4.0/x86_64-linux/objspace.so

hestolz: 🙏 thank you 🙏

source/code/plugins/agent_telemetry_script.rb
hestolz: Fixed, service side not expecting all the fields yet. Sending limited fields in the meantime...

hestolz: fixed: incorrect serialization of telemetry object

PR #904, opened 2019-03-25
Include libjemalloc.so as ruby lib dependency

abenbachir:
Add jemalloc as external dependency
Building jemalloc and including libjemalloc.so into ruby lib folder
Make jemalloc as optional, in order to enable jemalloc build use: ./configure --enable-ulinux --enable-ruby-jemalloc

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,3 +6,7 @@
 	path = source/ext/fluentd
 	url = git@github.com:fluent/fluentd.git
 	ignore = dirty
+[submodule "source/ext/jemalloc"]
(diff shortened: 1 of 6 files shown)


.gitmodules
```diff
@@ -6,3 +6,7 @@
 	path = source/ext/fluentd
 	url = git@github.com:fluent/fluentd.git
 	ignore = dirty
+[submodule "source/ext/jemalloc"]
```
simathih: how are you going to verify the jemalloc addition?

PR #906, opened 2019-03-28
Reducing CPU usage when using filter_syslog_security.rb for higher EPS

abenbachir:
Improving syslog filters
Increase out_oms buffer chunk limit to 15m
Implement faster utc to iso8601 time format
Move lock around ip_cache when modifing the cache


abenbachir:
@taglines need some reviews for these changes.

PR #937, opened 2019-05-02
vmss fix

hestolz:
Need to parse out instance number to match the schema where {instanceId} is just the VM's instance ID. 
/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Compute/virtualMachineScaleSets/{vmScaleSetName}/virtualmachines/{instanceId}

Below you can see the instanceId being properly parsed out
image

### DIFF
--- a/source/code/plugins/oms_configuration.rb
+++ b/source/code/plugins/oms_configuration.rb
@@ -147,7 +147,7 @@ module OMS
             if (imds_instance_json_compute['vmScaleSetName'].empty?)
               azure_resource_id = azure_resource_id + 'virtualMachines/' + imds_instance_json_compute['name']
             else
-              azure_resource_id = azure_resource_id + 'virtualMachineScaleSets/' + imds_instance_json_compute['vmScaleSetName'] + '/virtualMachines/' + imds_instance_json_compute['name']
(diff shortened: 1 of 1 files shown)


source/code/plugins/oms_configuration.rb
```diff
@@ -147,7 +147,7 @@ def get_azure_resid_from_imds()
             if (imds_instance_json_compute['vmScaleSetName'].empty?)
               azure_resource_id = azure_resource_id + 'virtualMachines/' + imds_instance_json_compute['name']
             else
-              azure_resource_id = azure_resource_id + 'virtualMachineScaleSets/' + imds_instance_json_compute['vmScaleSetName'] + '/virtualMachines/' + imds_instance_json_compute['name']
+              azure_resource_id = azure_resource_id + 'virtualMachineScaleSets/' + imds_instance_json_compute['vmScaleSetName'] + '/virtualMachines/' + imds_instance_json_compute['name'].split('_')[-1]
```
simathih: is there an example of how the instance name with "_" looks like in VMSS?

hestolz: image

simathih: shouldn't the array index by 1 instead of -1 then?

hestolz: -1 gives the last element, 'name' is always the vmss name + '_' + instanceid, so -1 will always give us instanceId

simathih: looks good

PR #948, opened 2019-05-14
Correcting brackets for comparision

piyushgupta12:
While running the scipt and reaching the below fixed line of comparision, we are getting an error becuase of improper syntax of if condition. Fixed the same. Tested the script on machine.


hestolz:
why comment the test?
I did not get any syntax error from (( )) myself, but [ ] is more appropriate anyways

PR #955, opened 2019-05-25
Syslog: add unix domain socket support

abenbachir:
To improve performance
Fixing issue with syslog filter


source/code/plugins/in_oms_syslog.rb
```diff
@@ -0,0 +1,313 @@
+#
```
hestolz: Does this come from existing in_syslog.rb fluentd code? If so is it identical, or what is new/changed?

abenbachir: not identical, I've added unix domain support.
At some point we need to create a branch because we can't keep patching changes like that.

source/ext/patches/fluentd/in_syslog.patch
```diff
@@ -9,12 +37,79 @@
...
++        @log.trace { "Connected syslog socket object_id=#{self.object_id}" }
++      end
++
++      def on_read(data)
++        @buffer << data
++        pos = 0
++
++        while i = @buffer.index(@delimiter, pos)
++          msg = @buffer[pos...i]
++          @callback.call(msg, @addr)
```
hestolz: whats `@addr`/where does it come from

source/ext/patches/fluentd/in_syslog.patch
```diff
@@ -9,12 +37,79 @@
...
++          pos = i + @delimiter.length
++        end
++        @buffer.slice!(0, pos) if pos > 0
```
hestolz: whats the point of doing this here, outside the loop

source/ext/patches/fluentd/in_syslog.patch
```diff
@@ -9,12 +37,79 @@
...
+         @usock = SocketUtil.create_udp_socket(@bind)
+         @usock.bind(@bind, @port)
          SocketUtil::UdpHandler.new(@usock, log, @message_length_limit, callback, @resolve_hostname)
-       else
+-      else
++      elsif @protocol_type == :tcp
++        log.info "listening syslog socket on #{@bind}:#{@port} with #{@protocol_type}"
          # syslog family add "\n" to each message and this seems only way to split messages in tcp stream
 -        Coolio::TCPServer.new(@bind, @port, SocketUtil::TcpHandler, log, "\n", callback, @resolve_hostname)
 +        Coolio::TCPServer.new(@bind, @port, SocketUtil::TcpHandler, log, @delimiter, callback, @resolve_hostname)
```
hestolz: if we already have a patch to add delimiter, is there any conflict by changing it here as well

abenbachir: I have updated the same patch, this patch will include both delimiter + unix support

source/ext/patches/fluentd/in_syslog.patch
abenbachir: fixed

source/ext/patches/fluentd/in_syslog.patch
abenbachir: That line ensure to not clear buffer entirely, only from 0 to last read position

hestolz: ah ok. so we clear up to the last read position AKA the last delimiter in the input so far

abenbachir: yes

PR #960, opened 2019-05-29
Custom logs update from fluentd's in_tail plugin

nidhanda:
1) Updated the expand_paths method of tailfilereader.rb to add more logs and checks to make sure the file exists, is readable and is not a directory.
2) Ported changes in the position file code from fluentd's in_tail plugin into the tailfilereader.rb. Changes include adding mutex locks on the position file and updating the get/set methods. 

Testing Strategy - 
1) Ran the recently added regression unit tests to make sure no functionality is broken
2) Conducted manual tests to confirm that the logs are flowing into the backend and are visible in the correct form on the portal.

### DIFF
--- a/source/code/plugins/tailfilereader.rb
+++ b/source/code/plugins/tailfilereader.rb
@@ -268,8 +268,9 @@ module Tailscript
     class PositionFile
       UNWATCHED_POSITION = 0xffffffffffffffff
 
-      def initialize(file, map, last_pos)
+      def initialize(file, file_mutex, map, last_pos)
         @file = file
+        @file_mutex = file_mutex
         @map = map
         @last_pos = last_pos
       end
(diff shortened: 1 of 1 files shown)


source/code/plugins/tailfilereader.rb
```diff
@@ -39,12 +39,31 @@ def expand_paths()
...
           Dir.glob(path).select { |p|
-            @log.info "Following tail of #{p}"
-            expanded_paths << p
+          begin
+            is_file = !File.directory?(p)
+            if File.readable?(p) && is_file
+              @log.info "Following tail of #{p}"
+              expanded_paths << p
+            elsif !File.readable?(p)
+              @log.warn "#{p} is excluded since it's unreadable or doesn't have proper permissions."
```
simathih: this logs are super useful!

source/code/plugins/tailfilereader.rb
```diff
@@ -332,31 +360,37 @@ class FilePositionEntry
...
+          @file.write "%016x" % pos
+        }
+        @pos = pos
       end
 
       def read_inode
-        @file.pos = @seek + INO_OFFSET
-        raw = @file.read(INO_SIZE)
-        raw ? raw.to_i(16) : 0
+        @inode
```
simathih: we removed this and below?

simathih: how are we going to verify these changes? will our test automation suite cover these scenarios?

simathih: also, we need to make identical change in DSC repo as these are shipped as part of nxOMSSudoCustomLog module

nidhanda: So instead of reading from file every time and calculating the inode and pos within this method, we pass in these two values to the object when we initialize it and store it in the instance variables, which are updated whenever we call the update method. So it gets rid of the redundant file reads. 
Position File entry is crucial in the tailing process and if it wasn't working correctly the logs wouldn't be tailed correctly, which is tested by the unit tests and the manual testing.
I will create a DSC repo PR once this is merged.

PR #972, opened 2019-06-06
Updated debug logging in tailfilereader.rb

nidhanda:
Added debug logging for position file and tailed log file's read buffer.
Removed redundant logs for the tailed log file's size.


nidhanda:
I've already added more debug logs to in_sudo_tail in #961

PR #998, opened 2019-07-18
Graceful shutdown: will fix data loss/duplicate (corner cases)

abenbachir:
Reduce OMS shutdown to 1 second.
Set 3 seconds timeout while root_agent is calling shutdown on each input plugin.
Send SIGTERM instead of SIGKILL to give the agent time to cleanup during shutdown.

### DIFF
--- a/build/buildRuby.sh
+++ b/build/buildRuby.sh
@@ -192,12 +192,6 @@ if [ ! -z ${RUBY_CONFIGURE_QUALS_JEMALLOC} ]; then
         sudo make install_bin install_include install_lib
         sudo ldconfig
     fi
-
-#    export PATH=${JEMALLOC_SRCDIR}/bin:$PATH
-#    LDFLAGS=\"-L${JEMALLOC_LIB_SO}\"
-#    RUBY_CONFIGURE_QUALS=( "${RUBY_CONFIGURE_QUALS[@]}" "CPPFLAGS=-I${JEMALLOC_SRCDIR}/include"  )
-#    export LD_LIBRARY_PATH=${JEMALLOC_LIBPATH}:$LD_LIBRARY_PATH
-#    export PKG_CONFIG_PATH=${JEMALLOC_LIBPATH}/pkgconfig:$PKG_CONFIG_PATH
 fi
 
 # Clean the version of Ruby from any existing files that aren't part of source
--- a/source/code/plugins/in_dsc_monitor.rb
+++ b/source/code/plugins/in_dsc_monitor.rb
@@ -102,6 +102,8 @@ OMS Settings failed – please report issue to github.com/Microsoft/PowerShell-D
       super
       @finished_check_install = true
       @finished_check_status = true
+      @thread_check_install.exit
+      @thread_check_status.exit
       @thread_check_install.join
       @thread_check_status.join
     end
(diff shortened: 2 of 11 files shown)


build/buildRuby.sh
```diff
@@ -287,6 +281,8 @@ echo "========================= Performing Building FluentD"
 cd ${FLUENTD_DIR}
 elevate bundle install --local
 elevate bundle exec rake build
+# currently 7 tests are failing
```
simathih: is this related?

source/ext/patches/fluentd/in_exec.patch
```diff
@@ -1,10 +1,59 @@
...
++
++        begin
++          # Need to reap the process because it will become a zombie process
++          Process.detach(@pid)
++        rescue #Errno::ECHILD, Errno::ESRCH, Errno::EPERM
++        end
++        # cleanup the thread
++        Thread.kill(@thread)
++        @thread.join(0.1)
+       end
```
simathih: do we really care about the data lost here? how many CRIs do we see due to this?

simathih: my worry is this is deep in fluentd engine and could cause other unintended issues if not verified properly

abenbachir: This change will only impact the shutdown sequence, if there is a crash that happen because of this change, it will not cause any issue since it will happens only during shudown.

abenbachir:
This is a graceful shutdown, where fluentd will do:
- Clean-up.
- Flush memory to buffer.
- Release sub-process to prevent zombie process.

I have an upcoming PR that needs graceful shutdown.

build/buildRuby.sh
abenbachir: Not related, I wanted to enable fluentd unittests, but there are 7 tests already failing, it doesn't have to do with this change.

abenbachir:
@hestolz with this PR, you can use the shutdown function to push to Geneva some QoS/errors.

source/code/plugins/in_oms_heartbeat.rb
```diff
@@ -32,6 +32,8 @@ def start
     def shutdown
       if @interval
         @finished = true
+        log.debug("in_oms_hearthbeat: thread state before shutdown '#{@thread.status}'")
```
hestolz: `@log.debug`

source/code/plugins/in_dsc_monitor.rb
```diff
@@ -102,7 +102,11 @@ def shutdown
       super
       @finished_check_install = true
       @finished_check_status = true
+      log.debug("in_dsc_monitor: thread 'check_install' state before shutdown was '#{@thread_check_install.status}'")
```
hestolz: `@log.debug` for both

source/code/plugins/in_agent_telemetry.rb
```diff
@@ -57,6 +57,8 @@ def start
     def shutdown
       if defined?(OMS::Configuration.telemetry_interval)
         @finished = true
+        log.debug("in_agent_telemetry: thread state before shutdown '#{@thread.status}'")
```
hestolz: `@log.debug`

source/code/plugins/in_agent_telemetry.rb
```diff
@@ -57,6 +57,8 @@ def start
...
+        @thread.exit
```
hestolz: already added in my finally-merged qos-errors pr :)

abenbachir: Great then I will remove this. :+1:

PR #1001, opened 2019-07-20
Logrotate watcher.log and extension.log files when installed via VM extension

canfikret:
Reduce disk space usage by log files.


installer/conf/logrotate.conf
```diff
@@ -15,3 +15,21 @@
     size 50k
     copytruncate
 }
+
+/var/log/azure/Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux/watcher.log {
```
simathih: just to confirm, if GA team changes this path, this won't cause logrotate to crash due to invalid path right?

canfikret: no, it wouldn't crash; missing is ok.

Of course, we'd update the path.

canfikret:
Abondoning due to merge conflicts. Will submit a new PR.

PR #1003, opened 2019-07-22
change to https form

RedSquirrelious:
Change format to enable easier Azure DevOps authentication and help with build.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,12 +1,12 @@
 [submodule "source/ext/ruby"]
 	path = source/ext/ruby
-	url = git@github.com:ruby/ruby.git
+	url = https://github.com/ruby/ruby.git
 	ignore = dirty
 [submodule "source/ext/fluentd"]
 	path = source/ext/fluentd
-	url = git@github.com:fluent/fluentd.git
+	url = https://github.com/fluent/fluentd.git
 	ignore = dirty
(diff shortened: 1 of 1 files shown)


.gitmodules
```diff
@@ -1,12 +1,12 @@
 [submodule "source/ext/ruby"]
 	path = source/ext/ruby
-	url = https://github.com/ruby/ruby.git
+	url = ../../ruby/ruby.git
```
canfikret: Not clear how git knows to fetch from github.com. Is it going to use the same host as OMS?

RedSquirrelious: Yes, it assumes that the submodule locations are relative to the main project's default remote repository.  Here is the git documentation on relative submodule URLs:  https://git-scm.com/docs/git-submodule#_commands

And here is the SO post that helped us find this option:  https://stackoverflow.com/questions/40841882/automatically-access-git-submodules-via-ssh-or-https

Also, the double backtrack brings us back to `https://github.com` (i.e., `../../{organization_name}/{repo_name}.git`) which is needed for repos outside the Microsoft GH organization but which may be confusing for repos in the Microsoft organization.  For a Microsoft project that shares the URL path `https://github.com/microsoft`, you could just reference it as a submodule with a single backtrack (i.e., `../{repo_name}.git`)

PR #1014, opened 2019-08-03
Improve TSGs for omsagent install/uninstall issues

abenbachir:


docs/Troubleshooting.md
```diff
@@ -280,6 +280,15 @@ The user will then have to edit the correct rsyslogd or syslog_ng config file an
 * Uninstall LAD extension.
 * Remove LAD files from the machine if they present in the following location: `/var/lib/waagent/Microsoft.Azure.Diagnostics.LinuxDiagnostic-<version>/` and `/var/opt/microsoft/omsagent/LAD/`
 
+#### Restoring and fixing current agent install:
+Follow these steps to recover from a bad state, like changing/removing important files under `/[etc/var]/opt/microsoft/omsagent`:
+* Extension Install:
+    * Purge the agent manually: `wget https://raw.githubusercontent.com/Microsoft/OMS-Agent-for-Linux/master/installer/scripts/onboard_agent.sh && sh onboard_agent.sh --purge`
```
nidhanda: Shouldn't #1 and #2 be flipped? I have seen an ICM where the customer installed the omsagent through extension and then removed it by purging through the shell bundle and then was unable to remove through the extensions. I think we should recommend them to first uninstall using the extension from azure portal and then purge.

abenbachir: Can you send me that ICM link ?
What about transition state issue ? flipping #1 by #2 will work on that case ?

simathih: I agree, we should purge with bundle only if uninstall from portal is stuck or fails

PR #1017, opened 2019-08-07
OMSAgent Multihoming test suite

nidhanda:
This test suite is built from the OMS Bundle Docker test suite. The files that are modified are - 
1) oms_multihoming_tests.py
2) oms_run_script.py
3) verify_e2e.py
4) omsfiles/ (directory reorganized to allow multiple workspace configs)
5) parameters.json
6) README.md
7) omsadmin.sh and service_control (edited for mulithoming to work, will be changed when the mulithoming shell bundle is out)


abenbachir:
Some files should not be pushed to git, like this:
test/multihoming-test/omsfiles/workspace_1/error.log

nidhanda:
> Some files should not be pushed to git, like this:
> test/multihoming-test/omsfiles/workspace_1/error.log

This is not a real log, it is a dummy file that is used to ingest logs into apache error logs so that the apache plugin can tail these logs and populate them in the portal.

abenbachir:
> This is not a real log, it is a dummy file that is used to ingest logs into apache error logs so that the apache plugin can tail these logs and populate them in the portal.

But you should not commit dummy files/logs to git repo, it will just pollute the repo

nidhanda:
> > This is not a real log, it is a dummy file that is used to ingest logs into apache error logs so that the apache plugin can tail these logs and populate them in the portal.
> 
> But you should not commit dummy files/logs to git repo, it will just pollute the repo
This file's content is used by the python script to add the log lines to a file that the custom log plugin tails. We add the content of this file to a file that the apache plugin is tailing so that we can see the results in the portal. Without this file, we will not be able to test the apache plugin so we need this in the repo.

test/multihoming-test/build_images.py
```diff
@@ -0,0 +1,26 @@
+"""Build all or a subset of Docker images."""
```
hestolz: I'm wondering if rather than semi-copy pasting the existing test suite, is there a clean way of just integrating it? for instance correct me if I'm wrong but the dockerfiles that you've put here seem to be the same as for the unihomed test suite; any reason you can't just reference them in place?

hestolz: I realize this is after you did a lot of work but you could check out the way that we handle the parsed options for the unihomed suite; for instance, there are options for short- or long-running, or autoupgrade, etc. What if you could run multihoming tests (and just multihoming) by calling `oms_docker_tests.py multihomed [distro list]`? If you would rather still keep the multihoming and existing code somewhat seperate, you could define 3 files
- some 'master' file that includes uni- and multi-homed test suite files, which takes all the existing + multihomed command line arguments and runs the corresponding suites
- unihomed test suite file
- multihomed test suite file

nidhanda: Yes, I understand that since this test suite is built on top of the existing uni-homed test suite, a lot of the code base is basically the same. however i had to make a lot of changes in the main files to support multihoming testing oms_multihoming_tests.py, oms_run_script.py and even in verify_e2e.py. 
1) To integrate this back to the main code base would take equivalent amount of work just because we would need to have checks for every change since each method has some few changes in it. 
2)If we think about incorporating the two into one master file, then we would still need to have to separate the oms_multihoming_tests.py, oms_run_script.py and verify_e2e.py files and the omsfiles/ structure for multihoming and change the unihomed test cases to point to one of the omsfiles/workspace directories by default to get the configs from there and the main script will only contain a few lines, either run "python oms_docker_test.py or python oms_multihoming_tests.py". This is certainly doable but will be kind of a superficial integration and is basically equivalent to us switching directories and running the two commands ourselves. We would only prevent the duplication of docker files in the two folders. I can work on this if this way is more efficient though. Shouldn't take that long to fix it in.

hestolz:

PR #1018, opened 2019-08-08
Background jobs: use disposable process to run a job (sync / async)

abenbachir:
BJobs is able run code on different (disposable) process and get results once it finished, you can choose to wait for that job to complete or not wait, and pass a callback which will be called asynchronously.

By enabling BJobs for our output plugins out_oms, out_oms_blob showed 6x memory reduction.

We can turn ON/OFF this feature from configuration for output plugins.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,6 @@ before_script:
 
 script:
   # System tests are not run because OMS-Agent-for-Linux-testconfig.git is a private repo
-  - docker run --rm -t -v /home/travis/build/Microsoft/Build-OMS-Agent-for-Linux:/home/scratch/Build-OMS abenbachir/oms-centos6-x64:latest /bin/sh -c "cd /home/scratch/Build-OMS/omsagent/build; ./configure --enable-ulinux --enable-ruby-jemalloc; make unittest; make compile_only"
-  #- docker run --rm -t -v /home/travis/build/Microsoft/Build-OMS-Agent-for-Linux:/home/scratch/Build-OMS abenbachir/oms-centos5-x64:latest /bin/sh -c "cd /home/scratch/Build-OMS/omsagent/build; make distclean; ./configure --enable-ulinux --enable-ruby-jemalloc; make"
+  - docker run --rm -t -v /home/travis/build/Microsoft/Build-OMS-Agent-for-Linux:/home/scratch/Build-OMS abenbachir/oms-centos6-x64:latest /bin/sh -c "cd /home/scratch/Build-OMS/omsagent/build; ./configure --enable-ulinux; make unittest; make compile_only"
+  #- docker run --rm -t -v /home/travis/build/Microsoft/Build-OMS-Agent-for-Linux:/home/scratch/Build-OMS abenbachir/oms-centos5-x64:latest /bin/sh -c "cd /home/scratch/Build-OMS/omsagent/build; make distclean; ./configure --enable-ulinux; make"
 
--- a/source/code/plugins/agent_telemetry_script.rb
+++ b/source/code/plugins/agent_telemetry_script.rb
@@ -177,6 +177,17 @@ module OMS
       @@qos_events.clear
     end
 
+    def self.push_back_qos_event(source, event)
+      if @@qos_events.has_key?(source)
+        if @@qos_events[source].size >= QOS_EVENTS_LIMIT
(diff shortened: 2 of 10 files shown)


.travis.yml
```diff
@@ -39,6 +39,6 @@ before_script:
 
 script:
   # System tests are not run because OMS-Agent-for-Linux-testconfig.git is a private repo
-  - docker run --rm -t -v /home/travis/build/Microsoft/Build-OMS-Agent-for-Linux:/home/scratch/Build-OMS abenbachir/oms-centos6-x64:latest /bin/sh -c "cd /home/scratch/Build-OMS/omsagent/build; ./configure --enable-ulinux --enable-ruby-jemalloc; make unittest; make compile_only"
```
simathih: are we removing jemmaloc?

simathih: or made it default?

source/code/plugins/agent_telemetry_script.rb
```diff
@@ -177,6 +177,17 @@ def self.clear
       @@qos_events.clear
     end
 
+    def self.push_back_qos_event(source, event)
+      if @@qos_events.has_key?(source)
+        if @@qos_events[source].size >= QOS_EVENTS_LIMIT
+          @@qos_events[source].shift # remove oldest qos event to cap memory use
```
simathih: why is this separate fn needed?

source/code/plugins/filter_syslog.rb
```diff
@@ -25,12 +25,32 @@ def start
     def shutdown
       super
     end
+
+    # For perf debugging
+    def check_eps()
+      if @eps_counter == nil
+        @eps_counter = 0
```
simathih: shouild this be logged only when debug flag is enabled?

source/code/plugins/oms_common.rb
```diff
@@ -1019,4 +1019,191 @@ def to_s
...
+        log "results is not a hash, results=#{results}"
+        return results
+      end
+
+      # Handling Exception
+      if results.key?(:exception) and results[:exception] != nil
+        ex_class_name = results[:exception][:class]
+        ex_msg = results[:exception][:msg]
+        ex_backtrace = results[:exception][:backtrace]
+        if Object.const_defined?(ex_class_name)
```
simathih: this requires DSC module release

source/code/plugins/out_oms.rb
```diff
@@ -121,17 +114,34 @@ def write(chunk)
...
+    # 'chunk' is a buffer chunk that includes multiple formatted
+    # NOTE! This method is called by (out_oms) plugin thread not Fluentd's main thread. So IO wait doesn't affect other plugins.
+    def write(chunk)
+      # Quick exit if we are missing something
+      if !OMS::Configuration.load_configuration(omsadmin_conf_path, cert_path, key_path)
+        raise OMS::RetryRequestException, 'Missing configuration. Make sure to onboard.'
+      end
+
+      if run_in_background
+        OMS::BackgroundJobs.instance.run_job_and_wait { self_write(chunk) }
```
simathih: are we sure there won't be any synchronization issues? for example, thread safety of "chunk"?

.travis.yml
abenbachir: made it by default

source/code/plugins/agent_telemetry_script.rb
abenbachir: I have call push_back_qos_event to push QoS event passed from background jobs.

source/code/plugins/filter_syslog.rb
abenbachir: Not really, I only use this when I need to do EPS/benchmarking, if we add an if/else it may impact EPS little bit.

source/code/plugins/oms_common.rb
abenbachir: Yes, I did some testing with shell bundle 1.9, and it was working.

source/code/plugins/out_oms.rb
abenbachir: We upload that chunk in another process, so there is no issue with synchronization.
We are going to enable this feature incrementally.

abenbachir:
We should add test coverage during shell bundle testing and extension testing.
we can enable that with "run_in_background true" in the config.

PR #1033, opened 2019-09-04
Detect cylance and disable jemalloc

abenbachir:
Make MALLOC ARENA to 2 to reduce memory fragmentation while jemalloc is disabled (when cylance is detected).


source/ext/patches/ruby/malloc_gc.patch
```diff
@@ -1,19 +1,6 @@
...
-+        /* Explicitly free all eligible pages to the kernel. */
-+        if (do_full_mark) {
-+            malloc_trim(0);
-+        }
-     }
-     gc_prof_timer_stop(objspace);
- 
-@@ -10127,6 +10131,14 @@
+--- ../source/ext/ruby/gc.c	2019-09-04 21:40:23.967143473 +0000
++++ ../source/ext/ruby/gc.c.new	2019-09-04 21:41:10.115246447 +0000
```
simathih: why the new file?

abenbachir: To be able to comment out the LD_PRELOAD

PR #1040, opened 2019-09-17
EPM validation fixes for NPMD config

vihanda:


source/code/plugins/npmd_config_lib.rb
```diff
@@ -309,8 +309,8 @@ def self.createEpmElements(epmHash)
                     _ruleHash["IngestionWorkspaceId"] = _iRule.has_key?("IngestionWorkspaceId") ? _iRule["IngestionWorkspaceId"] : String.new
                     _ruleHash["WorkspaceAlias"] = _iRule.has_key?("WorkspaceAlias") ? _iRule["WorkspaceAlias"] : String.new
                     _ruleHash["Redirect"] = "false"
-                    _ruleHash["NetTests"] = (_iRule.has_key?("NetworkThresholdLoss") and _iRule.has_key?("NetworkThresholdLatency")) ? "true" : "false"
-                    _ruleHash["AppTests"] = (_iRule.has_key?("AppThresholdLatency")) ? "true" : "false"
+                    _ruleHash["NetTests"] = (_iRule["NetworkThresholdLoss"] >= 0 and _iRule["NetworkThresholdLatency"] >= 0) ? "true" : "false"
```
sylar217: LossThreshold and LatencyThreshold can be >=-2

source/code/plugins/npmd_config_lib.rb
```diff
@@ -632,16 +632,18 @@ def self.getEpmHashFromJson(text)
...
                         _test = _h[EpmTestInfoTag][testId]
                         _rule = Hash.new
                         _rule["ID"] = testId
                         _rule["Name"] = _test["Name"]
                         _rule["Poll"] = _test["Poll"]
                         _rule["AppThresholdLoss"] = _test["AppThreshold"].has_key?("Loss") ? _test["AppThreshold"]["Loss"] : "-2"
-                        _rule["AppThresholdLatency"] = _test["AppThreshold"]["Latency"]
-                        _rule["NetworkThresholdLoss"] = _test["NetworkThreshold"]["Loss"]
-                        _rule["NetworkThresholdLatency"] = _test["NetworkThreshold"]["Latency"]
+                        _rule["AppThresholdLatency"] = _test["AppThreshold"].has_key?("Latency") ? _test["AppThreshold"]["Latency"] : "-2.0"
```
sylar217: you should also check here whethere test["AppThreshold"] is nil or not without making any assumptions that they will always be there

vihanda: Validation added.

source/code/plugins/npmd_config_lib.rb
vihanda: As discussed, NPMD Agent is not acting on NetTests and AppTests for auto detection of threshold values.

PR #1055, opened 2019-10-14
Added condition when network and application tests shouldn't be perfo…

vihanda:
…rmed by NPMD agent


source/code/plugins/npmd_config_lib.rb
```diff
@@ -735,7 +735,7 @@ def self.getERHashFromJson(text)
                     _circuitIdMap = _h[ERCircuitInfoTag]
 
                     if _privateTestMap.empty? && _microsoftTestMap.empty?
-                        Logger::logError "ER configuration rules deserialization failed.", Logger::resc
+                        Logger::logInfo "ER configuration rules deserialization failed."
```
sylar217: Should change the string to say something like no ER rules find in the config.

PR #1057, opened 2019-10-15
Plugin for VMInsights

pfrcks:


test/code/plugins/VMInsightsDataCollector_subtest.rb
```diff
@@ -382,6 +382,8 @@ def test_get_filesystems_live
             @object_under_test = DataCollector.new
             @object_under_test.baseline
             actual = @object_under_test.get_filesystems
+            filesystems = actual.collect{|fs| fs.device_name}.to_set
+            omit "Running on container" if expected.size != actual.size && filesystems.size == expected.size
```
mthompson83: This doesn't make sense to me. Assume there is a bug in the product code and it is **not** running in a container. Because of the bug, no file systems are returned in actual. The if condition will simply cause the test case to be omitted when it should be failing.

The decision to omit a test case must not depend upon the "actual" data. It must be be determined based upon the conditions found before exercising the code under test.

source/code/plugins/VMInsightsDataCollector.rb
```diff
@@ -141,6 +141,9 @@ def get_net_stats
 
         def get_disk_stats(dev)
             raise @baseline_exception if @baseline_exception
+            if dev.count('/') > 0
```
mthompson83: Not sure why this is needed? Isn't the code being updated to always store the device names without the "/dev/" initial portion?

pfrcks: In VMInsightsEngine we pass the device_name to get the disk stats. Will place the check there to pass without the dev

source/code/plugins/VMInsightsDataCollector.rb
```diff
@@ -99,17 +99,12 @@ def get_number_of_cpus
...
-                cmd = "#{df} --block-size=1 -T"
-            else
-                cmd = "#{df} --block-size=1 -T | awk '{print $2, $1, $7, $3, $5}'"
-            end
-            IO.popen(cmd, { :in => :close }) { |io|
+            IO.popen([df, "--block-size=1", "-T"], { :in => :close, :err => File::NULL }) { |io|
                 while (line = io.gets)
-                    if (line =~ /ext[234]/)
+                    a = line.split(" ")
+                    if (a[1] =~ /ext[234]/)
```
pfrcks: @mthompson83  reworked how the file system type was being checked. now only checks the correct field for filesystem type thereby reducing the need for a separate unit test

mthompson83: This is definitely an improvement over what I had originally.

Looking at it with the benefit of time away from it, please add to the mocked test case file system types "text2" and "ext3a" to make sure the product doesn't include them.

pfrcks: Will the df command ever return filesystem type that is malformed like text2 or ext3a?

source/code/plugins/VMInsightsDataCollector.rb
```diff
@@ -354,8 +346,12 @@ def get_up_net_devices
 
         class Fs
             def initialize(device_name, mount_point, size_in_bytes, free_space_in_bytes)
-                raise ArgumentError, device_name unless device_name.start_with?("/dev/")
                 raise ArgumentError, mount_point unless mount_point.start_with? "/"
+                ind = device_name.index('/', 1)
```
mthompson83: I don't see a test case where `ind` will be nil. Is that because the value of `device_name` must always begin with "/dev/"? If so, simplify the code.

test/code/plugins/VMInsightsDataCollector_subtest.rb
```diff
@@ -376,14 +376,15 @@ def test_get_filesystems_live
                 data = io.readlines
                 data.each { |line|
                     s = line.split(" ")
-                    expected[s[0]] = [ s[1], s[2].to_i, false ]
+                    ind = s[0].index('/', 1)
+                    key = s[0][ind+1..-1]
```
mthompson83: Suggestion: The code, searching for the / and using its index to remove the prefix seems to be overly general, and introduces unhandled error paths. Simplify. We know the string must start with "/dev/", and that is what we need to remove.

test/code/plugins/VMInsightsDataCollector_subtest.rb
```diff
@@ -376,14 +376,15 @@ def test_get_filesystems_live
...
+                    omit "running in container" if expected.has_key? key
```
mthompson83: Suggestion: `omit "#{key} is mounted multiple times. Running in a container." if expected.has_key? key`

test/code/plugins/VMInsightsDataCollector_subtest.rb
pfrcks: I agree with this. Have made changes to ensure uniformity of operation. Since in Linux all devices are located in /dev/ it has been made a necessary check.

test/code/plugins/VMInsightsEngine_subtest.rb
```diff
@@ -298,7 +298,6 @@ def test_null_computer
         end
 
         def test_multihome
```
mthompson83: This test is redundent

PR #1068, opened 2019-10-19
add hotfix for tcp in syslog

abenbachir:


installer/bundle/bundle_skel.sh
```diff
@@ -1357,6 +1366,7 @@ case "$installMode" in
         fi
 
         # Hotfix DSC bug for nxOMSPlugin: removing 15 files that were not cleaned up by DSC
+        echo "Applying DSC nxOMSSyslog hotfix..."
```
hestolz: + nxOMSSudoCustomLog

abenbachir: those comments are only for debugging purpose, they should be removed once we get to a stable stat.

PR #1071, opened 2019-10-23
Mithomps/task4876044/4

mthompson83:
Adding new metrics plugin for VMInsights.

### DIFF
--- a/source/code/plugins/VMInsightsDataCollector.rb
+++ b/source/code/plugins/VMInsightsDataCollector.rb
@@ -212,10 +212,10 @@ module VMInsights
                     RawDiskData.new(
                                     dev,
                                     Time.now,
-                                    data[1 - 1].to_i,
-                                    data[3 - 1].to_i,
-                                    data[5 - 1].to_i,
-                                    data[7 - 1].to_i,
+                                    data[0].to_i,
+                                    data[2].to_i,
+                                    data[4].to_i,
+                                    data[6].to_i,
                                     sector_size
                                     )
                 }
--- a/source/code/plugins/VMInsightsDataCollector.rb
+++ b/source/code/plugins/VMInsightsDataCollector.rb
@@ -1,0 +1,393 @@
+# frozen_string_literal: true
+
+require 'scanf'
+
+module VMInsights
+    require_relative 'VMInsightsIDataCollector.rb'
+
+    class DataCollector < IDataCollector
+        def initialize(root_directory_name="/")
+            @root = root_directory_name
+            @baseline_exception = RuntimeError.new "baseline has not been called"
+            @cpu_count = nil
+            @saved_net_data = nil
+            @saved_disk_data = DiskInventory.new(@root)
+        end
+
+        def baseline
+            @baseline_exception = nil
+            @cpu_count, is_64_bit = get_cpu_info_baseline
+            DataWithWrappingCounter.set_32_bit(! is_64_bit)
+            @saved_net_data = get_net_data
+            @saved_disk_data.baseline
+            u, i = get_cpu_idle
+            { :up => u, :idle => i }
+        end
+
+        def start_sample
+        end
+
+        def end_sample
+        end
+
+        def get_available_memory_kb
+            available = nil
+            total = nil
+            File.open(File.join(@root, "proc", "meminfo"), "rb") { |f|
+                begin
+                    line = f.gets
+                    next if line.nil?
+                    line.scanf("%s%d%s") { |label, value, uom|
+                        if (label == "MemTotal:" && value >= 0 && uom == "kB")
+                            total = value
+                        elsif (label == "MemAvailable:" && value >= 0 && uom == "kB")
+                            available = value
+                        end
+                    }
+                end until f.eof?
+            }
+
+            raise IDataCollector::Unavailable, "Available memory not found" if available.nil?
+            raise IDataCollector::Unavailable, "Total memory not found" if total.nil?
+
+            return available, total
+        end
+
+        # returns: cummulative uptime, cummulative idle time
+        def get_cpu_idle
+            uptime = nil
+            idle = nil
+            File.open(File.join(@root, "proc", "uptime"), "rb") { |f|
+                line = f.gets
+                next if line.nil?
+                line.scanf(" %f %f ") { |u, i|
+                    uptime = u
+                    idle = i
+                }
+            }
+
+            raise IDataCollector::Unavailable, "Uptime not found" if uptime.nil?
+            raise IDataCollector::Unavailable, "Idle time not found" if idle.nil?
+
+            return uptime, idle
+        end
+
+        # returns:
+        #   number of CPUs available for scheduling tasks
+        # raises:
+        #   Unavailable if not available
+        def get_number_of_cpus
+            raise @baseline_exception if @baseline_exception
+            raise @cpu_count if @cpu_count.kind_of? StandardError
+            @cpu_count
+        end
+
+        # return:
+        #   An array of objects with methods:
+        #       mount_point
+        #       size_in_bytes
+        #       free_space_in_bytes
+        #       device_name
+        def get_filesystems
+            result = []
+            df = File.join(@root, "bin", "df")
+            IO.popen([df, "--block-size=1", "-T"], { :in => :close, :err => File::NULL }) { |io|
+                while (line = io.gets)
+                    a = line.split(" ")
+                    if (a[1] =~ /^ext[234]$/)
+                        begin
+                            result << Fs.new(a[0], a[6], a[2], a[4]) if a.size == 7
+                        rescue ArgumentError => ex
+                            # malformed input
+                        end
+                    end
+                end
+            }
+            result
+        rescue => ex
+            raise IDataCollector::Unavailable.new ex.message
+        end
+
+        # returns:
+        #   An array of objects with methods:
+        #       device
+        #       bytes_received  since last call or baseline
+        #       bytes_sent      since last call or baseline
+        #   Note: Only devices that are "UP" or had activity are included
+        def get_net_stats
+            raise @baseline_exception if @baseline_exception
+            result = []
+            new_data = get_net_data
+            new_data.each_pair { |key, new_value|
+                previous_value = @saved_net_data[key]
+                if previous_value.nil?
+                    result << new_value if new_value.up
+                else
+                    diff = new_value - previous_value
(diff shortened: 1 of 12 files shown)


source/code/plugins/in_vminsights.rb
```diff
@@ -0,0 +1,84 @@
...
+    def configure(conf)
+        super
+
+        begin
+            @heartbeat_uploader = conf[:MockMetricsEngine] || ::VMInsights::MetricsEngine.new
+            @heartbeat_upload_configuration = make_heartbeat_configuration
+        rescue Fluent::ConfigError
+            raise
+        rescue => ex
+            @heartbeat_upload_configuration = nil
```
simathih: how are you e2e testing this?

simathih: please send us info about that as we would release the shell bundle part of the change

source/code/plugins/VMInsightsEngine.rb
```diff
@@ -0,0 +1,313 @@
...
+                super() {
+                    @cummulative_data.initialize_from_baseline @data_collector.baseline
+                    MetricTuple.computer computer
+                    begin
+                            @log.info "Starting polling loop at #{interval} second interval"
+                            while @run do
+                                    @mutex.synchronize {
+                                            @condvar.wait(@mutex, interval) if @run
+                                    }
+                                    yield_metrics_message() if @run
```
aydany: How does the code account for the time it takes to generate the metrics? It seems that the loop will always wait for interval seconds between each data collection, which is probably now what we want. Ideally, we want to collect the metrics once per 60 seconds, which means that we may need to adjust the wait time, based on how long we've spent collecting them.

mthompson83: Will change.

source/code/plugins/in_vminsights.rb
mthompson83: At this time, we have

- manually added the plugin to sample systems and restarted the omsagent,
- verified the data flowing through ODS, etc. into the workspace's Kusto tables is as expected.

Once we can get this code into DSC, we will submit the PR to connect the shell bundle version with the intelligence pack in test and verify again before going to production.

source/code/plugins/VMInsightsEngine.rb
mthompson83: Changes completed.

source/code/plugins/VMInsightsEngine.rb
```diff
@@ -0,0 +1,323 @@
...
+                        expected_wakeup_time = Time.now
+                        while @run do
+                            expected_wakeup_time += interval
+                            now = Time.now
+                            if expected_wakeup_time <= now
+                                expected_wakeup_time = now
+                            else
+                                sleep_time = expected_wakeup_time - now
+                                @mutex.synchronize {
+                                    @condvar.wait(@mutex, sleep_time) if @run
```
aydany: Sleeps of less than 15ms typically cannot be satisfied. In that case, you can buypass the sleep and go directly to data collection.

source/code/plugins/VMInsightsEngine.rb
```diff
@@ -0,0 +1,323 @@
...
+                    @cummulative_data.initialize_from_baseline @data_collector.baseline
+                    MetricTuple.computer computer
+                    begin
+                        @log.info "Starting polling loop at #{interval} second interval"
+                        # try to keep as close to the polling interval, independent of
+                        # any delay waking up from the sleep, collecting data, or
+                        # emitting the message.
+                        expected_wakeup_time = Time.now
+                        while @run do
+                            expected_wakeup_time += interval
```
aydany: This approach attempts to keep the schedule and will do just fine as long as the clock on the machine does not jump back/forward in time, as it relies on checks against Time.now every iteration.

An alternative implementation is to measure the elapsed time of data collection (using something that does not depend on Time.now but on tick differences, and then computing the sleep time. That way we can avoid the use of Time.now and be less susceptible to clock fluctuations.

That said, the data output is tagged with Time.now, so the current implementation should be good enough.

mthompson83: Yes, "large" clock changes, both forward and backwards, mess up both the polling and rate calculations. As I recall, it is also an issue for the Windows version. When the issue was identified during design, we decided to defer addressing it.

source/code/plugins/VMInsightsEngine.rb
mthompson83: So it will wake up immediately. In production the polling interval is 60 seconds. If it takes 59.985 seconds to do the data collection and queue the message, we have bigger problems.

source/code/plugins/VMInsightsIDataCollector.rb
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+
+module VMInsights
+
+    # Design assumptions:
+    # 1. Changing the number of active CPUs requires a system reboot.
```
aydany: On windows we are enumerating the number of cpus and memory size every minute.

source/code/plugins/VMInsightsEngine.rb
aydany: I don't think it will wake up immediately. It will wake up after the deadline, instead.

source/code/plugins/VMInsightsDataCollector.rb
```diff
@@ -0,0 +1,393 @@
...
+            result = []
+            df = File.join(@root, "bin", "df")
+            IO.popen([df, "--block-size=1", "-T"], { :in => :close, :err => File::NULL }) { |io|
+                while (line = io.gets)
+                    a = line.split(" ")
+                    if (a[1] =~ /^ext[234]$/)
+                        begin
+                            result << Fs.new(a[0], a[6], a[2], a[4]) if a.size == 7
+                        rescue ArgumentError => ex
+                            # malformed input
```
aydany: Should we log this?

source/code/plugins/VMInsightsDataCollector.rb
```diff
@@ -0,0 +1,393 @@
...
+        #       bytes_received  since last call or baseline
+        #       bytes_sent      since last call or baseline
+        #   Note: Only devices that are "UP" or had activity are included
+        def get_net_stats
+            raise @baseline_exception if @baseline_exception
+            result = []
+            new_data = get_net_data
+            new_data.each_pair { |key, new_value|
+                previous_value = @saved_net_data[key]
+                if previous_value.nil?
```
aydany: Is this correct? Shouldn't we skip the first sample until we have two samples to subtract? Otherwise, it seems that the first sample may a really large value.

source/code/plugins/VMInsightsDataCollector.rb
```diff
@@ -0,0 +1,393 @@
...
+                    result << new_value if new_value.up
+                else
+                    diff = new_value - previous_value
```
aydany: How are we handling non-sensical results, e.g., the new sample showing smaller values and the difference being negative?

source/code/plugins/VMInsightsDataCollector.rb
```diff
@@ -0,0 +1,393 @@
...
+                        while (line = io.gets)
+                            s = line.split(" ")
+                            next if s.length < 2
+                            if (devices.empty? || devices.include?(s[0]))
+                                result[s[0]] = s[1].to_i
+                            end
+                        end
+                    }
+                rescue Errno::ENOENT => ex
+                    # telemetry
```
aydany: Do we need a TODO comment?

source/code/plugins/VMInsightsDataCollector.rb
```diff
@@ -0,0 +1,393 @@
...
+            def get_disk_data(dev, sector_size)
+                path = File.join(@root, "sys", "class", "block", dev, "stat")
+                File.open(path, "rb") { |f|
+                    line = f.gets
+                    raise Unavailable, "#{path}: is empty" if line.nil?
+                    data = line.split(" ")
+                    RawDiskData.new(
+                                    dev,
+                                    Time.now,
+                                    data[1 - 1].to_i,
```
aydany: What is the idea of expressing the index like this, rather than 0, 2, 4, and 6?

source/code/plugins/VMInsightsDataCollector.rb
```diff
@@ -0,0 +1,393 @@
...
+        end
+
+        def get_disk_stats(dev)
+            raise @baseline_exception if @baseline_exception
+            @saved_disk_data.get_disk_stats(dev)
+        end
+
+    private
+
+        class DataWithWrappingCounter
```
aydany: I might have missed it, but do we have unit tests to validate this class?

source/code/plugins/VMInsightsIDataCollector.rb
mthompson83: To the best of my knowledge, neither the number of active CPUs nor the total memory can change without a reboot. If we can find evidence to contradict that, we'll need to also change how the calculations are done.

source/code/plugins/VMInsightsDataCollector.rb
mthompson83: Next PR

source/code/plugins/VMInsightsDataCollector.rb
mthompson83: Already in next PR

source/code/plugins/VMInsightsDataCollector.rb
mthompson83: Yes. There are 32-bit and 64-bit wide counter rollover tests of the data returned by the DataCollector.

source/code/plugins/VMInsightsDataCollector.rb
mthompson83: When the MP starts, it does a baseline. The first metrics are reported after one polling interval, so the first metrics are the delta between the first and second sample.

If there is no previous sample, that means the interface went from a down to up status during the polling interval. The activity will all be during the polling interval, so no delta computation required.

source/code/plugins/VMInsightsDataCollector.rb
mthompson83: Yes. It is built into the DataWithWrappingCounter class and its subclasses.

source/code/plugins/VMInsightsDataCollector.rb
mthompson83: The documentation I was referencing, which I failed to cite (bad on me) used one based column numbering.
Will change before check-in.

PR #1073, opened 2019-10-23
Fix install failures on redhat 6.7 due to sudo tty

abenbachir:


installer/scripts/omsadmin.sh
```diff
@@ -660,7 +660,14 @@ onboard()
             fi
 
             if [ "$USER_ID" -eq "0" ]; then
-                sudo -u $AGENT_USER $METACONFIG_PY > /dev/null || error=$?
+                su - $AGENT_USER -c $METACONFIG_PY > /dev/null || error=$?
```
simathih: do we need to set the error context?

simathih:
please verify this in all RHEL distros before release

installer/scripts/omsadmin.sh
abenbachir: we have another if/else after that also checks $error, I'm resetting it before running:
sudo -u $AGENT_USER $METACONFIG_PY > /dev/null || error=$?
If this fails it will be catched by the following condition.
            if [ $error -eq 0 ]; then
                log_info "Configured omsconfig"
            else
                log_error "Error configuring omsconfig. Error: $error"
                return $ERROR_GENERATING_METACONFIG
            fi

PR #1083, opened 2019-10-30
add no-remove flag to APT

abenbachir:


docs/CODEOWNERS
```diff
@@ -0,0 +1,3 @@
+# These owners will be the default owners for everything in
```
hestolz: This isn't really true though ... think of the plugin .rb files for stuff like changetracking, security, etc.

abenbachir: At least someone should be involved in every PR

abenbachir: This way we can track any change made to OMS repo

hestolz:
realizing this is hotfix scenario so we don't want too many changes (and this is mostly for 1 customer) but this only addresses APT, what about YUM? is there an analogous flag?

abenbachir:
No flags for yum or zypper

PR #1086, opened 2019-11-01
task 4970858 - Enhance VMInsights plugin logging

mthompson83:
Enhance the plugin logging.

- Adding debug logging in several places.
- Add functionality so that plugin logs always include a specific prefix for ease of identification.
- Added unit tests.

### DIFF
--- a/source/code/plugins/VMInsightsDataCollector.rb
+++ b/source/code/plugins/VMInsightsDataCollector.rb
@@ -6,12 +6,13 @@ module VMInsights
     require_relative 'VMInsightsIDataCollector.rb'
 
     class DataCollector < IDataCollector
-        def initialize(root_directory_name="/")
+        def initialize(log, root_directory_name="/")
+            @log = log
             @root = root_directory_name
             @baseline_exception = RuntimeError.new "baseline has not been called"
             @cpu_count = nil
             @saved_net_data = nil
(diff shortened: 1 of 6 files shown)


source/code/plugins/VMInsightsDataCollector.rb
```diff
@@ -99,6 +100,7 @@ def get_filesystems
                             result << Fs.new(a[0], a[6], a[2], a[4]) if a.size == 7
                         rescue ArgumentError => ex
                             # malformed input
+                            @log.debug() { "#{__method__}: #{ex}" }
```
aydany: Do you want to log the line as well?

source/code/plugins/in_vminsights.rb
```diff
@@ -21,29 +21,30 @@ def configure(conf)
         super
 
         begin
+            @log = LogWrapper.new(LogPrefix, @log)# unless @log.kind_of? LockWrapper
```
aydany: Do you want to keep the commented out part?

source/code/plugins/VMInsightsDataCollector.rb
mthompson83: good catch

source/code/plugins/in_vminsights.rb
mthompson83: Ooops!

PR #1098, opened 2019-11-12
Added -e tag  in configure_logrotate for LAD

nidhanda:
Adding an -e tag in logrotate to have better/safe  file formatting


hestolz:
broke out into separate file in #1128, don't need this any longer

PR #1101, opened 2019-11-13
don't use 'service' to check the existance of cylance

abenbachir:


installer/bundle/bundle_skel.sh
```diff
@@ -244,13 +244,9 @@ is_suse11_platform_with_openssl1(){
 }
 
 detect_cylance(){
+  # Don't use 'service' to check the existance of a service
+  # because it will fail if there is no service available
```
nidhanda: Will removing the service check and just checking for the files be sufficient? Are we sure we won't miss any scenario?

abenbachir: nop

PR #1104, opened 2019-11-14
Update docs for 1 12 15 release

abenbachir:


hestolz:
@abenbachir we should just remove the year-month in the readme header. it serves no function. if people want to see how new the readme is, they can just see the last commit date

abenbachir:
> @abenbachir we should just remove the year-month in the readme header. it serves no function. if people want to see how new the readme is, they can just see the last commit date

We can do it next time

Issue #1111, opened 2019-11-21
Invalid logrotate file for LAD


taoyama:
I can see error while logrotate.
`# /usr/sbin/logrotate -d /etc/logrotate.conf
...
reading config file omsagent-LAD
error: omsagent-LAD:1 lines must begin with a keyword or a filename (possibly in double quotes)
...`

It seems that the logrotate file is invalid.
`# cat /etc/logrotate.d/omsagent-LAD
/var/opt/microsoft/omsagent/LAD/log/omsagent.log {\n\trotate 10\n\tmissingok\n\tnotifempty\n\tcompress\n\tsize 100M\n\tcopytruncate\n}`

hestolz:
@taoyama is there an ICM/CRI associated with this? Does this repro on multiple distros or just some?
@nidhanda please take a look at #1112 as you are the original author of this code

sbonds:
I noticed this same issue and it has been reported as part of a larger Microsoft support request. I'm hesitant to post case numbers publicly, but I referenced this issue within the case.

hestolz:
Fixed in #1128, release ETA is February.

PR #1112, opened 2019-11-21
Fix invalid logrotate generation

taoyama:
#1111

`-e` option is required for `echo`

```
$ echo "/var/opt/microsoft/omsagent/LAD/log/omsagent.log {\n\trotate 10\n\tmissingok\n\tnotifempty\n\tcompress\n\tsize 100M\n\tcopytruncate\n}"
/var/opt/microsoft/omsagent/LAD/log/omsagent.log {\n\trotate 10\n\tmissingok\n\tnotifempty\n\tcompress\n\tsize 100M\n\tcopytruncate\n}
```

```
$ echo -e "/var/opt/microsoft/omsagent/LAD/log/omsagent.log {\n\trotate 10\n\tmissingok\n\tnotifempty\n\tcompress\n\tsize 100M\n\tcopytruncate\n}"
/var/opt/microsoft/omsagent/LAD/log/omsagent.log {
        rotate 10
        missingok
        notifempty
        compress
        size 100M
        copytruncate
}
```

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -1141,7 +1141,7 @@ configure_logrotate()
 
     # create the logrotate file for omsagent if workspace is LAD
     if [ "$WORKSPACE_ID" = "LAD" ]; then
-        echo "/var/opt/microsoft/omsagent/LAD/log/omsagent.log {\n\trotate 10\n\tmissingok\n\tnotifempty\n\tcompress\n\tsize 100M\n\tcopytruncate\n}" > /etc/logrotate.d/omsagent-$WORKSPACE_ID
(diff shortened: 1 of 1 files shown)


nidhanda:
We have discussed fixing this with the team in our last hot-fix but the general consensus was that this would be replaced by using a separate file for LAD and copying it over instead of using ‘echo’ to create the file. That fix will be part of the next release. 

Is there a particular distro that this logrotate file without formatting doesn’t work? Logrotate seems to be able to still work without the correct formatting

taoyama:
There are some query incoming relating this issue.
I agree with your strategy so please close this request once the proper fix is merged.
Thanks.

nidhanda:
Taoyama - is there any specific distro where the current file is failing for?

taoyama:
This is not happening on specific distro.
I have done some research and founded out that the environment which installed both LinuxDiagnostic and OMSAgentForLinux Extension has this issue.

This is because LinuxDiagnostic zip bundles omsagent-1.11.0-9.x86_64 and OMSAgentForLinux zip bundles omsagent-1.12.15-0.x86_6. And the problematic shell script is only in 1.12-15.

I hope this is helpful for you.

hestolz:
Fixed #1128

PR #1125, opened 2020-01-14
Added support for HybridRP IMDS Endpoint 

nidhanda:
oms_configuration.rb now checks if hybrid agent is installed on a VM and if it is, then it grabs the IMDS endpoint for the HybridRP from the config file located on the VM and queries that to get the Azure Resource ID. 

Testing Strategy - (Checked whether the following work as expected and report errors correctly)
1. Install the oms build into a VM without Hybrid agent (No Resource id found, heartbeat doesnt have all VM data)
2. Install Hybrid agent and restart and reinstall omsagent (HyrbidRP IMDS Resource id queried and used successfully)
3. Install oms build on Azure VMs (Regular IMDS resource id created and used successfully)

### DIFF
--- a/source/code/plugins/oms_configuration.rb
+++ b/source/code/plugins/oms_configuration.rb
@@ -189,6 +189,10 @@ module OMS
             azure_resource_id = imds_instance_json_compute['resourceID']
             
             return azure_resource_id
+          else
+            #config file containing HyrbidRP IMDS endpoint not found. 
+            raise "The config file #{@@HybridRPConfFile} containing the HybridRP IMDS Endpoint is not found."
+          end
         rescue
           # this may be a container instance or a non Hybrid Agent VM
           OMS::Log.warn_once("Could not fetch Azure Resource ID from IMDS, Reason: #{e}")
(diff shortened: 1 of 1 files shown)


source/code/plugins/oms_configuration.rb
```diff
@@ -167,9 +213,20 @@ def update_azure_resource_id()
             break if retries > max_retries
             azure_resource_id = get_azure_resid_from_imds()
```
hestolz: Might be cleaner if your first action after the `break if` is to check whether or not you are in the hybrid env by checking `azcmagent show`, and then based on that you can make the call to either `get_azure_resid_from_imds` or `get_azure_resid_from_hybridagent`

That way you can avoid the sleep/increment retries code.

nidhanda: Sure. Good idea. I'll add that change and update the PR

source/code/plugins/oms_configuration.rb
```diff
@@ -159,13 +162,62 @@ def get_azure_resid_from_imds()
...
+            return nil if !imds_instance_json.has_key?("compute") || imds_instance_json['compute'].empty? #classic vm
+            imds_instance_json_compute = imds_instance_json['compute']
+            
+            return nil if !imds_instance_json_compute.has_key?("resourceID") || imds_instance_json_compute['resourceID'].empty? 
+            azure_resource_id = imds_instance_json_compute['resourceID']
+            
+            return azure_resource_id
+          else
+            #config file containing HyrbidRP IMDS endpoint not found. 
+		        OMS::Log.warn_once("Could not find the config file #{@@HybridRPConfFile} containing the HybridRP IMDS Endpoint. Failed to fetch azure resource id.")
```
hestolz: nit indent

PR #1128, opened 2020-01-18
Added a separate logrotate config for LAD 

nidhanda:
Created a separate logrotate config for lad - lad-logrotate.conf and copied that over to sysconf and then further into /etc/logrotate/omsagent-LAD during onboarding for LAD instead of echoing raw text into a config file for lad during that process.

Testing Strategy - 
1) Built and installed this shell bundle on a VM and then onboarded LAD and verified that the config file was correctly copied over
2) Manually ran logrotate command with debug enabled on the VM and verified the correctness of its output with this new lad-logrotate.conf file


installer/scripts/omsadmin.sh
```diff
@@ -1141,7 +1141,10 @@ configure_logrotate()
 
     # create the logrotate file for omsagent if workspace is LAD
     if [ "$WORKSPACE_ID" = "LAD" ]; then
-        echo "/var/opt/microsoft/omsagent/LAD/log/omsagent.log {\n\trotate 10\n\tmissingok\n\tnotifempty\n\tcompress\n\tsize 100M\n\tcopytruncate\n}" > /etc/logrotate.d/omsagent-$WORKSPACE_ID
+        # create the logrotate file for the LAD if it doesn't exist
```
hestolz: nit indent

installer/scripts/omsadmin.sh
```diff
@@ -1141,7 +1141,10 @@ configure_logrotate()
...
+    	if [ ! -f /etc/logrotate.d/omsagent-$WORKSPACE_ID ]; then
+            cp $SYSCONF_DIR/lad-logrotate.conf  /etc/logrotate.d/omsagent-$WORKSPACE_ID
```
hestolz: since this is `cp` and below is `cat ... >` (not append), is this correct? In the LAD + OMS side-by-side case, wouldn't we want both the logrotate.conf AND lad-logrotate.conf to be installed?

hestolz: also I assume yes, but just to sanity check—omsadmin.sh will be called for LAD onboard regardless of whether omsagent package is already installed (since in side-by-side install LAD will skip omsagent package if installed second)

nidhanda: When LAD is installed, it calls the omsadmin.sh to onboard to a workspace named LAD so yes this method will be called. The catch is that the cx should have this omsadmin.sh file there which can be through two ways - 
1) The cx has the latest omsagent - 1.13.* and then installs LAD, in which case LAD simply uses this current omsadmin.sh and sets up logrotate.
2) The cx has the latest LAD release which contains 1.13.* oms . This will be done in a future LAD release once we know OMS 1.13.* is stable.

nidhanda: To answer your first question. We do cat in case of OMS because we want to use sed to update the workspace id in the logrotate conf. In case of LAD we know that the workspace will be called LAD so we can simply write that in the default config and copy it over.

installer/scripts/omsadmin.sh
nidhanda: The comment indent was matched according to the if statement below. I think that is the usual practice right?

PR #1129, opened 2020-01-23
Fixing LAD & OMS conflicts during OMS uninstall

abenbachir:


installer/bundle/bundle_skel.sh
```diff
@@ -941,13 +941,21 @@ if [ "$installMode" = "R" -o "$installMode" = "P" ]; then
...
+        azsec_mdsd_installed=$?
+        check_if_program_exists_on_system lad-mdsd
+        lad_mdsd_installed=$?
+
+        # if at least one of mdsd product is installed
+        MDSD_INSTALLED=$(( $azsec_mdsd_installed && $lad_mdsd_installed ))
+
+        # If LAD was installed don't remove omsagent, but proceed if purge mode was selected
+        if [ $lad_mdsd_installed -ne 0 -o "$installMode" = "P"]; then
+            pkg_rm omsagent
```
nidhanda: Why are we checking if installMode is Purge? In the case where LAD is installed on the VM and CX is running installMode as purge through shell bundle, This if statement will go and try to purge oms and still fail due to lad oms dependency issue right?

abenbachir: That how it was done in the next line, I didn't want to change the behavior.

abenbachir: Also our LAD/OMS extensions use "--remove", so it safe.

nidhanda: But then in case where LAD is installed and then we uninstall LAD and it calls/tries to remove OMS. Wouldn't that just skip this pkg_rm omsagent since lad_mdsd_installed might return True (depending on at what step Lad uninstall process OMS removal is called) and not remove OMS even though we want to completely remove LAD and all it's dependencies?

abenbachir: LAD will remove lad-mdsd first then calls omsagent-*.sh --remove next.
T

installer/bundle/bundle_skel.sh
```diff
@@ -941,13 +941,31 @@ if [ "$installMode" = "R" -o "$installMode" = "P" ]; then
...
 
         pkg_rm omsconfig
-        pkg_rm omsagent
 
-        # If MDSD is installed and we're just removing (not purging), leave SCX
-        MDSD_INSTALLED=1
-        check_if_program_exists_on_system azsec-mdsd
-        if [ $? -eq 0 -o -d /var/lib/waagent/Microsoft.OSTCExtensions.LinuxDiagnostic-*/mdsd ]; then
-            MDSD_INSTALLED=0
+        # If MDSD/LAD is installed and we're just removing (not purging), leave OMS, SCX and OMI
```
canfikret: Pls fix the comment wrt purging. It should not be uninstalled even when purging.

installer/scripts/uninstall.sh
```diff
@@ -67,15 +67,33 @@ check_if_pkg_is_installed() {
 # Do our stuff - uninstall the OMS agent
 
 pkg_rm omsconfig
-pkg_rm omsagent
 
-# If MDSD is installed and we're just removing (not purging), leave SCX
-
-MDSD_INSTALLED=1
+# If MDSD/LAD is installed and we're just removing (not purging), leave OMS, SCX and OMI
```
canfikret: why is this code duplicated?

abenbachir: I don't know, it was like that. @simathih probably knows the rason behind that.

installer/bundle/bundle_skel.sh
abenbachir: We will handle purging on a different PR, let focus here on the remove.

installer/bundle/bundle_skel.sh
```diff
@@ -941,13 +941,31 @@ if [ "$installMode" = "R" -o "$installMode" = "P" ]; then
...
+        check_if_pkg_is_installed azsec-mdsd
+        azsec_mdsd_installed=$?
+        check_if_pkg_is_installed lad-mdsd
+        lad_mdsd_installed=$?
+
+        # if at least one of mdsd product is installed
+        MDSD_INSTALLED=$(( $azsec_mdsd_installed && $lad_mdsd_installed ))
```
simathih: I have seen the case where only one of them is installed and it could still break OMS uninstall

simathih: comment says we are checking for "At least one," but we are doing a &&

installer/bundle/bundle_skel.sh
```diff
@@ -941,13 +941,31 @@ if [ "$installMode" = "R" -o "$installMode" = "P" ]; then
...
+
+        # If LAD was installed don't remove omsagent, but proceed if purge mode was selected
+        if [ $lad_mdsd_installed -ne 0 -o "$installMode" = "P" ]; then
+            pkg_rm omsagent
```
simathih: actually won't apt-get not allow omsagent to be removed since omsagent mdsd uses omi too?

simathih: are you sure this will succeed as omsagent has dependency on omi?

installer/bundle/bundle_skel.sh
```diff
@@ -941,13 +941,31 @@ if [ "$installMode" = "R" -o "$installMode" = "P" ]; then
...
+                primary_ws_id=`grep WORKSPACE_ID ${ws_conf_dir}/omsadmin.conf | cut -d= -f2`
+            fi
+
+            if [ "${primary_ws_id}" != "" -a "${primary_ws_id}" != "LAD" ]; then
+                echo "--- Unboarding the workspace ${primary_ws_id}... ---"
+                /opt/microsoft/omsagent/bin/omsadmin.sh -x ${primary_ws_id}
+            fi
         fi
 
         if [ $MDSD_INSTALLED -ne 0 -o "$installMode" = "P" ]; then
```
simathih: whats the difference between this and the check above? both are looking for LAD MDSD

installer/scripts/uninstall.sh
simathih: could be one is called from shell bundle and another from omsadmin.sh....I am not 100% sure though

installer/scripts/uninstall.sh
```diff
@@ -67,15 +67,33 @@ check_if_pkg_is_installed() {
...
+        primary_ws_id=`grep WORKSPACE_ID ${ws_conf_dir}/omsadmin.conf | cut -d= -f2`
+    fi
+
+    if [ "${primary_ws_id}" != "" -a "${primary_ws_id}" != "LAD" ]; then
+        echo "--- Unboarding the workspace ${primary_ws_id}... ---"
+        /opt/microsoft/omsagent/bin/omsadmin.sh -x ${primary_ws_id}
+    fi
 fi
+
 if [ $MDSD_INSTALLED -ne 0 -o "$installMode" = "P" ]; then
```
simathih: whats the difference between this and the check above? both are looking for LAD MDSD

installer/bundle/bundle_skel.sh
abenbachir: because the value "0" means installed, so I have to use XOR or AND.

installer/bundle/bundle_skel.sh
abenbachir: The old code was removing OMS all the time, that condition will skip it if LAD was installed.
LAD has 3 dependencies: omsagent, scx and omi.

installer/bundle/bundle_skel.sh
abenbachir: That condition is to make sure we are not unboarding LAD workspace, in order to make the code robust.

installer/scripts/uninstall.sh
abenbachir: the code in uninstall.sh is a duplicate of the code in bundle_skel.sh

installer/scripts/uninstall.sh
abenbachir: nop, I just figured out that the duplication is important if the "uninstall.sh" is not there, like in old agents.

installer/bundle/bundle_skel.sh
canfikret: ok

PR #1135, opened 2020-02-04
Detect VM UUID in Oracle 6 from dmi

canfikret:
VMUUID is showing empty GUID in omsadmin.conf in previous gen distros

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -547,6 +547,8 @@ onboard()
         case "$ASSET_TAG" in
             77*) OMSCLOUD_ID=$ASSET_TAG ;;       # If the asset tag begins with a 77 this is the azure guid
         esac
+        PRODUCT_UUID=$(cat /sys/devices/virtual/dmi/id/product_uuid)
+        UUID=$(echo $PRODUCT_UUID | tr '[:upper:]' '[:lower:]')
     fi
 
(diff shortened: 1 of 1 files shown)


nidhanda:
Not sure if this is related, but when i first started, I fixed a bug with this PR #913  where if the UUID read from the omsadmin.conf is not in a valid UUID format, then I returned an empty UUID (0000-0000-....) instead of returning the invalid UUID as that was causing an issue in the portal when it displayed the UUID in logs.

hestolz:
@nidhanda was it just UUID=<empty> or was it doubled/some other invalid value
In any case seems like before UUID wasn't even set at all on this code path so I don't think there should be any negative side effects

canfikret:
> Not sure if this is related, but when i first started, I fixed a bug with this PR #913 where if the UUID read from the omsadmin.conf is not in a valid UUID format, then I returned an empty UUID (0000-0000-....) instead of returning the invalid UUID as that was causing an issue in the portal when it displayed the UUID in logs.

No, this orthogonal.  Your change is a fallback to GUID.Zero if the it is empty or invalid.  This PR reads it from sys.

canfikret:
> @nidhanda was it just UUID= or was it doubled/some other invalid value
> In any case seems like before UUID wasn't even set at all on this code path so I don't think there should be any negative side effects

It was UUID=

PR #1136, opened 2020-02-04
fluentd filter_parser security patch to avoid escape sequence injection

aswatt:


aswatt:
Fixes escape sequence injection open vulnerability in fluentd

source/ext/patches/fluentd/filter_parser.patch
```diff
@@ -0,0 +1,11 @@
+--- ../source/ext/fluentd/lib/fluent/plugin/filter_parser.rb	2019-09-09 11:51:09.485143653 -0700
++++ ../source/ext/fluentd/lib/fluent/plugin/filter_parser.rb.new	2020-02-04 14:40:45.848862819 -0800
+@@ -53,7 +53,7 @@
+             if @emit_invalid_record_to_error
+               router.emit_error_event(tag, time, record, ::Fluent::ParserError.new("pattern not match with data '#{raw_value}'"))
+             else
+-              log.warn "pattern not match with data '#{raw_value}'" unless @suppress_parse_error_log
++		    log.warn "pattern not match with data '#{raw_value.dump}'" unless @suppress_parse_error_log
```
hestolz: nit somehow the indent here is weird, you generated with diff though right? 
also is there any way we can test this

hestolz:
reference of change https://github.com/fluent/fluentd/pull/1733/files

source/ext/patches/fluentd/filter_parser.patch
abenbachir: To test this we have to enale fluentd unit tests.

aswatt: Huh that's weird, all I did to edit it was just add '.dump' to the line in .new and generate it with diff, but yeah I'll fix it fast.

PR #1137, opened 2020-02-05
Use resolv-replace module for DNS lookups

AaronRM:
Use Resolv module instead of Socket::getaddrinfo for DNS lookups

### DIFF
--- a/source/code/plugins/oms_common.rb
+++ b/source/code/plugins/oms_common.rb
@@ -17,6 +17,7 @@ module OMS
     require 'digest'
     require 'date'
     require 'securerandom'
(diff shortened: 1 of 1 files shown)


source/code/plugins/oms_common.rb
```diff
@@ -17,6 +17,7 @@ class Common
     require 'digest'
     require 'date'
     require 'securerandom'
+    require 'resolv-replace'
```
hestolz: we don't need any additional gem for this ?

AaronRM: No, it's in the Ruby stdlib.

PR #1138, opened 2020-02-06
Ods rate limiting

abenbachir:
Send more headers to ODS like: User-Agent, retry-count.
Set the correct exponential back off configuration to match requirements: retry timeout of 30 seconds up to 6 attempts during ~30 mins.

### DIFF
--- a/installer/conf/omsagent.conf
+++ b/installer/conf/omsagent.conf
@@ -59,7 +59,7 @@
   flush_interval 60s
   retry_limit 10
   retry_wait 30s
-  max_retry_wait 9m
+  max_retry_wait 30m
 </match>
 
 <match oms.** docker.**>
@@ -81,7 +81,7 @@
   flush_interval 20s
   retry_limit 10
   retry_wait 30s
-  max_retry_wait 9m
+  max_retry_wait 30m
 </match>
 
 <match diag.oms diag.oms.**>
@@ -103,7 +103,7 @@
   flush_interval 10s
   retry_limit 10
   retry_wait 30s
-  max_retry_wait 9m
+  max_retry_wait 30m
 </match>
 
 # Catch all unprocessed data and output it
--- a/installer/conf/omsagent.conf
+++ b/installer/conf/omsagent.conf
@@ -57,7 +57,7 @@
   buffer_queue_limit 10
   buffer_queue_full_action drop_oldest_chunk
   flush_interval 60s
-  retry_limit 10
+  retry_limit 6
   retry_wait 30s
   max_retry_wait 30m
 </match>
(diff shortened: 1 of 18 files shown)


source/code/plugins/oms_common.rb
```diff
@@ -732,6 +732,11 @@ def create_ods_request(path, record, compress, extra_headers=nil, serializer=met
           headers["Content-Encoding"] = "deflate"
         end
 
+        headers["User-Agent"] = "LinuxMonitoringAgent/#{OMS::Common.get_agent_version}"
+        headers[OMS::CaseSensitiveString.new("x-ms-app")] = "LinuxMonitoringAgent"
+        headers[OMS::CaseSensitiveString.new("x-ms-client-version")] = OMS::Common.get_agent_version
```
hestolz: Could get_agent_version call fail/return empty

installer/conf/omsagent.conf
```diff
@@ -57,9 +57,9 @@
   buffer_queue_limit 10
   buffer_queue_full_action drop_oldest_chunk
   flush_interval 60s
-  retry_limit 10
+  retry_limit 6
```
simathih: does this do exponential backoff?

source/code/plugins/oms_common.rb
abenbachir: Then we have to fix the get_agent_version itself.

installer/conf/omsagent.conf
abenbachir: exponential backoff is implemented in fluentd by default.

source/code/plugins/oms_common.rb
hestolz: In some case it could be nil, at least should do a check like preexisting headers here against that to avoid sending a nil field

abenbachir: nil field will be converted to empty string and the results will be User-Agent="LinuxMonitoringAgent/"

I guess that is fine, or I can make it as User-Agent="LinuxMonitoringAgent/0.0.0-0"

hestolz: I think the latter is more useful / sensible.

source/code/plugins/oms_common.rb
```diff
@@ -652,7 +652,8 @@ def get_agent_version(conf_path = "/etc/opt/microsoft/omsagent/sysconf/installin
             @@AgentVersion = agent_version
           end
         end
-        return @@AgentVersion
+
+        return @@AgentVersion.nil? ? '0.0.0-0': @@AgentVersion
```
hestolz: with this approach we hit the file read every time get_agent_version is called assuming version is still not detectable, but in any case agent is a bad state. looks fine

PR #1140, opened 2020-02-18
Cleanup run_in_background configuration and use ENV instead

abenbachir:


source/code/plugins/out_oms.rb
```diff
@@ -29,6 +29,11 @@ def initialize
     def configure(conf)
       s = conf.add_element("secondary")
       s["type"] = ChunkErrorHandler::SecondaryName
+
+      # override default run_in_background config
+      if !OMS::BackgroundJobs::enabled.nil?
+        conf['run_in_background'] = OMS::BackgroundJobs::enabled
```
hestolz: you can reassign conf dict? why not just reassign `@run_in_background`

abenbachir:
> How many customers are currently relying on this?
> How are you going to avoid breaking this feature for them?

And no one is using it for now. 
Nothing will break, I'm introducing a new ENV variable o handle turning on/off that feature.

source/code/plugins/out_oms.rb
abenbachir: If you assign @run_in_background fluentd will reset that value from "conf['run_in_background']".
Look at other fluentd internal plugins, they all use this technique to override configs.

hestolz: yeah after i commented i checked, whoops

PR #1154, opened 2020-02-28
debian10 container: add '-s' arg during kill term

abenbachir:
In Debian10:

If you run kill without '-s' arg you get this error:
```
kill -sigterm
-bash: kill: igterm: invalid signal specification
```

```
 kill -sigkill
-bash: kill: igkill: invalid signal specification
```


abenbachir:
> it seems it should be `sigkill` not `-sigkill`

done

PR #1160, opened 2020-03-11
Generate DSC Metaconfig

hestolz:
CRI 177130188
ADO 6369194

Need to regenerate DSC metaconfig when we get DSC endpoint in topology request response. Testing methodology:
- emplace modified `agent_maintenance_script.rb`
- `/opt/microsoft/omsagent/bin/service_control restart`
- modify `DSC_ENDPOINT` in `omsadmin.conf` with fake value
- `su - omsagent -c /opt/microsoft/omsconfig/Scripts/OMS_MetaConfigHelper.py`
- check that `/etc/opt/omi/conf/omsconfig/generated_meta_config.mof` has new fake value
- wait for topology request, check `generated_meta_config.mof` again for replaced, corrected value

### DIFF
--- a/source/code/plugins/agent_maintenance_script.rb
+++ b/source/code/plugins/agent_maintenance_script.rb
@@ -1,4 +1,5 @@
 require 'optparse'
+require 'open3'
 
 module MaintenanceModule
 
@@ -190,7 +191,42 @@ module MaintenanceModule
       # When apply_dsc_endpoint is called from onboarding, dsc_endpoint will be returned in file
       update_config("DSC_ENDPOINT", dsc_endpoint)
 
+      generate_dsc_metaconfig
+
       return dsc_endpoint
+    end
+
+    # Generate the DSC MetaConfig file. If the dsc endpoint is updated, DSC will
+    # not directly pick the endpoint from omsadmin.conf (instead, it is picked
+    # from MetaConfig.mof), so we must manually trigger metaconfig (re)generation.
+    def generate_dsc_metaconfig
+      # OMS_MetaConfigHelper.py will log to omsconfig.log by default,
+      command = "/opt/microsoft/omsconfig/Scripts/OMS_MetaConfigHelper.py"
+      timeout = 10
+      begin
+        Open3.popen2(command, :pgroup=>true) do |_, _, wait_thr|
+          pid = wait_thr.pid
+          pgid = Process.getpgid(pid)
+          deadline = Time.now + timeout
+          sleep 0.5 until Time.now > deadline or !wait_thr.status
+          if wait_thr.status # still hasn't completed
+            begin
+              `pkill -TERM -g #{pgid}`
+              sleep 1
+              if wait_thr.status
+                `pkill -KILL -g #{pgid}`
+              end
+            rescue => e
+              log_error("Failed to kill process(es) for command '#{command}'. #{e}")
+            end
+            OMS::Log.warn_once("DSC MetaConfig generation failed due to subprocess timeout")
+          elsif wait_thr.value.exitstatus != 0 # nonzero exit
+            OMS::Log.warn_once("DSC MetaConfig generation failed with nonzero exit code #{wait_thr.value.exitstatus}, check omsconfig.log for details")
+          end
+        end
+      rescue => e
+        OMS::Log.warn_once("DSC MetaConfig generation failed due to subprocess call failure: #{e}")
+      end
     end
 
     # Pass the server response from an XML file to apply_dsc_endpoint and apply_certificate_update_endpoint
--- a/source/code/plugins/agent_maintenance_script.rb
+++ b/source/code/plugins/agent_maintenance_script.rb
@@ -123,6 +123,7 @@ module MaintenanceModule
     # Update omsadmin.conf with the specified variable's value
     def update_config(var, val)
       if !File.exist?(@omsadmin_conf_path)
+        log_error("Missing configuration file: #{@omsadmin_conf_path}")
         return OMS::MISSING_CONFIG_FILE
       end
 
@@ -132,6 +133,10 @@ module MaintenanceModule
       File.open(@omsadmin_conf_path, "w") { |file|
         file.puts(new_text)
       }
+
+      if old_text != new_text
+        log_info("omsadmin.conf updated with #{var}=#{val}")
+      end
     end
 
     # Updates the CERTIFICATE_UPDATE_ENDPOINT variable and renews certificate if requested
@@ -233,6 +238,7 @@ module MaintenanceModule
     # Save DSC_ENDPOINT and CERTIFICATE_UPDATE_ENDPOINT variables in file to be read outside of this script
     def apply_endpoints_file(xml_file, output_file)
       if !OMS::Common.file_exists_nonempty(xml_file)
+        log_error("Missing configuration file: #{xml_file}")
         return OMS::MISSING_CONFIG_FILE
       end
 
--- a/source/code/plugins/agent_maintenance_script.rb
+++ b/source/code/plugins/agent_maintenance_script.rb
@@ -42,6 +42,10 @@ module MaintenanceModule
       @LOG_FACILITY = nil
       @CERTIFICATE_UPDATE_ENDPOINT = nil
 
+      @current_DSC_ENDPOINT = nil
+      @pending_DSC_ENDPOINT = nil
+      @pending_metaconfig_generation = false
+
       @load_config_return_code = load_config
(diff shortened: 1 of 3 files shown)


Bhargava-Chary-Chollaty:
Can we consider adding log when metaconfig.mof is re-generated and what was the change in endpoint. Old and Updated endpoint.

source/code/plugins/agent_maintenance_script.rb
```diff
@@ -190,9 +191,44 @@ def apply_dsc_endpoint(server_resp)
...
 
+      generate_dsc_metaconfig
+
       return dsc_endpoint
     end
 
+    # Generate the DSC MetaConfig file. If the dsc endpoint is updated, DSC will
+    # not directly pick the endpoint from omsadmin.conf (instead, it is picked
+    # from MetaConfig.mof), so we must manually trigger metaconfig (re)generation.
+    def generate_dsc_metaconfig
```
rishabh1jain: Umm will it more efficient to check if it has changed, rather than killing and restarting the process each time a topology response is received i.e. every hour?

hestolz: the process will not be killed

rishabh1jain: Oh I see, you are running the command and monitoring that. my bad.

hestolz: No worries—basically we are running a .py script that will write to `generated_meta_config.mof`. Neither DSC nor OMS needs to be restarted for the change to that file to take effect. Since we run this as a subprocess, we need to make sure it doesn't  hang or turn zombie. For instance, if someone else is has already taken the DSC lock, this call could fail or take some time. In that case we shouldn't wait forever. Next time topology request is made we will try again.

hestolz:
> Can we consider adding log when metaconfig.mof is re-generated and what was the change in endpoint. Old and Updated endpoint.

I added a log statement if any value changes in `omsadmin.conf`. In all cases `omsconfig.log` will show the result when we do these metaconfig generation calls.

chsamala (commented):
- Because  of locking if you failed to apply new metaconfiguration , then please retry rather than waiting for next topology request.

- if there is no change in topology configuration(DSC part) then dont apply the meta configuration.

source/code/plugins/agent_maintenance_script.rb
```diff
@@ -190,13 +196,49 @@ def apply_dsc_endpoint(server_resp)
...
+            begin
+              `pkill -TERM -g #{pgid}`
+              sleep 1
+              if wait_thr.status
+                `pkill -KILL -g #{pgid}`
+              end
+            rescue => e
+              log_error("Failed to kill process(es) for command '#{command}'. #{e}")
+            end
+            OMS::Log.warn_once("DSC MetaConfig generation failed due to subprocess timeout")
```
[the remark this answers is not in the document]
hestolz: I have tested this and it works. I already detailed how I tested this in the PR description.

If you check the OMS_MetaConfigHelper.py it doesn't need any parameters if it is being run to generate metaconfig. The parameters are only necessary if you are trying to disable or enable DSC config management.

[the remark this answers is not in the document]
Bhargava-Chary-Chollaty: OMS_MetaConfigHelper.py persists the argument/flag that it was executed with previously by creating file (/etc/opt/omi/conf/omsconfig/omshelper_disable).

Assume a user runs OMS_MetaConfigHelper.py with --disable flag. This would change the Metaconfig from PULL mode to PUSH mode. If the user runs the script second time with out any flag, then the script will still use previous argument and configures Metaconfig in PUSH mode where DSC endpoint is not configured.

We need to test the changes considering scenario where the DSC endpoint change is happened when the omsagent is in disables state and is enabled later by the user.

source/code/plugins/agent_maintenance_script.rb
Bhargava-Chary-Chollaty: I think the mof file to be checked if the change is successfully applied is at `/etc/opt/omi/conf/omsconfig/configuration/MetaConfig.mof`

`generated_meta_config.mof` is an intermediate file generated which is sent to dsc while applying meta config.

source/code/plugins/agent_maintenance_script.rb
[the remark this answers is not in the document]
hestolz: I've added the check to see if DSC is disabled. If so, I won't call `OMS_MetaConfigHelper.py` but will persist a flag so once DSC is re-enabled we will call.

hestolz:
Addressed Feedback
- Retry up to 3 times
- Only generate the metaconfig when there is a change
- Add logging indicating when there is a change, and if so log the old/new values
- Handle possible DSC disabled case

source/code/plugins/agent_maintenance_script.rb
```diff
@@ -171,32 +183,105 @@ def apply_certificate_update_endpoint(server_resp, check_for_renew_request = tru
...
+        end
+      end
+
+      OMS::Log.warn_once("Exceeded max attempts to generate DSC metaconfig, will retry after next topology request")
+    end
+
+    # Call OMS_MetaConfigHelper.py to trigger metaconfig generation, return success
+    def call_oms_metaconfighelper
+      # Note that OMS_MetaConfigHelper.py will log to omsconfig.log
+      command = "/opt/microsoft/omsconfig/Scripts/OMS_MetaConfigHelper.py"
```
hestolz, line 246: this will have to change once [this](https://github.com/microsoft/PowerShell-DSC-for-Linux/pull/691/files) is taken

source/code/plugins/agent_maintenance_script.rb
```diff
@@ -171,32 +188,105 @@ def apply_certificate_update_endpoint(server_resp, check_for_renew_request = tru
...
+        if success
+          @pending_metaconfig_generation = false
+          return
+        else
+          sleep 5
+          retries += 1
+        end
+      end
+
+      log_warn("Exceeded max attempts to generate DSC metaconfig, will retry after next topology request")
```
Bhargava-Chary-Chollaty, line 245: If all the retries are exhausted, we are saying that the dsc endpoint is updated in next topology request. 

In the first request, we call the update_config with new endpoint which i assume will update the omsadmin.conf. If all the retries fail we never revert this change, how will we able to identify the change in endpoint in this scenario for next topology requests ?

source/code/plugins/oms_configuration.rb
```diff
@@ -365,6 +365,15 @@ def set_request_intervals(topology_interval, telemetry_interval)
         OMS::Log.info_once("OMS agent management service telemetry request interval now #{@@TelemetryInterval}")
       end
 
+      # Check whether DSC has been disabled (which prevents automatic config management)
+      def is_dsc_disabled
+        begin
+          return File.exist?("/etc/opt/omi/conf/omsconfig/omshelper_disable")
+        rescue => e
+          OMS::Log.error_once("Failed to get DSC enabled/disabled status. #{e}")
```
Bhargava-Chary-Chollaty, line 373: We do not return a boolean to caller on exception. How would this effect the flow ?

source/code/plugins/agent_maintenance_script.rb
hestolz: Think of it in terms of invariants

1. omsadmin.conf will always be updated with the newest DSC_ENDPOINT following every topology request
2. pending_metaconfig_generation will always be set when a change is detected *and never unset until the metaconfig is successfully generated*.

Since the flag is never unset there is no reason to revert the change to omsadmin.conf; when the next topology request comes around metaconfig generation will be retried.

Bhargava-Chary-Chollaty: Would it mean the state of variable 'pending_metaconfig_generation' is always persisted. Even for different executions of the script. I am assuming that this is not executed by a continuous process and there is an initialization of the variable to false in the start. Sorry if it is obvious, I did not get how it works in rb.

hestolz: No problem. It is persisted because this file defines a class `Maintenance` which is instantiated in `in_heartbeat_request.rb`, which is an input plugin that starts up and runs for the entire life cycle of the agent. You can see in that file we call `@maintenance_script.heartbeat` (do a topology request) and `@maintenance_script.generate_dsc_metaconfig` (call this method).

Bhargava-Chary-Chollaty: Wouldn't machine restart effect this ?

hestolz:

PR #1167, opened 2020-03-27
Added check to prevent onboarding to a workspace if omsagent is stopped

nidhanda:
Currently if the omsagent is not running, then omsadmin.s -w <ws-id> -s <ws-key> command will allow it to re-onboard to the same workspace. This is because we only currently only check if the agent is running or stopped and if it is stopped we go forward with the onboarding. To prevent this, I added checks to make sure that if a workspace is already onboarded, regardless of the status of omsagent, we do not onboard, but rather ask the cx to un-onboard first.

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -460,11 +460,7 @@ onboard()
     # If a test is not in progress then call service_control to check on the workspace status
     if [ -z "$TEST_WORKSPACE_ID" -a -z "$TEST_SHARED_KEY" ]; then
         $SERVICE_CONTROL is-running $WORKSPACE_ID > /dev/null 2>&1
-        #The service_control is-running method will always return 1 for workspaces that are not currently onboarded because the check method exits with error code 1 in that script
-        #But since we check for exit code 1 as success, it assumes the agent is running on that workspace and prevents onboarding. This can be fixed by using exit code 0 for 
-        #success, as is convention in Linux. But since Multihoming isn't currently supported, this actually acts as a prevention check for multihoming. After discussion, leaving it
-        #as it is for now. Can fix this in future when we want to support multihoming. 
-        if [ $? -eq 1 ]; then
+        if [ $? -eq 0 ]; then
             echo "Workspace $WORKSPACE_ID already onboarded and agent is running."
             if [ -z "$MULTI_HOMING_MARKER" -a ! -h $DF_CONF_DIR ]; then
                 echo "Symbolic links have not been created; re-onboarding to create them"
@@ -791,9 +787,9 @@ show_workspace_status()
     local is_primary=$3
     local status='Unknown'
 
-    # 1 if omsagent-ws_id is running, 0 otherwise
+    # 0 if omsagent-ws_id is running, 1 otherwise
     $SERVICE_CONTROL is-running $ws_id
-    if [ $? -eq 1 ]; then
+    if [ $? -eq 0 ]; then
         status='Onboarded(OMSAgent Running)'
     elif [ -f ${ws_conf_dir}/.service_registered ]; then
         status='Warning(OMSAgent Registered, Not Running)'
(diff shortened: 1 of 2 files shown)


installer/scripts/omsadmin.sh
```diff
@@ -469,6 +469,39 @@ onboard()
...
+        # no default conf folder, check all the potential workspace folders
+        for ws_id in `ls -1 $ETC_DIR | grep -E '^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$'`
+        do
+            found_ws=1
+        done
+    fi
+
+    if [ $found_ws -ne 0 ]; then
+        echo "Already Onboarded to a workspace, please un-onboard first before trying to onboard to a new workspace"
+        return 0
```
simathih, line 502: why 0? it should return a code that is sent back to the python code and ultimately should show up in the VM extensoin tab in portal.

hestolz: I think this case would only arise if the customer had installed bundle, stopped the agent, and then tried to enable the extension. But I agree, probably should not return 0

nidhanda: Yeah this case only happens when the agent is installed successfully and then is stopped manually and then cx tries to re-onboard. The only caveat is that if the agent is installed as a VM extension then Waagent will repeatedly call enable on it until it succeeds, so i guess yeah it is a good idea to use an error code instead of 0. Should I use 53 (The extension failed due to missing or wrong configuration parameters) instead?

hestolz: of the 50s, 53 is probably best but IDK if it is exactly right. regardless, it should fall into one of the codes we ignore for SLA purposes

nidhanda: Hmm thinking again on this, shouldnt it still be 0? This case isnt really an enable failure. It is simply preventing re-onboaring step but the enable process will still continue after this and it will restart the stopped omsagent in case the agent was onboarded before but is manually stopped. Which is what we expect from the Enable command. If we return 53 then the Enable will fail. Not sure which should be the right behavior here. Any ideas @simathih @hestolz ?

hestolz: What if the cx is trying to onboard a different workspace. Will that be caught by either the bundle onboard script or the extension? I know we have stop on multiple connections for the extension now, but the same logic is not part of bundle

nidhanda: It will first be caught by the extension code, and then in the bundle itself as well. Currently the Service Control is_running method returns a non 0 exit code for a diff workspace. This wasn't by design but is rather a happy co-incidence (issue) as it prevents multihoming by default. I had added a small change to fix that issue as part of my multihoming work but since that is on hold for now, I decided to leave the service_control script as it is to avoid unnecessary MH problems.

simathih: if this returns 0, then extension will assume install was successful and skip?

nidhanda: The issue is unlikely to happen in case of install as the agent will not be onboarded to another workspace before and it won't have the workspace-id directories. But yeah i agree that there can be edge cases or cases with incomplete installation or uninstallation that can hit this issue. So, after discussion with Henry, **I changed the exit code to 53** as that seemed to be the best option.

installer/scripts/omsadmin.sh
```diff
@@ -469,6 +469,39 @@ onboard()
             fi
         fi
     fi
+
+    #check for workspaces alredy onboarded
```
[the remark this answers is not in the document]
nidhanda: Fixed in the new commit.

nidhanda:
Fixed the service_control is-running bug as well where it was always returning 1 for new ws-id's . My new check prevents mulithoming regardless so there was no point in keeping the service_control issue unsolved. @abenbachir  @hestolz  can you please take a look again. Thanks

abenbachir:
looks good.

PR #1171, opened 2020-04-12
Keep Cisco event Id

Ronmarsiano:
During the extraction of the identifier from Cisco ASA we removed the event ID.

In this PR we keep the event ID in the Identifier instead of returning just "%ASA"


Ronmarsiano:
@abenbachir  @mgladi  could one of you merge it ? I don't have write access

mgladi:
@Ronmarsiano I also don't have write access. @abenbachir can you help?

PR #1172, opened 2020-04-13
Bump auoms version to 2.1.1-0

taglines:


NarineM:
No longer needed. Version is already newer.

PR #1176, opened 2020-04-21
Changing event ID parsing for Cisco

Ronmarsiano:
Changing the parsing of event ID for Cisco ASA 
Encountered a bug where the event ID was dropped since we used to return the "%ASA" string only


ofshomro:
@abenbachir,
Hi, my name is Ofir SHomron.
I'm the new CEF owner in Azure Sentinel.
Do you have any update on this, we have a lot of customers that want this.

PR #1191, opened 2020-05-08
Fix syslog-ng check

abenbachir:


nidhanda:
Have you tested this on a LAD repro? Can we check if this works with a manual  ./diagnostic.py -enable command for LAD just to be sure?

abenbachir:
> Have you tested this on a LAD repro? Can we check if this works with a manual ./diagnostic.py -enable command for LAD just to be sure?

I removed that SLES11 hack in LAD here https://github.com/Azure/azure-linux-extensions/pull/1104

test/code/plugins/oms_configuration_test.rb
```diff
@@ -151,6 +151,7 @@ def test_load_configuration()
       assert_equal(TEST_UUID, Configuration.uuid, "VM UUID should be loaded")
       assert_not_equal(nil, Configuration.cert, "Certificate should be loaded")
       assert_not_equal(nil, Configuration.key, "Key should be loaded")
+      p "get_azure_region_from_imds=", OMS::Configuration.get_azure_region_from_imds()
```
nidhanda, line 154: Why are we doing a print here instead of assert check?

abenbachir: testing to fix the UT

Issue #1199, opened 2020-05-29
Agent fails to install on RPM based systems with dpkg installed


roberthawdon:
An edge case we have discovered is when an RPM based system has `dpkg` installed (e.g. a CentOS machine used for building which uses Alien to convert deb packages to rpms). The Oms installation script detects `dpkg` and tries to use that to install the agent.

Ideally, the installation script should run `lsb_release -si` and compare the output to a list of known distributions and their default package managers, then fall back to detecting `dpkg` or `rpm` if the lookup fails.

PR #1200, opened 2020-05-29
Improve package manager detection

roberthawdon:
Should address issue #1199 where dpkg might be chosen on rpm systems if
the dpkg application is present.

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -293,21 +293,31 @@ ulinux_detect_installer()
 {
     INSTALLER=
 
-    # If DPKG lives here, assume we use that. Otherwise we use RPM.
-    check_if_program_in_path dpkg
-    if [ $? -eq 0 ]; then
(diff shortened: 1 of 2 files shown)


hestolz:
Hi @roberthawdon, thanks for the contribution. This needs to work for all supported distros (CentOS 6/7 Oracle 6/7 RHEL 6/7/8 Debian 8/9 Ubuntu 14/16/18 SLES 12/15), have you verified against these?

roberthawdon:
It's been tested against CentOS 7 and Ubuntu 16. Do you have any suggestion for automating the test across the supported distros?

The test will fall back to blindly looking for `dpkg` and `rpm` if the checking of the `*-release` files doesn't work.

hestolz:
Hi @roberthawdon, thanks for the help but I will go a more explicit way of package manager detection. The approach in this PR works on Ubuntu but not, for instance, on Debian. You can see the approach in #1205.

PR #1204, opened 2020-06-09
Python Aliasing and 2/3

hestolz:
Make alterations to the small areas in the shell bundle as well as the log collector that need to check for python. 
- Check for python during bundle and docker provider install, just like we check for the `python-ctypes` package
- Ensure that python scripts are called with `python2` or `python3` depending on which is available, rather than the unversioned `python` command
- Made some changes to the log collector e.g. parenthesized `print()` statements, briefly tested with python3 and it seems to work fine. Is not a critical path as we can always have customers run it with python2 or upload a new version quickly

Not in scope:
- Some scripts remain untouched e.g. one from HDInsight and another from Automation Worker. I have not altered these; in the former case, the author is no longer at MSFT and in the latter, from documentation it appears that Automation Worker does not support python 3 workbooks in any case.


hestolz:
Once I get this approved, I'll create a log collector bundle download and upload that before merging.

installer/bundle/bundle_skel.sh
```diff
@@ -1066,6 +1098,20 @@ if [ "$installMode" = "I" -o "$installMode" = "U" ]; then
     compare_install_type
     compare_arch
 
+    python_installed
+    if [ $? -ne 0 ]; then
+        if [ -z "${forceFlag}" ]; then
+            echo "Error: Python is not installed on this system, installation cannot continue." >&2
+            echo "Please install either the python2 or python3 package." >&2
+            echo "You can run this shell bundle with --force; in this case, we will install omsagent," >&2
+            echo "but omsconfig (DSC configuration) will not be available and will need to be re-installed." >&2
```
nidhanda, line 1107: Will the omsagent work correctly (albeit without the configs) without DSC installed and won't crash because of some missing file that dsc needs to create that we might overlook??

hestolz: look below
this is duplicate behavior of what we do if python-ctypes is not detected

nidhanda: Oh nvm. I didn't see the code below.

hestolz:
log collector download added

PR #1212, opened 2020-06-15
Fix Restart Failure

hestolz:
Often we see in CRIs the message `restart_omsagent failed with result '1' on workspace ...`, this change seeks to fix that. Often with those CRIs if I had a call I was able to fix the state by running `systemctl restart omsagent-<wsid>`. 

In the old function, if omsagent _wasn't_ running then restart would call `start_omsagent`. However, I got access to a repro machine and noticed if I just changed `start_omsagent` to call `systemctl restart` it was fixed. Rather than changing that function with unexpected consequences, let's just make `restart_omsagent` work as expected and use the logic it already has to call `systemctl restart`.

### DIFF
--- a/installer/scripts/service_control
+++ b/installer/scripts/service_control
@@ -470,11 +470,9 @@ stop_omsagent()
 
 restart_omsagent()
 {
+    enable_omsagent_service
     remove_PIDfile_unless_omsagent_running
-    if this_omsagent_running; then
-        enable_omsagent_service
-    else
-        start_omsagent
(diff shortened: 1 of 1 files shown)


installer/scripts/service_control
```diff
@@ -470,11 +470,9 @@ stop_omsagent()
 
 restart_omsagent()
 {
+    enable_omsagent_service
```
hestolz, line 473: added here because in any case, with previous setup it would either be called if `this_omsagent_running` returned true, or in `start_omsagent` when it is called.

PR #1221, opened 2020-07-14
XFS filesystem support for VMInsights

pfrcks:


test/code/plugins/VMInsightsDataCollector_subtest.rb
```diff
@@ -1169,13 +1174,15 @@ def initialize(s)
 
         def make_expected_disks(n)
             seed = Random.rand(26)
+            filesystems = ["ext2", "ext3", "ext4", "xfs"]
```
mthompson83, line 1177: Looking at how we randomized the file system types before, there was a 30% chance that 4 random selections might omit one of the options. Now, with 4 distinct types, there is now a 42% chance that one of the types may not be represented.

Recommendation: use filesystems[i % filesystems.length] to select the type.

test/code/plugins/VMInsightsDataCollector_subtest.rb
```diff
@@ -1127,27 +1130,29 @@ def initialize(dev, mp, size, free, type=(Random.rand(3) + 2))
...
-                @free == actual.free_space_in_bytes
+                @free == actual.free_space_in_bytes &&
+                @type == actual.filesystem
             end
 
             def <=>(o)
                 r = @dev <=> o.dev; return r unless r.zero?
                 r = @mp <=> o.mp; return r unless r.zero?
                 r = @size <=> o.size; return r unless r.zero?
                 @free <=> o.dev
```
mthompson83, line 1148: What the??? This line, `@free <=> o.dev` is wrong! free should be compared to free, not dev!

test/code/plugins/VMInsightsDataCollector_subtest.rb
```diff
@@ -1127,27 +1130,29 @@ def initialize(dev, mp, size, free, type=(Random.rand(3) + 2))
...
+                r = @type <=> o.type; return r unless r.zero?
```
mthompson83, line 1149: Logic error.

test/code/plugins/VMInsightsDataCollector_subtest.rb
```diff
@@ -315,16 +315,19 @@ def test_get_filesystems_garbage
...
             ].concat(expected)
             # these look funky, but they should be interpreted as decimal numbers
             stuff << DfGarbage.new("/dev/bar1 ext4 042 42 17 3 /shazam")
-            expected << Fs.new("/dev/bar1", "/shazam", 42, 17, 4)
+            expected << Fs.new("/dev/bar1", "/shazam", 42, 17, "ext4")
             stuff << DfGarbage.new("/dev/bar2 ext4 42 42 017 2 /shazam")
-            expected << Fs.new("/dev/bar2", "/shazam", 42, 17, 4)
+            expected << Fs.new("/dev/bar2", "/shazam", 42, 17, "ext4")
+            stuff << DfGarbage.new("/dev/bar3 xfs 42 42 017 2 /shazam")
+            expected << Fs.new("/dev/bar3", "/shazam", 42, 17, "xfs")
```
mthompson83, line 330: Our original test case is broken. We should not be creating an invalid scenario of two devices on the same mount point.

test/code/plugins/VMInsightsDataCollector_subtest.rb
```diff
@@ -1127,27 +1130,29 @@ def initialize(dev, mp, size, free, type=(Random.rand(3) + 2))
...
             def equivilent?(actual)
                 @dev == '/dev/' + actual.device_name &&
                 @mp == actual.mount_point &&
                 @size == actual.size_in_bytes &&
-                @free == actual.free_space_in_bytes
+                @free == actual.free_space_in_bytes &&
+                @type == actual.filesystem
             end
 
             def <=>(o)
```
mthompson83, line 1144: Given the existing bug noted below and the new logic error, I have to wonder if this routine is actually used.

source/code/plugins/VMInsightsDataCollector.rb
```diff
@@ -349,7 +350,7 @@ def get_up_net_devices
         end
 
         class Fs
-            def initialize(device_name, mount_point, size_in_bytes, free_space_in_bytes)
+            def initialize(device_name, mount_point, size_in_bytes, free_space_in_bytes, filesystem)
```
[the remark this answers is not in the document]
pfrcks: Will rename to filesystem_format

test/code/plugins/VMInsightsDataCollector_subtest.rb
```diff
@@ -1127,7 +1130,7 @@ def initialize(dev, mp, size, free, type=(Random.rand(3) + 2))
             end
 
             def to_s
-                "%-10s %-15s %15d %15d" % [ @dev, @mp, @size, @free ]
+                "%-10s %-15s %15d %15d %-10s" % [ @dev, @mp, @size, @free, @type ]
```
mthompson83, line 1133: Do not add type to `to_s`, `eqivilent?` or as a readable attribute.

test/code/plugins/VMInsightsDataCollector_subtest.rb
```diff
@@ -1141,13 +1144,14 @@ def <=>(o)
                 r = @dev <=> o.dev; return r unless r.zero?
                 r = @mp <=> o.mp; return r unless r.zero?
                 r = @size <=> o.size; return r unless r.zero?
-                @free <=> o.dev
+                r = @free <=> o.free; return r unless r.zero?
+                @type <=> o.type
```
mthompson83, line 1148: I'm confused how this could work. Isn't `o` always supposed to be an object created by the production code? If so, it isn't defined to have a `type` attribute.

pfrcks: this o is not from production code. see how o contains dev, mp, size, type which are attributes of the Fs class defined in test code

PR #1224, opened 2020-07-24
Azure Arc fixes

nidhanda:
Ref workitem - https://dev.azure.com/msazure/One/_workitems/edit/7786346
Ref icm - https://icm.ad.msft.net/imp/v3/incidents/details/194830574/home

Changes - 
1) The ruby system command to check the existence of Arc's azcmagent gives the wrong error code in CentOS/RHEL. Using systemctl himds status instead of azcmagent show
2) The ruby module resolv-replace has an issue with resolving localhost in urls. Which caused all the http request to localhost fail. Replaced localhost with resolved 127.0.0.1 to fix this issue.

### DIFF
--- a/source/code/plugins/oms_configuration.rb
+++ b/source/code/plugins/oms_configuration.rb
@@ -173,8 +173,11 @@ module OMS
               OMS::Log.warn_once("Failed to read the IMDS Endpoint from #{@@HybridRPConfFile}. Failed to fetch azure resource id")
               return nil
             end
-            azureHybridIMDSEndpoint = azureHybridIMDSEndpoint.split("=")[1][0..-2]
+
+            #Subbing localhost with 127.0.0.1 because Ruby's resolv-replace module (imported in oms_common.rb fails to correctly resolve localhost)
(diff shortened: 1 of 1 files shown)


source/code/plugins/oms_configuration.rb
```diff
@@ -173,8 +173,11 @@ def get_azure_resid_from_hybridagent()
               OMS::Log.warn_once("Failed to read the IMDS Endpoint from #{@@HybridRPConfFile}. Failed to fetch azure resource id")
               return nil
             end
-            azureHybridIMDSEndpoint = azureHybridIMDSEndpoint.split("=")[1][0..-2]
+
+            #Subbing localhost with 127.0.0.1 because Ruby's resolv-replace module (imported in oms_common.rb fails to correctly resolve localhost)
+            azureHybridIMDSEndpoint = azureHybridIMDSEndpoint.split("=")[1][0..-2].sub("localhost","127.0.0.1")
```
hestolz, line 178: why do you need `[0..-2]`

nidhanda: It is to remove the '**\"**' at the end of the url. the scan() function above adds the extra slash in between the url and the end quotes.

PR #1227, opened 2020-07-29
updated the sudoers.d file to enable python3 support for automationworker

krmanupa:


installer/conf/sudoers
```diff
@@ -53,6 +53,8 @@ Defaults:nxautomation lecture = never
 nxautomation ALL=(ALL) NOPASSWD: ALL
 omsagent ALL=(ALL) NOPASSWD: /usr/bin/python /opt/microsoft/omsconfig/modules/nxOMSAutomationWorker/DSCResources/MSFT_nxOMSAutomationWorkerResource/automationworker/worker/main.py *
```
hestolz, line 54: these should be to /usr/bin/python2
for instance on RHEL 8 which comes with no python, if a customer then installs just python2, there will be no alias and `/usr/bin/python` will not exist.

PR #1234, opened 2020-08-12
add network connectivity info to curl error

aswatt:
When curl errors out and omsadmin.sh returns 8, error could be due to either incorrect workspace key or an issue connecting to the `<wsid>.oms.opinsights.azure.com` URL. Adding this info to output.

### DIFF
--- a/installer/scripts/omsadmin.sh
+++ b/installer/scripts/omsadmin.sh
@@ -576,6 +576,7 @@ onboard()
     if [ $CURL_VERSION_SYSTEM -gt $CURL_VERSION_WITH_DEFAULT_HTTP2 ]; then
       CURL_HTTP_COMMAND="--http1.1"
     fi
+    CURL_URL=`https://${WORKSPACE_ID}.oms.${URL_TLD}/AgentService.svc/LinuxAgentTopologyRequest`
 
     RET_CODE=`curl --header "x-ms-Date: $REQ_DATE" \
         --header "x-ms-version: August, 2014" \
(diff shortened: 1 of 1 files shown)


aswatt:
Joint with this PR in azure-linux-extensions: https://github.com/Azure/azure-linux-extensions/pull/1187

installer/scripts/omsadmin.sh
```diff
@@ -604,7 +605,7 @@ onboard()
         cleanup_certs
         return $INVALID_PROXY
     elif [ $error -ne 0 ]; then
-        log_error "Error during the onboarding request: curl returned $error. Check the correctness of the workspace ID and shared key or run omsadmin.sh with '-v'"
+        log_error "Error during the onboarding request: curl returned $error when trying to connect to $CURL_URL . Check the correctness of the workspace ID and shared key, as well as the connectivity to the url, or run omsadmin.sh with '-v' to get verbose output"
```
hestolz, line 608: - is there a command we can specify to run here to check connectivity to the URL? for instance `telnet ${WORKSPACE_ID}.oms.${URL_TLD} 443`? admittedly I am not positive what the most robust command to run would be, but you should be able to test out by restricting the NSG and/or installing a firewall (have seen this commonly ... would be good to account for that not just NSG) and then seeing what commands are a reliable indicator of connectivity.
- nit no space after `$CURL_URL`

installer/scripts/omsadmin.sh
```diff
@@ -576,6 +576,7 @@ onboard()
     if [ $CURL_VERSION_SYSTEM -gt $CURL_VERSION_WITH_DEFAULT_HTTP2 ]; then
       CURL_HTTP_COMMAND="--http1.1"
     fi
+    CURL_URL=`https://${WORKSPACE_ID}.oms.${URL_TLD}/AgentService.svc/LinuxAgentTopologyRequest`
```
hestolz, line 579: I don't think you want this in backticks, should be quotes right?

hestolz:

PR #1236, opened 2020-08-13
Fix version  to find url of newer version

akenamon:
Avoid 404 error


hestolz:
@akenamon thanks for the contribution—I have opened a PR myself just because there are a couple more places that need to be changed ( #1237 )

akenamon:
> @akenamon thanks for the contribution—I have opened a PR myself just because there are a couple more places that need to be changed ( #1237 )

thanks @hestolz  !

PR #1239, opened 2020-08-25
Update NPM ruby plugin files to latest that are available in Powershell-DSC-for-Linux

vakaranamsft:
This PR is to fetch:

All latest commits that has happened to Powershell-DSC-for-Linux w.r.t nxOMSAgentNPMConfig DSC module's plugin files.


NarineM:
No longer applicable

PR #1242, opened 2020-09-01
fix empty Computer field problem when sending CEF logs

NoamLandress:


abenbachir (changes requested):
Please rebase and make sure build is green

Issue #1238, opened 2020-08-13
The Computer field does not appear on the logs in Azure Log Analitycs


buckmuu:
There is a BUG in the file /opt/microsoft/omsagent/plugin/filter_syslog_security.rb that does not populate the Host field of the record.

buckmuu:
Please include the two circled lines to solve the problem.

![fix](https://user-images.githubusercontent.com/49308192/90166317-52148500-dd9a-11ea-9769-35161f4c15c9.png)

buckmuu:
Another way to fix the same issue fixed in te PR #1244

hestolz:
Thank you for your submission. As [previously announced](https://azure.microsoft.com/en-us/updates/were-retiring-the-log-analytics-agent-in-azure-monitor-on-31-august-2024/), the Log Analytics agent has been **deprecatated** and has no support as of **August 31, 2024.** If you use the Log Analytics agent to ingest data to Azure Monitor, [migrate now to the new Azure Monitor agent](https://docs.microsoft.com/en-us/azure/azure-monitor/agents/azure-monitor-agent-migration). As part of repo archival, open issues and pull requests will be closed.

PR #1244, opened 2020-09-06
Fix Empty Computer field issue

NoamLandress:
The Computer field is derived from the Host field which isn't configured in filter_syslog_security.rb
This commit fixes this issue.


buckmuu:
We have the same issue. We have suggest a similar solution at #1238 
Both fix the issue, please implement one of them.
Thanks

Lukeout:
@abenbachir  @hestolz @NoamLandress  can we merge this PR in?

hestolz (approved):

PR #1255, opened 2020-10-09
Create purge script

aswatt:
Based off of this purge documentation: https://supportability.visualstudio.com/AzureLogAnalytics/_wiki/wikis/Azure-Log-Analytics.wiki/281924/How-to-Purge-and-reinstall-the-Linux-agent

Tested on multiple distros (both shell bundle and extension, and with LAD installed as well). Standalone script from the agent designed to purge OMS and all related information from the machine, for a fresh reinstall.


tools/purge_omsagent.sh
```diff
@@ -0,0 +1,407 @@
...
+}
+
+ulinux_detect_servicecontroller()
+{
+    CONTROLLER=""
+
+    # Detect based on systemd
+    if which systemctl > /dev/null 2>&1; then
+        CONTROLLER="systemctl"
+
```
hestolz, line 67: nit no newline to make unified if-else block

tools/purge_omsagent.sh
```diff
@@ -0,0 +1,407 @@
...
+{
+    WAAGENT_NAME=""
+
+    if [ "$CONTROLLER" = "systemctl" ]; then
+        if systemctl status waagent > /dev/null 2>&1; then
+            WAAGENT_NAME="waagent"
+        elif systemctl status walinuxagent > /dev/null 2>&1; then
+            WAAGENT_NAME="walinuxagent"
+        fi
+    elif [ "$CONTROLLER" = "service" ]; then
```
hestolz, line 138: nit
can make second 2 cases into 1 else and use `$CONTROLLER waagent status`
same for other functions

tools/purge_omsagent.sh
```diff
@@ -0,0 +1,407 @@
...
+        else
+            echo "Unknown input. Please type 'y' (yes) or 'n' (no) to proceed."
+        fi
+    done
+
+    # continuing with LAD purge
+    echo "Please go to portal.azure.com and uninstall the LAD extension."
+    echo "Go to Azure Portal -> Virtual Machines -> <vm_name> -> Settings -> Extensions"
+    echo "and then click the '...' in the LinuxDiagnostic row and click 'Uninstall'"
+    read -p "Press enter to proceed once the Extension is uninstalled." toss2
```
hestolz, line 272: what is `toss2` used for here?

tools/purge_omsagent.sh
```diff
@@ -0,0 +1,407 @@
...
+    fi
+    if [ -d $OMS_LAD_DIR ]; then
+        rm -rf $OMS_LAD_DIR
+    fi
+
+    # remove LAD package
+    echo "---------- Removing LAD package ----------"
+    pkg_rm lad-mdsd
+
+    # remove MDSD package
```
hestolz, line 288: how come this is necessary; I would not expect any mdsd package but `lad-mdsd`

tools/purge_omsagent.sh
```diff
@@ -0,0 +1,407 @@
...
+echo "--------------------------------------------------------------------------------"
+echo "------------------------------ STARTING OMS PURGE ------------------------------"
+echo "--------------------------------------------------------------------------------"
+echo ""
+
+# Check if installed via extension
+ls /var/lib/waagent | grep -i Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux- > /dev/null 2>&1
+EXTENSION_INSTALLED=$?
+if [ $EXTENSION_INSTALLED -eq 0 ]; then
+    echo ""
```
hestolz, line 310: If the customer has a policy or ASC set up to push OMS agent to the machine, that will end up pretty quickly re-installing the OMS extension; maybe that should be called out here. It might not be a big concern as long as they come back to the purge script and quickly resume, but that is not a given ...

tools/purge_omsagent.sh
```diff
@@ -0,0 +1,407 @@
+#! /bin/bash
```
hestolz, line 1: 1. should check that we are running as root and warn+exit if not
2. what's your plan for shipping this file, will this be bundled a-la the TST? customers will have to DL from GitHub?

tools/purge_omsagent.sh
```diff
@@ -0,0 +1,407 @@
...
+        echo "  https://raw.githubusercontent.com/Microsoft/OMS-Agent-for-Linux/master/installer/scripts/onboard_agent.sh" >&2
+        echo "Or ensure the onboard_agent.sh script is available in the same directory as this script." >&2
+        call_exit $ONBOARD_AGENT_MISSING
+    fi
+fi
+echo ""
+
+echo "---------- Running OMS Purge Script ----------"
+chmod +x ./onboard_agent.sh
+./onboard_agent.sh --purge
```
hestolz, line 335: this will download the entire bundle again, right? is there some way to not download the entire bundle (will [this](https://github.com/microsoft/OMS-Agent-for-Linux/blob/master/installer/scripts/uninstall.sh) suffice? IDK if this is equal/superset of purge function from `bundle_skel`)

if we have to download the bundle, can you make sure it is deleted afterwards? its 100MB+

tools/purge_omsagent.sh
```diff
@@ -0,0 +1,407 @@
...
+if [ -f $WAAGENT_XML ]; then
+    STOPPED_WAAGENT=1
+    echo "Temporarily stopping waagent..."
+    stop_waagent
+
+    echo "Removing OMS Extension files..."
+    rm -rf $WAAGENT_XML
+fi
+
+# remove directories (if they exist)
```
hestolz, line 348: are there any omi-related directories?

tools/purge_omsagent.sh
```diff
@@ -0,0 +1,407 @@
...
+dir_rm /var/opt/microsoft/dependency-agent dependency-agent
+dir_rm /var/opt/microsoft/omsagent omsagent
+dir_rm /var/opt/microsoft/omsconfig omsconfig
+dir_rm /var/opt/microsoft/scx scx
+if [ ! "$(ls -A /var/opt/microsoft > /dev/null 2>&1)" ]; then
+    dir_rm /var/opt/microsoft
+fi
+
+if [ $EXTENSION_INSTALLED -eq 0 ]; then
+    dir_rm /var/lib/waagent/Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux-*
```
hestolz, line 377: 1. doesn't `dir_rm` take 2 parameters? 
2. not an expert on wildcard expansion but pretty sure this will expand to all possible extension versions _before_ dir_rm is executed, which I would expect to then give an error from dir_rm since param 2 would not be a package name. we want this to remove all possible versions, though, so either dir_rm needs to be changed to take package name first and then >= 1 dirpath to remove, or we need a loop here to iterate over the wildcard expansion

hestolz (changes requested):

tools/purge_omsagent.sh
```diff
@@ -0,0 +1,407 @@
+#! /bin/bash
+
+SOLO_LAD_DIR=/var/lib/waagent/Microsoft.Azure.Diagnostics.LinuxDiagnostic-*
+OMS_LAD_DIR=/var/opt/microsoft/omsagent/LAD
+
+OMS_ONBOARD_AGENT=onboard_agent.sh
```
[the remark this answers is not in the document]
aswatt: I wanted to make it so if there was firewall issues, getting the onboard_agent.sh script not via a `wget` and then placing it in the same folder as the purge script still worked
(Although I see an issue now, since the onboard_agent.sh script uses `wget` anyway, so it would just prolong the inevitable, I'll change it)

tools/purge_omsagent.sh
aswatt: It's not used for anything, but `read` kept erroring out if I didn't put a variable to catch the input so I just made a variable to toss

tools/purge_omsagent.sh
aswatt: It was just in case of some weird installation issues where MDSD got installed under a different name or similar

tools/purge_omsagent.sh
aswatt: 1. Yes sorry I was going to add that but forgot
2. The second, I was just going to have it as a standalone that customers can download directly from GitHub

tools/purge_omsagent.sh
aswatt: Actually I think the `uninstall.sh` script you linked would be perfect, I didn't see it before that would work a lot better, thank you!

tools/purge_omsagent.sh
aswatt: Yes sorry need to add those, they weren't in the same `opt/microsoft/` location

tools/purge_omsagent.sh
aswatt: 1. It's a weird workaround but since the second parameter is empty it gets passed to `pkg_rm`, which has a check to make sure that the input isn't empty (or if it is empty, it does nothing)
2. I think a loop would probably work best, I'll make that change

tools/purge_omsagent.sh
aswatt: Fixed in this commit: https://github.com/microsoft/OMS-Agent-for-Linux/pull/1255/commits/6084ab38cd50be252158f02bd813a2b094ffb573

tools/purge_omsagent.sh
aswatt: Fixed in this commit: https://github.com/microsoft/OMS-Agent-for-Linux/pull/1255/commits/5849d518e8127291f13356832c8c7a023e24041e

tools/purge_omsagent.sh
aswatt: Fixed in this commit: https://github.com/microsoft/OMS-Agent-for-Linux/pull/1255/commits/e7138a75afb34623b64d2811b51f366a35ca4b57

tools/purge_omsagent.sh
aswatt: Fixed in this commit: https://github.com/microsoft/OMS-Agent-for-Linux/pull/1255/commits/386c30f11d39fafcd8398703d5b2377252afac50

tools/purge_omsagent.sh
aswatt: Both fixed in this commit: https://github.com/microsoft/OMS-Agent-for-Linux/pull/1255/commits/1b4155724692379c7c06374f903f0109755cd418

tools/purge_omsagent.sh
aswatt: 1. Fixed in this commit: https://github.com/microsoft/OMS-Agent-for-Linux/pull/1255/commits/fb08aac1fb438c916c00116b69ab79411553a011

tools/purge_omsagent.sh
aswatt: Fixed in this commit: https://github.com/microsoft/OMS-Agent-for-Linux/pull/1255/commits/6084ab38cd50be252158f02bd813a2b094ffb573

hestolz (approved):

PR #1269, opened 2020-11-18
Create Hardening.md

aswatt:
Create customer-facing documentation on our hardening support based on [this internal documentation](https://supportability.visualstudio.com/AzureLogAnalytics/_wiki/wikis/Azure-Log-Analytics.wiki/306340/Required-permissions-for-OS-Hardening-Customizations-Linux-Agent).


docs/Hardening.md
```diff
@@ -0,0 +1,197 @@
+# Linux Hardening while using OMS Log Collector
```
hestolz, line 1: think the "while ..." can be omitted

docs/Hardening.md
```diff
@@ -0,0 +1,197 @@
...
+We support some industry standard hardening (please see [Supported Hardening or Customization](#supported-hardening-or-customization) for more details).
+We also can provide some [troubleshooting steps](#common-hardening-issues) for customized hardening.
```
hestolz, line 3: Should we specify our support boundary here? e.g. does a customer who has done custom hardening who has followed this TSG and still been unable to get OMS to work qualify for support or not? Or are we saying FIPS and Cylance are OK, but the rest of this is unsupported (but here's some tips anyways)? @Lukeout

docs/Hardening.md
```diff
@@ -0,0 +1,197 @@
...
+- SELinux
+- Center for Internet Security (CIS)
+
+## Common Hardening Issues
+If you're using a customized hardening setup, and are having issues running the OMS Agent, try going through these common problems encountered with hardening.
+### Directory Permissions
+#### Issue
+Several major folders in Linux need to have appropriate file permissions, otherwise the OMS Agent won't run properly. Specifically, we need the following:
+| Directory | Permissions (numeric value) | Permissions (symbolic notation) |
+| --- | --- | --- |
```
hestolz, line 35: how about `/opt/`?

docs/Hardening.md
```diff
@@ -0,0 +1,197 @@
...
+| `/var` | 755 | `drwxr-xr-x` |
+| `/var/log` | 775 | `drwxrwxr-x` |
+| `/etc` | 775 | `drwxrwxr-x` |
+| `/tmp` | 775 | `drwxrwxr-x` |
+
+#### How to Check
+You can check what the permissions for these folders are by the following:
+```
+$> ls -l / | grep var
```
hestolz, line 44: is there a way that we could do this in a single command to simplify this for customers? e.g. (untested) `grep -E '/var|/var/log|...'`

docs/Hardening.md
```diff
@@ -0,0 +1,197 @@
...
+
+#### How to Check
+You can check the version of Ruby that's running with the following commands:
+```
+$> ruby --version
+ruby 2.2.2p95 (2015-04-13 revision 50295) [x86_64-linux]
+$> which ruby
+/usr/local/rvm/rubies/ruby-2.2.2/bin/ruby
+```
+Standard Ruby install through OMS Agent should result in below:
```
hestolz, line 101: This confused me personally; my understanding was that most machines will have _system_ Ruby installed at `/usr/bin/ruby`, while we (OMS) install our own ruby at `/opt/microsoft/omsagent/ruby/bin/ruby`. This section seems to imply that we install the OMS-shipped ruby at `/usr/bin/ruby`, which is not the case. 

Assuming my understanding is correct (@abenbachir ?) maybe the line can be changed to:
"The standard Ruby install should result in the below:" 

And the "Issue" description might want to change to:
"OMS Agent ships a specific version of Ruby. Installing a different version of Ruby from the RVM (Ruby Version Manager) may define global envvar (e.g. `GEM_HOME`, `GEM_PATH`) that conflict with those used by the OMS-specific Ruby installation."

hestolz (changes requested):

docs/Hardening.md
abenbachir: @hestolz maybe I haven't explain it more clear, but the Official Ruby shipped by any linux distribution is not the issue here, but rather the custom Ruby installation that is made through RVM. Feel free to improve the wording.

abenbachir (approved):
Looks good for me.

hestolz:
Will this be promoted to Azure Docs or linked there?

docs/Hardening.md
aswatt: I was trying to imply the latter with this, but I don't really know of a good way to word it that doesn't seem blunt

docs/Hardening.md
aswatt: Abder didn't mention it when listing out which directories are required, but that seems important - @abenbachir are the permissions required the same as the majority of the others (775)?

docs/Hardening.md
aswatt: I couldn't get it in one command (since /var/log isn't at the same level as the rest), but I can combine two commands in one line to get everything, do you think the below is okay?
```
$> ls -l / | grep -E 'var|etc|opt|tmp'; ls -ld /var/log
drwxr-xr-x.  85 root root 8192 Nov 13 00:55 etc
drwxr-xr-x.   6 root root   55 Nov 10 18:15 opt
drwxrwxrwt.   7 root root   93 Nov 18 03:13 tmp
drwxr-xr-x.  20 root root  282 Nov 13 00:54 var
drwxr-xr-x. 10 root root 4096 Nov 15 03:36 /var/log
```

docs/Hardening.md
hestolz: Blunt is fine ... the document should not leave room for interpretation

aswatt:
> Will this be promoted to Azure Docs or linked there?

@Lukeout what do you think? I'm not sure, but if we add it to Azure Docs we can just add the info to the hardening section that's already there, or at least link it

docs/Hardening.md
aswatt: Fixed in this commit:
https://github.com/microsoft/OMS-Agent-for-Linux/pull/1269/commits/b6a96c6bef0fd670a3c6676c17026d5228135af5#diff-ff8800189743d6567bcda8a83de5b22c943b28c51e4488cce0ea2b185716cc37R1

docs/Hardening.md
aswatt: Changed here to be clearer:
https://github.com/microsoft/OMS-Agent-for-Linux/pull/1269/commits/b6a96c6bef0fd670a3c6676c17026d5228135af5#diff-ff8800189743d6567bcda8a83de5b22c943b28c51e4488cce0ea2b185716cc37R2

docs/Hardening.md
aswatt: Added `/opt` with 775 just in case:
https://github.com/microsoft/OMS-Agent-for-Linux/pull/1269/commits/b6a96c6bef0fd670a3c6676c17026d5228135af5#diff-ff8800189743d6567bcda8a83de5b22c943b28c51e4488cce0ea2b185716cc37R38

docs/Hardening.md
aswatt: Changed to one line command:
https://github.com/microsoft/OMS-Agent-for-Linux/pull/1269/commits/b6a96c6bef0fd670a3c6676c17026d5228135af5#diff-ff8800189743d6567bcda8a83de5b22c943b28c51e4488cce0ea2b185716cc37R44

docs/Hardening.md
aswatt: Clarified Ruby installation wording:
https://github.com/microsoft/OMS-Agent-for-Linux/pull/1269/commits/b6a96c6bef0fd670a3c6676c17026d5228135af5#diff-ff8800189743d6567bcda8a83de5b22c943b28c51e4488cce0ea2b185716cc37R90

docs/Hardening.md
hestolz: I wasn't clear—why is this scoped to log collector? Isn't this hardening for OMS agent itself?

aswatt: ...I'm an idiot, yes it is, sorry, I'll change it to OMS Agent

aswatt: Fixed in this commit: https://github.com/microsoft/OMS-Agent-for-Linux/pull/1269/commits/3d7de51694edac07a8c333f03a33464faa4347fc#
I wanted to keep the "OMS Agent" specification just to make sure it was clear that this has to do with hardening compatibility with OMS (if the fact that it was in our repo wasn't clear enough)

hestolz (approved):

PR #1278, opened 2020-12-04
Add purge script to auto-install

aswatt:
1. Add purge script to base_omsagent.data
2. Have purge script download from Github if initial install fails (like TST)


aswatt:
Bump on this PR for after the holidays? I forgot it's been sitting here without review for a few weeks. PR is necessary for a compliancy requirement - @Lukeout has more information

installer/bundle/bundle_skel.sh
```diff
@@ -468,6 +472,24 @@ install_troubleshooter()
     return 0
 }
 
+install_purge_script()
```
hestolz, line 475: Isn't inclusion in `base_omsagent.data` the only requisite for this file being installed? I understand the desire for more robustness but I think that install path is quite proven given that is what is being used to install all the plugin files etc.

aswatt: I added this for the same reason I added it for the troubleshooter - specifically in cases where installation fails due to some unforeseen circumstances, I wanted the purge script to be on the machine so that any possible remnants of OMS could be completely purged as to not get in the way of future installations.

Although I'm not extremely familiar with how the OMS cleanup from a failed install works - if you think that's sufficient, I can change it? I don't want it to be redundant or anything

aswatt: @hestolz upon going over this I was thinking of keeping it in mainly because the troubleshooter installs on the machine even if OMS installation fails, so the purge script would allow for users to purge the troubleshooter if they wanted

hestolz (approved):

PR #1286, opened 2020-12-17
Add OMSConfig error code to documentation

aswatt:
Shell bundle error code 18 denotes an error in OMSConfig (aka DSC) installation ([OMSAgent code here](https://github.com/microsoft/OMS-Agent-for-Linux/blob/master/installer/bundle/bundle_skel.sh#L64)), explicitly adding this to list of shell bundle error codes documented.

Joint PR to keep consistent with Azure Docs; [Azure Docs PR here](https://github.com/MicrosoftDocs/azure-docs-pr/pull/141254)


hestolz (approved):

hestolz (approved):
Could also add a link to the DSC repo and put messaging indicating that any github issues should be opened appropriately in that repo rather than this one

aswatt:
> Could also add a link to the DSC repo and put messaging indicating that any github issues should be opened appropriately in that repo rather than this one

I was thinking about doing that but wasn't sure since none of the other dependencies mention that, should I add it to all of them do you think?

hestolz:
> > Could also add a link to the DSC repo and put messaging indicating that any github issues should be opened appropriately in that repo rather than this one
> 
> I was thinking about doing that but wasn't sure since none of the other dependencies mention that, should I add it to all of them do you think?

I think it would be worth it, we could also add for the other dependencies. Seems we get semi-frequent github issues opened in this repo that belong in DSC/OMI/auoms/etc.

aswatt:
Added specific repos + links for DSC, OMI, and SCX; after the holidays would appreciate a second look to ensure all of the links are to the correct repos, and that I didn't miss any repos!

hestolz (approved):

PR #1290, opened 2020-12-18
OMI hang documentation

hestolz:
On a couple occasions it has been noted that a hung OMI agent process will correspondingly cause OMS agent to hang. This is reproduceable on an agent where perf counters are enabled:
- send SIGSTOP to omiagent; omsagent will stop processing data and no more lines will be written to `omsagent.log`
- send SIGCONT to omiagent; omsagent will recover and immediately write many lines to `omsagent.log`
- if all perf counters are removed from `omsagent.conf` and related `.conf` files, SIGSTOP of omiagent has not effect on omsagent

The reason for this is that omsagent collects perf counters from omiagent in a non-threaded manner. The call to enumerate counters is made [here](https://github.com/microsoft/OMS-Agent-for-Linux/blob/master/source/code/plugins/oms_omi_lib.rb#L145) and handled in [omi_interface.cpp](https://github.com/microsoft/OMS-Agent-for-Linux/blob/master/source/code/plugins/omi_interface.cpp) where the collection timeout either does not work properly or it is set too high ([90s](https://github.com/microsoft/OMS-Agent-for-Linux/blob/master/source/code/plugins/omi_interface.cpp#L592)). 

Given that the changes to this path would be difficult (omi_interface.cpp not documented) and this is a single point of failure for all perf counter collection, and considering the timeline for AMA introduction and the future of OMS, and the historical lack of any incidents related to this 4-5 year-old bug, we will opt out of making changes to address this. Instead, adding documentation here, in azure docs, and internal wiki to mitigate possible impact.

### DIFF
--- a/docs/Troubleshooting.md
+++ b/docs/Troubleshooting.md
@@ -341,6 +341,7 @@ Follow these steps to recover from a bad state, like changing/removing important
 * Connection to the OMS Service is blocked
 * VM was rebooted
 * OMI package was manually upgraded to a newer version compared to what was installed by OMS Agent package
+* OMI is frozen, blocking OMS agent
 * DSC resource logs "class not found" error in `omsconfig.log` log file
 * OMS Agent for Linux data is backed up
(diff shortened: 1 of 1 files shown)


nidhanda (approved):
Thanks for the detailed explanation in the PR description. This will be a helpful reference for future cases (if there are any) related to this

PR #1293, opened 2020-12-24
Make CEF format regex not greedy

NoamLandress:
Investigation an open CRI, we discovered that the current regex is too greedy in if there is somewhere in the log message 
another occurrence of "CEF:0|" than the message gets trimmed in the middle.
This little change tells the regex not to be greedy and to stop on the first occurrence of "CEF:0|" which is the header in the beginning of the message.


NarineM (approved):

hestolz:

PR #1301, opened 2021-01-16
Update TST tests to use openssl

aswatt:
Improvements to TST connection tests - specifically, updating all URL connection checks to use OpenSSL, as well as some smaller things: changing internet connect to check docs.microsoft.com (if there's a better site to use for general internet connectivity please let me know!) and a small bug fix with checking internet connectivity when grabbing the TSG from Github.


source/code/troubleshooter/modules/connect/check_endpts.py
```diff
@@ -1,27 +1,33 @@
...
 
-import socket
+import subprocess
 
 from error_codes import *
 from errors      import error_info
 from helpers     import geninfo_lookup
 
 OMSADMIN_PATH = "/etc/opt/microsoft/omsagent/conf/omsadmin.conf"
+SSL_CMD = "echo | openssl s_client -connect {0}:443 -brief"
```
hestolz, line 10: might be worth also checking that the certs/cert chain is verified, separately ... that's `Verification: OK` portion. Off the top of my head verification could fail in cases where the VM does not have the root certs used to sign the server cert; VM will say "I can't trust that you are who you say you are." There may be more verification failure modes but none come to mind at the moment.
```
root@____:~# echo | openssl s_client -connect 4e7403f8-515a-4df5-8e13-59f0e2b76e3a.ods.opinsights.azure.com:443 -brief
CONNECTION ESTABLISHED
Protocol version: TLSv1.2
Ciphersuite: ECDHE-RSA-AES256-GCM-SHA384
Requested Signature Algorithms: RSA+SHA256:RSA+SHA384:RSA+SHA1:ECDSA+SHA256:ECDSA+SHA384:ECDSA+SHA1:DSA+SHA1:RSA+SHA512:ECDSA+SHA512
Peer certificate: CN = eapeus2-oi-ods.trafficmanager.net
Hash used: SHA256
Signature type: RSA
Verification: OK
Server Temp Key: ECDH, P-384, 384 bits
DONE
```

Additionally, you could do the same call with the actual agent certs, although this would probably be best to do in a separate function, since TST could be run without having onboarded AKA no certs present yet
```
root@____:~# echo | openssl s_client -connect 4e7403f8-515a-4df5-8e13-59f0e2b76e3a.ods.opinsights.azure.com:443 -prexit -brief -state -cert /etc/opt/microsoft/omsagent/certs/oms.crt -key /etc/opt/microsoft/omsagent/certs/oms.key
```

hestolz (approved):
lgtm with some enhancement suggestions

aswatt:
Sorry this is so late! Forgot to push the changes

source/code/troubleshooter/modules/connect/check_endpts.py
aswatt: I added an optional check with certs if the general command fails, as well as including a check for verification as well as connection. Let me know if anything looks off!

PR #1323, opened 2021-03-26
Skip omsadmin.conf check

aswatt:
When running the installation portion of the TST, it checks for the existence of omsadmin.conf, which isn't created until onboarding.

### DIFF
--- a/source/code/troubleshooter/modules/install/check_files.py
+++ b/source/code/troubleshooter/modules/install/check_files.py
@@ -197,6 +197,9 @@ def check_dirs(dirs, exist_err, perms_err):
 def check_files(files, exist_err, perms_err):
     success = NO_ERROR
     for f in (files.keys()):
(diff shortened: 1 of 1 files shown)


source/code/troubleshooter/modules/install/check_files.py
```diff
@@ -197,6 +197,9 @@ def check_dirs(dirs, exist_err, perms_err):
 def check_files(files, exist_err, perms_err):
     success = NO_ERROR
     for f in (files.keys()):
+        # skip over omsadmin.conf, as it's created in onboarding
+        if (f.endswith("omsadmin.conf")):
+            continue
```
hestolz, line 202: maybe print an info message? may be informative to know if it is missing; e.g. install and onboarding reported success but later the file is deleted

hestolz (approved):

PR #1324, opened 2021-03-31
ipcache: set 3sec timeout during dns resolve

abenbachir:


source/code/plugins/oms_common.rb
```diff
@@ -922,7 +928,8 @@ def refresh_cache
         @cache_lock.synchronize {
           @condition.wait(@cache_lock, @refresh_interval_seconds)
           # Flush the cache completely to prevent it from growing indefinitly
```
AaronRM, line 930: Disabling this would mean that the cache now can grow indefinitely. Is that the behavior we want?

source/code/plugins/oms_common.rb
```diff
@@ -922,7 +928,8 @@ def refresh_cache
...
-          @cache = {}
+          # Disabling refresh functionality to prevent data loss btw rsyslog and OMS
+          # @cache = {}
```
AaronRM, line 932: This appears to be the only cache invalidation policy. Disabling this means the cache will contain more and more invalid data over time.

AaronRM (changes requested):

source/code/plugins/oms_common.rb
```diff
@@ -891,7 +891,13 @@ def initialize(refresh_interval_seconds)
       @cache_lock = Mutex.new
       @refresh_interval_seconds = refresh_interval_seconds
       @condition = ConditionVariable.new
-      @thread = Thread.new(&method(:refresh_cache))
+      # Important: Disabling refresh functionality to prevent data loss btw rsyslog and OMS
+      # This bug happen because OMS is blocked 15 seconds due to DNS timeout when resolving
```
AaronRM, line 895: How would disabling the cache refresh/invalidation address the issue of DNS lookups timing out?

AaronRM (changes requested):

source/code/plugins/oms_common.rb
abenbachir: During OMS start-up there would be some DNS looks timeout but these will be cached as null values. So if the same unkown hostname were sent later it won't cause OMS to timeout again. By removing the 5min refresh this will prevent OMS from being blocked and lose data on each 5min interval.

abenbachir: Setting the timeout to 3 seconds may help. I have updated the PR.

AaronRM (approved):

PR #1342, opened 2021-06-14
Removing the limitation on number of characters for the network devices

jatakiajanvi12:
There is a limitation currently to only take upto 9 characters for the name of the network device interface.

Removing this limitation.

### DIFF
--- a/source/code/plugins/VMInsightsDataCollector.rb
+++ b/source/code/plugins/VMInsightsDataCollector.rb
@@ -323,7 +323,7 @@ module VMInsights
                     line = line.split(" ")
                     next if line.empty?
                     dev = line[0]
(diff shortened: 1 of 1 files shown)


source/code/plugins/VMInsightsDataCollector.rb
```diff
@@ -323,7 +323,7 @@ def get_net_data
                     line = line.split(" ")
                     next if line.empty?
                     dev = line[0]
-                    next unless ((0...10).include? dev.length) && (dev.end_with? ":")
+                    next unless (dev.end_with? ":")
```
mthompson83, line 326: Test case changes?

jatakiajanvi12: No changes needed

abenbachir (approved):

PR #1349, opened 2021-07-27
Add config tags in RPM package for specific files

aswatt:
Moves affected files from base_omsagent.data to the specific linux_dpkg.data and linux_rpm.data files, and adds the config tag to the files in linux_rpm.data.

More details about the issue can be found here: https://msazure.visualstudio.com/One/_workitems/edit/9984439/

### DIFF
--- a/installer/datafiles/base_omsagent.data
+++ b/installer/datafiles/base_omsagent.data
@@ -95,16 +95,12 @@ MAINTAINER:              'Microsoft Corporation'
 /opt/microsoft/omsagent/plugin/security_lib.rb;                         source/code/plugins/security_lib.rb;                   744; root; root
 /opt/microsoft/omsagent/plugin/filter_syslog_security.rb;               source/code/plugins/filter_syslog_security.rb;         744; root; root
 /opt/microsoft/omsagent/plugin/filter_changetracking.rb;                source/code/plugins/filter_changetracking.rb;          744; root; root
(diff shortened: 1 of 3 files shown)


simathih (approved):

hestolz:
@aswatt what about `OMSAuditdPlugin.sh` ? This was called out in the ICM but I don't see it anywhere here

aswatt:
> @aswatt what about `OMSAuditdPlugin.sh` ? This was called out in the ICM but I don't see it anywhere here

@hestolz I was looking through the original data files but I couldn't find anywhere that copied over `OMSAuditdPlugin.sh` - all the other files were already explicitly specified in `base_omsagent.data` initially. I didn't want to add it in in case there was some unintended side effects, but if it's not explicitly specified here then theoretically it shouldn't be an issue? The original problem was caused by files explicitly specified in the rpm spec file getting updated, and I can't find a way that this file would be explicitly specified.

hestolz:
@aswatt [here](https://github.com/microsoft/PowerShell-DSC-for-Linux/blob/master/installbuilder/datafiles/Base_DSC.data#L81) in the DSC repo. If we don't fix that one customer security tool will still be flagging them and we won't be able to resolve

aswatt:
@hestolz I'll add the change then to the DSC branch, thank you for finding that!

PR #1367, opened 2021-09-18
OMIcheck: trigger goal state change in GA for faster extension update

canfikret:
If an update is requested and OMSLinuxAgent is present, nudge GA to pickup latest bits. If OMI is standalone, update it from github.

### DIFF
--- a/tools/OMIcheck/OMIcheck.ps1
+++ b/tools/OMIcheck/OMIcheck.ps1
@@ -1,6 +1,12 @@
-﻿$upgradeOMI = $false; #detect only
+﻿# This script lists the virtual machines in Azure subscriptions of the user,
+# and looks for vulnerable OMI packages. The report is grouped by subscription and resource group name.
 
-# Update these paths accordingly.
+# Set the flag below to tru if you want the vulnerable OMI to be patched.
+# If OMSLinuxAgent is installed, then a minor update will be triggered.
+# If OMI is standalone, the latest bits will be installed from Github.
+$upgradeOMI = $false; #detect only
+
+# Update these paths accordingly. In cloud shell, the path is usually /home/user_name
 $checkScriptPath  = "C:\linux\OMIcheck\omi_check.sh"
 $upgradeScriptPath = "C:\linux\OMIcheck\omi_upgrade.sh"
 
@@ -29,18 +35,21 @@ foreach ($sub in $subs)
             #    continue
             #}
 
+            # Group VMs by resource groups.
             if ($PreviousRG -ne $VM.ResourceGroupName)
             {
                 Write-Host "`tResourceGroup: " $VM.ResourceGroupName
                 $PreviousRG = $VM.ResourceGroupName
             }
 
+            # Update only running VMs.
             if ($VM.PowerState -ne "VM running")
             {
                 Write-Host -ForegroundColor Gray `t`t $VM.Name " in resource group : VM is not running"
                 continue
             }
 
+            # Update Linux only VMs.
             if ($VM.StorageProfile.OsDisk.OsType.ToString() -ne "Linux")
             {
                 Write-Host -ForegroundColor Gray `t`t $VM.Name ": VM is not running Linux OS"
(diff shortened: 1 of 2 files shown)


abenbachir (approved):

tools/OMIcheck/OMIcheck.ps1
```diff
@@ -1,6 +1,12 @@
-﻿$upgradeOMI = $false; #detect only
+﻿# This script lists the virtual machines in Azure subscriptions of the user,
+# and looks for vulnerable OMI packages. The report is grouped by subscription and resource group name.
 
-# Update these paths accordingly.
+# Set the flag below to tru if you want the vulnerable OMI to be patched.
```
canfikret, line 4: Minor typo: tru -> true.

tools/OMIcheck/OMIcheck.ps1
```diff
@@ -63,15 +72,31 @@ foreach ($sub in $subs)
...
                     {
-                        $upgrade = Invoke-AzVMRunCommand -ResourceGroupName $VM.ResourceGroupName -Name $VM.Name -CommandId 'RunShellScript' -ScriptPath $upgradeScriptPath
-                        Write-Host -ForegroundColor Red `t`t $VM.Name ": Result of OMI package upgrade attempt: " $upgrade.Value.Message
+                        # If OMSLinuxAgent is installed, nudge the Azure Guest Agent to pickup the latest bits.
+                        $VMobj = Get-AzVM -ResourceGroupName $VM.ResourceGroupName -Name $VM.Name
+                        $omsExt = $VMobj.Extensions | where { ($_.Publisher -eq "Microsoft.EnterpriseCloud.Monitoring" -and $_.VirtualMachineExtensionType -eq "OmsAgentForLinux") }
+                        if ($omsExt -ne $null)
+                        {
+                            # Trigger a goal state change for GA.
+                            Set-AzVMExtension -VMName $VM.Name -ResourceGroupName $VM.ResourceGroupName -Location $VM.Location -Publisher $omsExt.Publisher -Name $omsExt.Name -ExtensionType $omsExt.VirtualMachineExtensionType
```
hestolz, line 81: Why not set version as 1.13.40, will it necessarily upgrade if left unspecified?

hestolz: Also, should we run this `--no-wait` (or w/e the powershell equivalent is)? It's a trade-off between seeing the script output the result of each operation and the script running much faster

tools/OMIcheck/OMIcheck.ps1
```diff
@@ -63,15 +72,31 @@ foreach ($sub in $subs)
...
+                            Write-Host -ForegroundColor Green `t`t $VM.Name ": Set OMSLinuxAgent extension goal state for update"
```
hestolz, line 82: This will report success regardless of exit code of ^

tools/OMIcheck/OMIcheck.ps1
canfikret: I had tried setting that arg. -Version "1.13.40" and it gave me 400 bad request with ErrorMessage: The value of parameter typeHandlerVersion is invalid.

The error message didn't make sense.  Now, I tried "1.13" only and that proceeded.

AaronRM (approved):

tools/OMIcheck/OMIcheck.ps1
canfikret: RE: -no-wait (-NoWait in PS) would be fire&forget. Customer would need to run the script second time to validate update took place.

Another concern is ARM throttling on number of operations. Without wait we can submit 40-50 operations per minute. That is withing boundary of 1200 ops/hour.

tools/OMIcheck/OMIcheck.ps1
canfikret: Yes, that's by design. Now, with -NoWait; we'd like to fire an forget..
The customer will run the script 2nd time (after 10 mins or so), to validate all are green.

tools/OMIcheck/OMIcheck.ps1
hestolz: yep good point about ARM rate limits

hestolz: > I had tried setting that arg. -Version "1.13.40" and it gave me 400 bad request with ErrorMessage: The value of parameter typeHandlerVersion is invalid.
> 
> The error message didn't make sense. Now, I tried "1.13" only and that proceeded.

Cool, IME (but this was a while ago) running without version specification would do nothing, but if you were able to test this and see otherwise, that's perfect.

tools/OMIcheck/README.md
```diff
@@ -14,4 +14,6 @@
 
 - It run a remote command to detect OMI version, and reports findings.
 
-- Optional flag to install omi 1.6.8.1 is off by defualt.
\ No newline at end of file
+- Optional flag to install omi 1.6.8.1 is off by defualt.
```
hestolz, line 17: nit `defualt` -> `default`

hestolz (approved):

tools/OMIcheck/README.md
canfikret: Thx, will fix it in next iteration.

PR #1378, opened 2021-09-22
OMI detect: add VMSS and experimental ARC support

canfikret:


tools/OMIcheck/OMIcheck.ps1
```diff
@@ -1,127 +1,460 @@
...
-# If OMI is standalone, the latest bits will be installed from Github.
-$upgradeOMI = $false; #detect only
-$omsVersionGood = "1.13"
+#    - If OMSLinuxAgent or LinuxDiagnostics is installed, then a minor update will be triggered.
+#    - OMI package is installed from Github.
+$upgradeOMI = $false #detect only
+
+$OmiServerGoodVersion = "OMI-1.6.8-1"
+$OmiPkgGoodVersion = "1.6.8.1"
+$LadPkgGoodVersion = "1.5.110-LADmaster.1483"
```
AaronRM, line 11: This version of mdsd shipped in both LAD 3.0.131 and LAD 3.0.135. The presence of this package version wouldn't tell us whether or not LAD needs to be updated.

AaronRM (approved):

tools/OMIcheck/OMIcheck.ps1
canfikret: The VM extension also lacks the patch number. Therefore, I update the OMI package regardless.

tools/OMIcheck/OMIcheck.ps1
```diff
@@ -1,127 +1,460 @@
...
+$OmsPkgGoodVersion = "1.13.40.0"
```
hestolz, line 12: it is `1.13.40-0` not `.0`

tools/OMIcheck/OMIcheck.ps1
```diff
@@ -1,127 +1,460 @@
...
-# If OMSLinuxAgent is installed, then a minor update will be triggered.
-# If OMI is standalone, the latest bits will be installed from Github.
-$upgradeOMI = $false; #detect only
-$omsVersionGood = "1.13"
+#    - If OMSLinuxAgent or LinuxDiagnostics is installed, then a minor update will be triggered.
+#    - OMI package is installed from Github.
+$upgradeOMI = $false #detect only
+
+$OmiServerGoodVersion = "OMI-1.6.8-1"
+$OmiPkgGoodVersion = "1.6.8.1"
```
hestolz, line 10: `1.6.8-1`?

tools/OMIcheck/OMIcheck.ps1
```diff
@@ -1,127 +1,460 @@
...
-$subs = Get-AzSubscription
+# Parse the response string.
+function ParsePkgVersions($str)
+{
+    $splitLines = $str.Split("`n",[System.StringSplitOptions]::RemoveEmptyEntries)
 
-# For each subscription execute to find (and upgrade) OMI on Linux VMs
-# TODO: This can be converted to ForEach-Object -Parallel. However, it requires PSh >= 7.1
-foreach ($sub in $subs)
+    $omiServerResp   = $splitLines | where { $_.Contains("/opt/omi/bin/omiserver:") }
```
hestolz, line 28: seems like we no longer do this check in the `.sh`, just `grep -E 'omi|omsagent|lad'`

tools/OMIcheck/OMIcheck.ps1
```diff
@@ -1,127 +1,460 @@
...
+    $omiPkgLine      = $splitLines | where { $_.Contains(" omi ") }
+    $ladMdsdPkgLine  = $splitLines | where { $_.Contains(" lad-mdsd ") }
+    $omsAgentPkgLine = $splitLines | where { $_.Contains(" omsagent ") }
+    if ($omiServerResp -ne $null -and $omiServerResp.Length -gt 0)
+    {
+        $split = $omiServerResp.Split(" ",[System.StringSplitOptions]::RemoveEmptyEntries)
+        $omiVer = $split[1]
+
+        if ($omiPkgLine -ne $null)
```
hestolz, line 37: i think these checks for omi/lad/oms shoud not be inside the `if ($omiServerResp...`, should be at function scope

tools/OMIcheck/OMIcheck.ps1
```diff
@@ -1,127 +1,460 @@
...
+function UpdateOMSinVMSS($VMss)
+{
+    Write-Host -ForegroundColor Red `t`t $VMss.Name ": VM scale set has vulnerable OMI version " $pkgVer
+
+    $omsExt = $VMss.VirtualMachineProfile.ExtensionProfile.Extensions | where { $_.Publisher -eq $OMSpublisher -and $_.Type -eq $OMSextName }
+    if ($omsExt -ne $null)
+    {
+        # Trigger a goal state change for GA.
+        $omsExt.AutoUpgradeMinorVersion = $true
+        Update-AzVmss -VirtualMachineScaleSet $VMss -ResourceGroupName $VMss.ResourceGroupName -VMScaleSetName $VMss.Name
```
hestolz, line 159: if `AutoUpgradeMinorVersion` is already true, will `Update-AzVmss` still do anything?

tools/OMIcheck/OMIcheck.ps1
```diff
@@ -1,127 +1,460 @@
...
+function UpdateLAD($VM)
+{
+    $exts = Get-AzVMExtension -VMName $VM.Name -ResourceGroupName $VM.ResourceGroupName
+    $ladExt = $exts | where { ($_.Publisher -eq $LADpublisher -and $_.Name -eq $LADextName) }
+    if ($ladExt -ne $null)
+    {
+        if ($ladExt.ProvisioningState -eq "Succeeded")
+        {
+            # Trigger a goal state change for GA.
+            Set-AzVMExtension -VMName $VM.Name -ResourceGroupName $VM.ResourceGroupName -Location $ladExt.Location -Publisher $ladExt.Publisher -Name $ladExt.Name -ExtensionType $ladExt.ExtensionType -TypeHandlerVersion $ladExt.TypeHandlerVersion
```
hestolz, line 95: could save the response and check like
```
$response = Set-AzVMExtension ...
if ($response.IsSuccessStatusCode) ...

Comment (hestolz) [reply tools/OMIcheck/OMIcheck.ps1:95]: along with the other `Update*` functions

File: tools/OMIcheck/OMIcheck.ps1 [hunk 8]
```diff
@@ -1,127 +1,460 @@
...
+            if ($check.LadPkgVer -ne $LadPkgGoodVersion)
             {
-                Write-Host -ForegroundColor Gray `t`t $VM.Name " in resource group : VM is not running"
-                continue
+                UpdateLAD($VM)
             }
 
-            # Update Linux only VMs.
-            if ($VM.StorageProfile.OsDisk.OsType.ToString() -ne "Linux")
+            if ($check.OmsPkgVer -ne $OmsPkgGoodVersion)
```
hestolz, line 272: i thought we didn't want to upgrade OMS if we upgraded LAD already, in case of compat issues

tools/OMIcheck/OMIcheck.ps1
```diff
@@ -1,127 +1,460 @@
...
+        {
+            continue
+        }
+        if ($VMss.VirtualMachineProfile.OsProfile.LinuxConfiguration -eq $null)
+        {
+            Write-Host -ForegroundColor Gray `t`t $VMss.Name " is not running Linux. Skipping."
+            continue
+        }
+
+        $check = IsVMssVulnerableOMI $VMss $VMinstances[0]
```
hestolz, line 330: seems unlikely but in this case, if the first instance was different than other instances (e.g. for some reason doesn't have OMI installed but others do), we could skip when we should actually be patching

tools/OMIcheck/OMIcheck.ps1
canfikret: That's the package version, for example,  dpkg -l | grep omsagent
ii  omsagent                               1.13.40.0

tools/OMIcheck/OMIcheck.ps1
canfikret: Same as above.
ii  omi                                    1.6.8.1

tools/OMIcheck/OMIcheck.ps1
canfikret: doing both  checks.

tools/OMIcheck/OMIcheck.ps1
canfikret: The determining factor is the version string from omiserver. We bail out when there is no omi installed.

tools/OMIcheck/OMIcheck.ps1
canfikret: VMSS is uniform; that's same bits are on all instances. We got to pick an instance. Going thru all of them would take long time.

tools/OMIcheck/OMIcheck.ps1
canfikret: Relying on the running server version made more sense.

tools/OMIcheck/OMIcheck.ps1
canfikret: sure.

tools/OMIcheck/OMIcheck.ps1
canfikret: No, good catch. I added -ForceRerun param.

tools/OMIcheck/OMIcheck.ps1
canfikret: Given that now both LAD and OMS is available in all regions; it makes sense to update both. Also, the delta in the hotfix is minor; so previous versions compat should apply. what do you think?

PR #1382, opened 2021-09-28
added check for vulnerable scaleset instance

johan-afm:
As I understand it, scaleset instances may also be vulnerable. This PR adds a script with the same check against scaleset instances as was done against VMs in commit 31becbfae17dee313b566a2bede799a6b3a86de2. Since then, I can see that additional checks have been added to the original script, but perhaps this may serve as a starting point for incorporating checks against scaleset instances as well.


hestolz:

PR #1388, opened 2021-10-08
Fix 'microsoft' directory removal check

aswatt:
Current check always returned that the directory was empty, fixed it so it actually checks if the directory is empty.


NarineM:
This change is not needed and PR can be closed.  Here is the PR for directory check we ll take: https://github.com/microsoft/OMS-Agent-for-Linux/pull/1432

NarineM:
This change is not needed and PR can be closed.  Here is the PR for directory check we ll take: https://github.com/microsoft/OMS-Agent-for-Linux/pull/1432

PR #1390, opened 2021-10-12
OMIcheck: add loging to file, parallel for-each loops to speed up, and CSS feedback

canfikret:


tools/OMIcheck/OMIcheck.ps1
```diff
@@ -1,53 +1,73 @@
-﻿# This script lists the virtual machines and scale sets in Azure subscriptions of the user,
+﻿# This version of the script requires Powershell version >= 7 in order to improve performance via ForEach-Object -Parallel
```
AaronRM, line 1: Suggestion: For a future iteration, you may want to run this script through PSScriptAnalyzer for static analysis.

canfikret: Thanks for the feedback. I've been relying on VS code's squiggles from Powershell extension. And it uses PSScriptAnalyzer https://code.visualstudio.com/docs/languages/powershell#_psscriptanalyzer

I will run standalone next iteration to see if any difference.

PR #1403, opened 2021-11-09
TST fix for Python 3.8 deprecation

aswatt:
Python 3.8+ deprecates platform.linux_distribution() and platform.dist(), causing the troubleshooter to break when trying to figure out what OS is being used. I added a check to see if distro.linux_distribution() works (the replacement for above), as well as fixed the logic so if neither works it should grab the info from /etc/os_release, rather than erroring out.


source/code/troubleshooter/modules/helpers.py
```diff
@@ -13,6 +12,16 @@
 except ImportError:
     from urllib2 import urlopen
 
+# linux_distribution() function was moved to distro for Python 3.8+
+try:
+    from platform import linux_distribution
+    from platform import dist
+except ImportError:
+    try:
+        from distro import linux_distribution
```
hestolz, line 21: is `distro` part of python stdlib?

aswatt: I'm not sure, but on an Ubuntu20 VM with Python 3.8 `import distro` works without any issues

hestolz (approved):

PR #1415, opened 2021-12-27
fix cpu utilization

pfrcks:
- use /proc/stat to calculate cpu utilization instead of /proc/uptime

### DIFF
--- a/source/code/plugins/VMInsightsDataCollector.rb
+++ b/source/code/plugins/VMInsightsDataCollector.rb
@@ -22,7 +22,7 @@ module VMInsights
             @saved_net_data = get_net_data
             @saved_disk_data.baseline
             t, i = get_cpu_idle
-            { :total => t, :idle => i }
+            { :total_time => t, :idle => i }
         end
 
         def start_sample
@@ -64,18 +64,18 @@ module VMInsights
         # eg: cpu  2904083 315778 1613190 140077550 216726 0 88355 0 0 0
         # https://www.kernel.org/doc/html/latest/filesystems/proc.html#miscellaneous-kernel-statistics-in-proc-stat
         def get_cpu_idle
-            total = nil
+            total_time = nil
             idle = nil
             File.open(File.join(@root, "proc", "stat"), "rb") { |f|
                 line = f.gets
                 raise Unavailable, "/proc/stat empty" if line.nil?
-                a = line.split(" ")
+                time_entries = line.split(" ")
                 # cpu user nice system idle - remaining entries depend on kernel version
                 raise Unavailable, "/proc/stat insufficient entries" if a.length < 5
-                a = a.slice(1, a.length) # skip the first entry in row: "cpu"
-                idle, total = a[3].to_i, a.map(&:to_i).sum
+                time_entries = time_entries.slice(1, time_entries.length) # skip the first entry in row: "cpu"
+                idle, total_time = time_entries[3].to_i, time_entries.map(&:to_i).sum
                 }
-            return total, idle
+            return total_time, idle
         end
 
         # returns:
--- a/source/code/plugins/VMInsightsDataCollector.rb
+++ b/source/code/plugins/VMInsightsDataCollector.rb
@@ -71,9 +71,11 @@ module VMInsights
                 raise Unavailable, "/proc/stat empty" if line.nil?
                 time_entries = line.split(" ")
                 # cpu user nice system idle - remaining entries depend on kernel version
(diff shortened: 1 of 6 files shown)


source/code/plugins/VMInsightsDataCollector.rb
```diff
@@ -54,23 +54,28 @@ def get_available_memory_kb
...
             idle = nil
-            File.open(File.join(@root, "proc", "uptime"), "rb") { |f|
+            File.open(File.join(@root, "proc", "stat"), "rb") { |f|
                 line = f.gets
-                next if line.nil?
-                line.scanf(" %f %f ") { |u, i|
-                    uptime = u
-                    idle = i
+                raise Unavailable, "/proc/stat empty" if line.nil?
+                a = line.split(" ")
```
jatakiajanvi12, line 72: "a" doesn't seem an intuitive name for a variable. Can you change to something meaningful?

jatakiajanvi12 (approved):

hestolz (approved):

source/code/plugins/VMInsightsDataCollector.rb
```diff
@@ -54,23 +54,28 @@ def get_available_memory_kb
...
-                next if line.nil?
-                line.scanf(" %f %f ") { |u, i|
-                    uptime = u
-                    idle = i
+                raise Unavailable, "/proc/stat empty" if line.nil?
+                time_entries = line.split(" ")
+                # cpu user nice system idle - remaining entries depend on kernel version
+                raise Unavailable, "/proc/stat insufficient entries" if time_entries.length < 5
+                time_entries = time_entries.slice(1, time_entries.length) # skip the first entry in row: "cpu"
+                idle, total_time = time_entries[3].to_i, time_entries.map(&:to_i).sum
```
mthompson83, line 76: This implementation does not match [get_cpu_time](https://msazure.visualstudio.com/InfrastructureInsights/_git/DependencyAgent?path=/collector/service/platform/linux/machine-metrics.cc&version=GBmaster&line=104&lineEnd=105&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contentsl)

mthompson83, line 76: If all the underlying kernel counters are 64-bit as assumed in [get_cpu_time](https://msazure.visualstudio.com/InfrastructureInsights/_git/DependencyAgent?path=/collector/service/platform/linux/machine-metrics.cc&version=GBmaster&line=104&lineEnd=105&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contentsl), and Jiffies are on the order of 1/100th of a second, then rollover isn't a concern. The C++ implementation implicitly will handle rollover, but the ruby implementation does not.

test/code/plugins/VMInsightsDataCollector_subtest.rb
```diff
@@ -386,34 +400,43 @@ def test_get_filesystems_live
...
+        end
+
+        def test_proc_stat_garbage
+            File.open(@proc_stat, WriteASCII) { |f|
+                f.puts "ga rb age tr ash"
+            }
+            total_time, idle  = @object_under_test.get_cpu_idle
+            assert_equal 0, total_time
+            assert_equal 0, idle
         end
```
mthompson83, line 441: test case for first word on the line is not "cpu" but everything else is ok.

test/code/plugins/VMInsightsDataCollector_subtest.rb
```diff
@@ -386,34 +400,43 @@ def test_get_filesystems_live
...
-            assert ex.message.include?("Uptime not found"), ex.inspect
+            assert ex.message.include?("/proc/stat empty"), ex.inspect
         end
 
-        def test_proc_uptime_garbage
-            File.open(@proc_uptime, WriteASCII) { |f|
-                f.puts "gobble", "d", "gook"
+        def test_proc_stat_insufficient_entries
+            File.open(@proc_stat, WriteASCII) { |f|
+                f.puts "cpu 10 20"
```
mthompson83, line 425: Make this exactly one element less than minimum. Add a test case with exactly the minimum number of elements.

mthompson83 (changes requested):

source/code/plugins/VMInsightsDataCollector.rb
pfrcks: not sure I understand. Are you referring to the "* 10000LL" in the C++ code?

mthompson83 (approved):

Issue #1419, opened 2022-01-14
Feature Request: support Alma Linux


athiede-vendavo:
CentOS 8 is no longer supported. Several free rebuilds of RHEL exist in the form of Alma Linux and Rocky Linux. They are both available in Azure and are good options for those who want to continue using a free RHEL rebuild.

- https://almalinux.org/
- https://rockylinux.org/

Please support these operating systems in the agent.

athiede-vendavo:
Closing as dupe, since https://github.com/microsoft/OMS-Agent-for-Linux/issues/1363 exists. Just didn't realize it was there already since a search of 'alma' doesn't show an issue mentioning 'almalinux'.

Issue #1363, opened 2021-09-16
Support for AlmaLinux


johanburati:
I've seen some big annoncement that [AlmaLinux was now available on Azure Marketplace](https://twitter.com/AlmaLinux/status/1427649332946288653).
However it looks like this distro is still not supported by the **OmsAgentForLinux** extension:
```
2021/09/16 00:55:54 ERROR:Unsupported operating system: AlmaLinux 8.4
2021/09/16 00:55:54 ERROR:[Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux-1.13.39] Unsupported operating system: AlmaLinux 8.4
2021/09/16 00:56:01 [Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux-1.13.39] Install,failed,51,Unsupported operating system: AlmaLinux 8.4
```
**AlmaLinux** is supposed to be a drop in remplacement for **CentOS**, so it would be nice if you could add it to the supported distro list.

NicolasKarnick:
@johanburati where are the logs from? Cant find it on our machine. 
If Azure supports AlmaLinux, there has to be a support für OMS-Agent-for-Linux.

dse002:
Same here, log is from /var/log/waagent.log

johanburati:
> @johanburati where are the logs from? Cant find it on our machine. If Azure supports AlmaLinux, there has to be a support für OMS-Agent-for-Linux, too.

`/var/log/azure/Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux/extension.log`

dse002:
@NicolasKarnick @johanburati Any news on this?

tiagoabelha:
Yes, any news on this would be appreciated

athiede-vendavo:
Opened this: https://github.com/microsoft/OMS-Agent-for-Linux/issues/1363
I believe that'll make it installable. Tried to hack my way around it on an Alma Linux system and it was able to install with these changes. Still didn't report metrics.

NarineM:
Alma 8 support enabled in omsagent version 1.14.16.

PR #1420, opened 2022-01-18
add alma and rocky to check_os

athiede-vendavo:
Pretty sure this will solve https://github.com/microsoft/OMS-Agent-for-Linux/issues/1419 (edit - https://github.com/microsoft/OMS-Agent-for-Linux/issues/1363)
I did the manual install by editing this file and it seemed to work.


athiede-vendavo:
Is anyone available to help review this, so we can start using Alma Linux in Azure?
@abenbachir @hestolz @aswatt @simathih
Apologies for the at, but we really would like to see this included supported in the OMS Agent so we can start using it. This is a blocker for projects we're working on.

aswatt:
Thank you for the feedback - agent product group may consider this in the future. However, today OMS will continue to support base off the Azure Endorsed Distro list.

athiede-vendavo:
Thank you for the response, even if it's not what I had wanted. I appreciate you taking the time to look at this.

PR #1430, opened 2022-03-07
Change to add PackageClassification property into collection

nikhim-um:


NarineM (changes requested):
we only update to stable ruby version that is compatible with agent fluentd providers, there are also other files that need to be updated if ruby is updated. You need to come up with a solution for patch management that does not depend on a ruby version that is not part of the release, or wait till newer version is released.

NarineM:
Needs to include dsc related pieces.

NarineM:
No longer needed.

PR #1431, opened 2022-03-08
fix upgrade when the same version of omsagent package already installed

NarineM:


installer/bundle/bundle_skel.sh
```diff
@@ -1551,11 +1551,12 @@ you should first purge the existing installation and then install using the --sk
 
         # Update OMS Agent
         shouldInstall_omsagent
+        temp_status=$?
         rm -f "$OMS_CONSISTENCY_INVOKER" > /dev/null 2>&1
```
hestolz, line 1555: couldn't we just move this before `shouldInstall_omsagent` and have no other change necessary?

NarineM: Moving rm would not make any difference. This change is to explicitly  say that the return value of shouldinstall is used as a parameter in the update function. Because it was missing, it resulted in this bug.

hestolz: but imagine the code is like this

```
# Update OMS agent
rm -f "$OMS_CONSISTENCY_INVOKER" > /dev/null 2>&1
shouldInstall_omsagent
pkg_upd $OMS_PKG omsagent $?
```

How is `pkg_update` not using the return value of `shouldInstall_omsagent`?

NarineM:
@hestolz I think your review was reset. Can you re-submit it? thanks.

aswatt (approved):

installer/bundle/bundle_skel.sh
```diff
@@ -1550,12 +1550,12 @@ you should first purge the existing installation and then install using the --sk
         fi
 
         # Update OMS Agent
-        shouldInstall_omsagent
         rm -f "$OMS_CONSISTENCY_INVOKER" > /dev/null 2>&1
```
hestolz, line 1553: Sorry, one more thought ... is it safe to always rm the OMS_CONSISTENCY_INVOKER cron job here? Can't tell what scenarios it will be recreated per https://github.com/microsoft/OMS-Agent-for-Linux/blob/c7cff7009a78254c89b3e6d0ea24302e3b7930ae/installer/bundle/bundle_skel.sh#L1619

NarineM: @hestolz  yes, cron job needs to be deleted before agent and dsc are upgraded because if it stays in running it might pull config while these packages are being upgraded. It is being created by DSC LCM engine that is why you do not see it here. There are better ways of doing cron job clean up, but changing it now would not bring any functional impact.

PR #1432, opened 2022-03-09
avoid deleting directory unexpectedly

antomatody:
! "$(ls -A <dir> > /dev/null 2>&1)"
is always true, no matter it's empty or not .

### DIFF
--- a/tools/purge_omsagent.sh
+++ b/tools/purge_omsagent.sh
@@ -363,7 +363,7 @@ dir_rm /etc/opt/microsoft/auoms
 dir_rm /etc/opt/microsoft/dependency-agent
 dir_rm /etc/opt/microsoft/omsagent
 dir_rm /etc/opt/microsoft/scx
-if [ ! "$(ls -A /etc/opt/microsoft > /dev/null 2>&1)" ]; then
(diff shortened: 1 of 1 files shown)


NarineM:
@antomatody,looks look, please match the EOL at the end of the file and I can sign off.

tools/purge_omsagent.sh
```diff
@@ -363,7 +363,7 @@ dir_rm /etc/opt/microsoft/auoms
 dir_rm /etc/opt/microsoft/dependency-agent
 dir_rm /etc/opt/microsoft/omsagent
 dir_rm /etc/opt/microsoft/scx
-if [ ! "$(ls -A /etc/opt/microsoft > /dev/null 2>&1)" ]; then
+if [ "$(ls -A /etc/opt/microsoft 2>/dev/null  |wc -w)" = "0" ]; then
```
aswatt, line 366: Small nitpick, but this can be done a bit more cleanly by just checking if the output of the `ls` command is NULL, instead of piping the result and then counting the words in it:

`if [ ! "$(ls -A /etc/opt/microsoft)" ]; then`

aswatt (approved):

PR #1433, opened 2022-03-10
update auoms to latest kits

NarineM:


hestolz (approved):
LGTM, also tagging @taglines as auoms owner

PR #1435, opened 2022-03-14
Fix ODSIngestion status file write permissions

NarineM:
Removed write for all users permission for ODSIngestion status files.


hestolz (approved):
If file is already created (if there is previous install), will this also fix the permissions?

NarineM:
@hestolz no, it would not change already existing files, only when creating new file. However , this file is harmless, for existing environments or machines that are not opt in for auto updates , it can be manually mitigated.

NarineM:
@hestolz yes, we can do that, but since these files are recording temporally states  , they do not represent security threats and are not worth fixing for old installations . Also there will be always VMs that are not optin for auto updates and will stay with old permissions.

PR #1437, opened 2022-03-16
Improve error logging for bundle  install and onboarding script

NarineM:


installer/bundle/bundle_skel.sh
```diff
@@ -591,7 +591,7 @@ pkg_upd() {
     pkg_name=$2
     pkg_allowed=$3
 
-    echo "----- Upgrading package: $pkg_name ($pkg_filename) -----"
+    echo "----- Upgrading to package: $pkg_name ($pkg_filename) -----"
```
hestolz, line 594: I feel like the original message may be clearer; if the idea is to make it clear that we are upgrading to a new VERSION of the same package, we could do something like "Upgrading $pkg_name to version $pkg_filename" ?

hestolz (approved):
LGTM, one suggestion :)

installer/bundle/bundle_skel.sh
NarineM: @hestolz changed, i please review. Looks more clear now. thanks.

installer/bundle/bundle_skel.sh
```diff
@@ -1384,17 +1384,17 @@ case "$installMode" in
 
             if [ $temp_status -ne 0 ]; then
                 if [ $temp_status -eq 2 ]; then
-                    ErrStr="System Issue with daemon control tool "
+                    ErrStr="system issue with daemon control tool "
                     ErrCode=$DEPENDENCY_MISSING
                 elif [ $temp_status -eq 3 ]; then # dpkg is messed up
-                    ErrStr="omi server conf file missing"
+                    ErrStr="missng omi server conf file"
```
hestolz, line 1390: typo `missing`

PR #1263, opened 2020-11-09
Add --replacepkgs Flag

hestolz:
This is a potential fix to address the issues on RedHat 7.5 during extension install. The displayed error is

```
[ExtensionOperationError] Non-zero exit code: 17, /var/lib/waagent/Microsoft.EnterpriseCloud.Monitoring.OmsAgentForLinux-1.13.29/omsagent_shim.sh -install
[stdout]
2020/10/30 17:34:12 Shell bundle exiting with code 17


[stderr]
Python 2.7.5
able
Trying to stop omi with systemctl
omi is stopped.
Trying to start omi with systemctl
omi is started.
----- Upgrading package: omsagent (omsagent-1.13.29-0.universal.x64) -----
	package omsagent-1.13.29-0.x86_64 is already installed
omsagent-1.13.29-0.universal.x64 package failed to upgrade and exited with status 1
OMS Troubleshooter is installed.
You can run the Troubleshooter with the following command:

  $ sudo /opt/microsoft/omsagent/bin/troubleshooter

Shell bundle exiting with code 17
```

Info on this flag [here](http://ftp.rpm.org/max-rpm/s1-rpm-install-additional-options.html).

PR #1440, opened 2022-03-21
Revert replace flag fix but redhat 7.5

NarineM:
To minimize number of calls to package manager to install dependencies if in dependency hierarchy the same version of packages already exists on the system, this PR is reverting replace flag change and only using it as default flag for RedHat 7.5 (initial PR to resolve an issue reported for RedHat 7.5 : https://github.com/microsoft/OMS-Agent-for-Linux/pull/1263 ) . This should help with package manager getting into lock state issues we have seen while upgrading omi/scx packages.

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -610,34 +610,16 @@ pkg_upd() {
         return $?
     else
         [ -n "${forceFlag}" ] && FORCE="--force" || FORCE=""
-        rpm --upgrade --replacepkgs $FORCE ${pkg_filename}.rpm
+        # Temp workaround for an upgrade issue seen on RedHat 7.5.
+        # Only for RedHat 7.5 call upgrade with --replacepkgs flag by default and only if force flag is not set. 
+        redhat_75=`cat /etc/*-release 2>/dev/null | grep -iP '.*?red.*?7\.5' 2>&1>/dev/null; echo $?`
+        if [ $FORCE -ne "--force" ] && [ $redhat_75 -eq 0 ]; then
+            rpm --upgrade --replacepkgs $FORCE ${pkg_filename}.rpm
+        else
+            rpm --upgrade $FORCE ${pkg_filename}.rpm
+        fi
         return $?
     fi
-}
-
-get_arch()
-{
-    if [ $(getconf LONG_BIT) = 64 ]; then
-        echo "x64"
-    else
-        echo "x86"
-    fi
-}
-
-compare_arch()
-{
-    # Check if the user is trying to install the correct bundle (x64 vs. x86)
-    echo "Checking host architecture ..."
-    HOST_ARCH=$(get_arch)
-
-    case $OMS_PKG in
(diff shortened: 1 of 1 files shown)


installer/bundle/bundle_skel.sh
```diff
@@ -610,36 +610,17 @@ pkg_upd() {
         return $?
     else
         [ -n "${forceFlag}" ] && FORCE="--force" || FORCE=""
-        rpm --upgrade --replacepkgs $FORCE ${pkg_filename}.rpm
+        # Only for RedHat 7.5 call upgrade with --replacepkgs flag and only if force flag is not set
+        redhat_7.5=`cat /etc/*-release 2>/dev/null | grep -iP '.*?red.*?7\.5' 2>&1>/dev/null; echo $?`
```
hestolz, line 614: does rhel provide minor version in the `-release` file? just checked my CentOS machine and don't see it

hestolz (approved):

installer/bundle/bundle_skel.sh
NarineM: yes, it does.

installer/bundle/bundle_skel.sh
```diff
@@ -1219,7 +1200,6 @@ fi
 
 if [ "$installMode" = "I" -o "$installMode" = "U" ]; then
     compare_install_type
-    compare_arch
```
AaronRM, line 1222: Why remove this check?

installer/bundle/bundle_skel.sh
```diff
@@ -610,36 +610,17 @@ pkg_upd() {
         return $?
     else
         [ -n "${forceFlag}" ] && FORCE="--force" || FORCE=""
-        rpm --upgrade --replacepkgs $FORCE ${pkg_filename}.rpm
+        # Only for RedHat 7.5 call upgrade with --replacepkgs flag and only if force flag is not set
```
AaronRM, line 613: Can you add a comment here explaining why RH 7.5 needs a special case here? Do we know this is specific to RH 7.5 or could other similar major.minor versions (e.g. 7.4, 7.6, 8,0, etc...) need this change too?

installer/bundle/bundle_skel.sh
NarineM: we no longer distribute x86 packages.

installer/bundle/bundle_skel.sh
NarineM: @AaronRM  added . It was potential fix for an issue reported only for redhat 7.5 (but the issue itself was not reproducible locally, only in customer env). I would have removed it completely, but because we never got to the actual root cause and whether this fix really resolved the issue, I feel we should leave it here for now. Eventually , we will remove it for red 7.5 too if customer see issues after we start using it with only 7.5

installer/bundle/bundle_skel.sh
NarineM: This part was not related to the PR, just some clean up of dead code that is called in every install,

PR #1452, opened 2022-04-12
1.12.40 does not have system_config.file/dir_permission

hestolz:


source/ext/patches/fluentd/in_tail.patch
```diff
@@ -14,8 +14,8 @@
      end
  
      def configure_parser(conf)
-+      @file_perm = system_config.file_permission || FILE_PERMISSION
-+      @dir_perm = system_config.dir_permission || DIR_PERMISSION
```
NarineM, line 18: why this change is needed now? Did ruby fluentd also make this change? Have you tested custom log flow with this change to see if it works ?

NarineM: What scenarios get enabled with pos fix for in_tail and why we need the patch?

hestolz: Hi @NarineM , I tested this, the patch applies and I am seeing custom logs successfully. 

> why this change is needed now? Did ruby fluentd also make this change? Have you tested custom log flow with this change to see if the issue is resolved?

This is to solve a customer-related issue (shared over teams, 279165709). In some scenarios you can get 

```
2021-12-21 12:46:15 +0000 [error]: ERROR Tailfilereader crashed due to an unexpected exit --- No such file or directory @ rb_sysopen
```

NarineM (approved):

PR #1460, opened 2022-05-19
Install GPG if not on VM

aswatt:
For other extra packages we need (e.g. tar, sed, curl), we attempt to install them on the VM if they don't exist. Setting up the same for gpg, to allow for smooth bundle install.

### DIFF
--- a/installer/bundle/bundle_skel.sh
+++ b/installer/bundle/bundle_skel.sh
@@ -1256,9 +1256,9 @@ if [ "$installMode" = "I" -o "$installMode" = "U" ]; then
         fi
     fi
 
-    check_if_program_exists_on_system gpg
+    install_if_program_does_not_exist_on_system gpg
     if [ $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


NarineM:
@aswatt Looks good.  l can sign off as soon as you confirm with this change bundle installs on debian 10 when onboarding it directly without vm extension

aswatt:
> @aswatt Looks good. l can sign off as soon as you confirm with this change bundle installs on debian 10 when onboarding it directly without vm extension. gpg might pull dependancy libs during it's install and we need to confirm that runs fine on azure debian 10 image

@NarineM Just confirmed that the shell bundle with this change installs without any issues, and data can flow successfully! It looks like GPG does install several of its dependencies, but none of them seem to cause any issues on Debian 10.

(Note for posterity: all other supported distros have GPG already installed on them, so this change wouldn't have any effect.)

NarineM (approved):

PR #1466, opened 2022-05-27
Remove read_from_head true from operation.conf

taglines:


NarineM:
@taglines do you think it should be set to false? or you do not want to support begining of file reading at all given that it does not check on the overall size or perform any slicing if pos file is missing?

NarineM (approved):

NarineM:
The default (wihen no setting) is false, There is no option for more complex logic with “@type tail”. With false, “@type tail” starts at the end of the file is .pos file is missing, otherwise it picks up from the .pos position.

PR #1483, opened 2022-07-19
Update onboarding script and dsc monitor plugin to use python3 over python2 if both versions are present

NarineM:


source/code/plugins/in_dsc_monitor.rb
```diff
@@ -57,8 +57,8 @@ def run_check_install
 
     def get_dsc_status
       begin
-        python = %x(which python2)
-        if !python.empty?
+        python = (%x(command -v python3)).to_i
```
hestolz, line 60: Are you sure we can use `command` in ruby code? when i try it in `irb`, it is not found

```
situhan@lazarus:~/Git/Compute-Runtime-Tux/mdsdtest/fuzzing/listeners$ ruby -c "%x(command -v python3)"
Traceback (most recent call last):
ruby: No such file or directory -- %x(command -v python3) (LoadError)
```

test/code/plugins/in_dsc_monitor_plugintest.rb
```diff
@@ -10,7 +10,7 @@ class DscMonitorTest < Test::Unit::TestCase
   CHECK_DSC_INSTALL = "dpkg --list omsconfig > /dev/null 2>&1; echo $?"
   CHECK_DSC_STATUS = "/opt/microsoft/omsconfig/Scripts/TestDscConfiguration.py"
   CHECK_DSC_STATUS_PYTHON_3 = "/opt/microsoft/omsconfig/Scripts/python3/TestDscConfiguration.py"
-  CHECK_PYTHON = "which python2"
+  CHECK_PYTHON = "command -v python3"
```
hestolz, line 13: same question here

hestolz (changes requested):

source/code/plugins/in_dsc_monitor.rb
NarineM: Yes, I wanted to make this code to use same path for checking python version as the shell script, but apparently, %x is not handling all system cmds same way. I will change it back to be with "which" command.

hestolz (approved):

PR #1489, opened 2022-08-01
Upgrade Ruby to 2.6.10

hestolz:
[Ruby diff](https://github.com/ruby/ruby/compare/v2_6_8...v2_6_10)


NarineM:
@hestolz Can you confirm that custom log and perf counters were tested with this change?

hestolz:
> @hestolz Can sign off if you confirm that custom log and syslog were tested with this change.

Hi @NarineM , this would be covered by the pre-release testing

NarineM (approved):

NarineM:
> > @hestolz Can sign off if you confirm that custom log and syslog were tested with this change.
> 
> Hi @NarineM , this would be covered by the pre-release testing

Any PR needs to be tested before checkin :) This particular change  also impacts partners as majority of our partner plugins are depending on tail plugin.

PR #1493, opened 2022-08-11
Update the agent link with new version

tduarte14:
Use link directly for latest instead of version. Was using a very old version (1.2.0-25, where latest is already 1.14.19-0)


tduarte14:
Can someone check this please? 1 month passed since reported

NarineM (approved):

PR #1502, opened 2022-09-15
Add rpm no-digest parameter support to omsagent installer to skip digest validation when installing dependencies if specified in cmdline

NarineM:


installer/bundle/bundle_skel.sh
```diff
@@ -96,6 +96,7 @@ usage()
     echo "  --force                         Force upgrade (override version checks)."
     echo "  --install                       Install the package from the system."
     echo "  --purge                         Uninstall the package and remove all related data."
+    echo "  --noDigest                      RPM manager skips verification of package or header digests when reading (same as rpm --nodigest --nofiledigest)"
```
hestolz, line 99: just wondering do we have an equivalent for DPKG, e.g. should we have this apply as well [--no-debsig](https://manpages.debian.org/testing/dpkg/dpkg.1.en.html#no~4)

hestolz (approved):

installer/bundle/bundle_skel.sh
NarineM: answered over teams channel :)

PR #1536, opened 2023-06-28
new bundle for docker-cimprov 1.0.0-41

sarahpeiffer:
New docker-cimprov package contains the latest OMI 1.7.1-0 built with compiler mitigated flags


NarineM (changes requested):
Please update the docker to include omi version 1.7.1-0 version that is the official candidate for MSRC.

NarineM (approved):

PR #1540, opened 2023-08-22
remove future python references to address s360

NarineM:
Should be checked in only after this PR is checked in: https://github.com/Azure/azure-linux-extensions/pull/1765


NarineM:
Accidental change. no longer needed.

PR #1541, opened 2023-10-02
Add openssl version 3 support

NarineM:


build/Makefile
```diff
@@ -212,6 +220,9 @@ ifeq ($(PF_ARCH),x64)
 	cd $(OMI_ROOT); $(RMDIR) output; ln -s output_openssl_1.1.0 output
 	$(MAKE) -C $(DSC_DIR) dsc110
 	$(MAKE) -C $(DSC_DIR) dsckit110
+	cd $(OMI_ROOT); $(RMDIR) output; ln -s output_openssl_3.0.0 output
```
simathih, line 223: this is already done 4 lines above except the linking part

installer/bundle/bundle_skel.sh
```diff
@@ -302,15 +302,18 @@ ulinux_detect_openssl_version()
     OPENSSL_SYSTEM_VERSION_10X=`echo $OPENSSL_SYSTEM_VERSION_FULL | grep -Eq '^1.0.'; echo $?`
     OPENSSL_SYSTEM_VERSION_100_ONLY=`echo $OPENSSL_SYSTEM_VERSION_FULL | grep -Eq '^1.0.0'; echo $?`
     OPENSSL_SYSTEM_VERSION_11X=`echo $OPENSSL_SYSTEM_VERSION_FULL | grep -Eq '^1.1.'; echo $?`
+    OPENSSL_SYSTEM_VERSION_30X=`echo $OPENSSL_SYSTEM_VERSION_FULL | grep -Eq '^3.0.'; echo $?`
```
simathih, line 305: basically we package 3 versions each of deb/rpm for 1.0.0, 1.1.0 and 3.0 and install based on what is available in cx vm? Is this correct? How much has package size increased to now?

simathih (approved):

build/Makefile
NarineM: ah, it is from merge probably, will take a look. thx

installer/bundle/bundle_skel.sh
NarineM: yes, total bundle size is now about 235MB. Previously it was about 155MB , increase is due to 2 ruby and adding extra packages for 300 omi,scx, omsagent, dsc.

build/Makefile
NarineM: I think it needs to stay to make sure CMAKE does not change the current folder, also to clean up inter output dir before building 300 kits. same is written for 100 and 110 paths.

PR #1542, opened 2023-10-02
update oms docker image to build opensslv3

NarineM:


.github/workflows/omsbuild.yml
```diff
@@ -34,13 +34,13 @@ jobs:
         ls -la Build-OMS-Agent-for-Linux/
 
     - name: Pull docker image
-      run: docker pull abenbachir/oms-centos6-x64:latest
+      run: docker pull narinem/oms-centos6-x64:3.2
```
simathih, line 37: is this some private ACR? why not linuxma team ACR?

simathih (approved):

.github/workflows/omsbuild.yml
NarineM: this one is just for github travis merge validation infra, our build pipeline is not using it. Travis can only pull from public repos , we had it setup with public docker hub before , it is just I do not have access to change it in Abdhir's repo, so renamed it to mine.

